PR #1, opened 2018-10-22
*: move github.com/pingcap/tidb/parser here

tiancaiamao:
Here is the dependency graph:
parser

TiDB need to update its "import github.com/pingcap/tidb/parser" to "import github.com/pingcap/parser"


tiancaiamao:
@shenli

kennytm:
Question about package management... As `parser.go` isn't checked in, do the dependencies still need to manually `make parser` and check it into their `vendor/*`?

tiancaiamao:
`parser.go` will be included in this repository, although I didn't do it in this commit @kennytm

PR #2, opened 2018-10-25
parser: support grammar use index(primary) without backquote

XuHuaiyu:
fix the issue raised in https://github.com/pingcap/tidb/issues/8038


kennytm:
`parser.go` isn't updated 😐

ngaut:
LGTM

Issue #6, opened 2018-11-01
This repository needs a CI


kennytm:
1. Ensure `parser.go` is up-to-date with `parser.y`
2. Run `go test`.

Unfortunately the `yyXLAT` array generated by `goyacc` is unstable regarding case-insensitive keywords:

<details><summary>Potential diff after running `make parser`</summary>

```diff
@@ -1054,8 +1053,8 @@ var (
 		57812: 533, // AssignmentList (3x)
 		57825: 534, // ByItem (3x)
 		57842: 535, // ColumnPosition (3x)
-		57848: 536, // Constraint (3x)
-		57380: 537, // constraint (3x)
+		57380: 536, // constraint (3x)
+		57848: 537, // Constraint (3x)
 		57850: 538, // ConstraintKeywordOpt (3x)
 		57892: 539, // ExplainableStmt (3x)
 		57909: 540, // FloatOpt (3x)
@@ -1292,8 +1291,8 @@ var (
 		58064: 771, // ShowIndexKwd (1x)
 		58068: 772, // ShowTargetFilterable (1x)
 		58072: 773, // Start (1x)
-		58073: 774, // Starting (1x)
-		57501: 775, // starting (1x)
+		57501: 774, // starting (1x)
+		58073: 775, // Starting (1x)
 		58075: 776, // StatementList (1x)
 		57504: 777, // stored (1x)
 		58080: 778, // StringType (1x)
@@ -1891,8 +1890,8 @@ var (
 		"AssignmentList",
 		"ByItem",
 		"ColumnPosition",
-		"Constraint",
 		"constraint",
+		"Constraint",
 		"ConstraintKeywordOpt",
 		"ExplainableStmt",
 		"FloatOpt",
@@ -2129,8 +2128,8 @@ var (
 		"ShowIndexKwd",
 		"ShowTargetFilterable",
 		"Start",
-		"Starting",
 		"starting",
+		"Starting",
 		"StatementList",
 		"stored",
 		"StringType",
@@ -3304,7 +3303,7 @@ var (
 		{539, 1},
 		{776, 1},
 		{776, 3},
-		{536, 2},
+		{537, 2},
 		{662, 1},
 		{662, 1},
 		{662, 4},
@@ -3525,8 +3524,8 @@ var (
 		{710, 3},
 		{736, 0},
 		{736, 3},
-		{774, 0},
-		{774, 3},
+		{775, 0},
+		{775, 3},
 		{737, 0},
 		{737, 3},
[...]

PR #9, opened 2018-11-01
goyacc: fix unstability of generated parser.go

kennytm:
Previously, running `make parser` two times may produce different contents in `parser.go`. This is because we have two pairs of tokens differed only by case: **"Starting"/"starting"** and **"Constraint"/"constraint"**. `goyacc` by default sort the tokens by usage and lower-cased name using `sort.Sort` which is unstable, causing non-deterministic output. This PR fixes the issue in the parser generator by forcing the upper-case tokens be sorted before the lower-case ones.

This PR is needed for #6.


tiancaiamao:
LGTM

shenli:
Good catch!

PR #12, opened 2018-11-03
*: support new aggregation function var_pop and var_samp

mccxj:


mccxj:
@zz-jason PTAL

tiancaiamao:
Would you please fix the conflicts and link the TiDB PR. @mccxj 
For more instructions, see https://github.com/pingcap/parser/#how-to-update-parser-for-tidb

parser.y
```diff
@@ -3853,13 +3857,37 @@ SumExpr:
 	{
 		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
 	}
-|	builtinStddevPop '(' BuggyDefaultFalseDistinctOpt Expression ')'
+|	builtinStddevPop '(' Expression ')'
```
lzmhhh123: We'd better support `distinct` clause in variance aggregate function.

zz-jason: @mccxj After discussion, we decide to support `distinct` for every aggregate functions.

PR #14, opened 2018-11-05
*: add ActionModifyTableCharsetAndCollate to support alter table charset and collate

winkyao:
Required for https://github.com/pingcap/tidb/pull/8037


ast/ddl.go
```diff
@@ -821,6 +821,8 @@ type TableOption struct {
 	Tp        TableOptionType
 	StrValue  string
 	UintValue uint64
+	// Skipped indicate the TableOption can be skipped.
+	Skipped bool
```
tiancaiamao: I think this field is confusing, it's not something belong to TableOption

model/ddl.go
```diff
@@ -29,28 +29,29 @@ type ActionType byte
...
+	ActionRebaseAutoID                 ActionType = 13
+	ActionRenameTable                  ActionType = 14
+	ActionSetDefaultValue              ActionType = 15
+	ActionShardRowID                   ActionType = 16
+	ActionModifyTableComment           ActionType = 17
+	ActionRenameIndex                  ActionType = 18
+	ActionAddTablePartition            ActionType = 19
+	ActionDropTablePartition           ActionType = 20
+	ActionCreateView                   ActionType = 21
+	ActionModifyTableCharsetAndCollate ActionType = 22
```
crazycs520: Need change `var actionMap = map[ActionType]string` ?

winkyao: address

PR #19, opened 2018-11-06
*: support for the statement of "admin show tableName next row id"

zimulala:
support for the statement of "admin show tableName next row id".


zimulala:
PTAL @winkyao @crazycs520 @zz-jason

PR #20, opened 2018-11-06
parser,ast: fix the TiDB issue #8153

dbjoa:
The TiDB PR [8206](https://github.com/pingcap/tidb/pull/8206) requires this PR.

### DIFF
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -699,7 +699,7 @@ func (n *ParenthesesExpr) Accept(v Visitor) (Node, bool) {
 type PositionExpr struct {
 	exprNode
 	// N is the position, started from 1 now.
-	N int
+	N ExprNode
 	// Refer is the result field the position refers to.
 	Refer *ResultField
 }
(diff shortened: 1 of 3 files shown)


zz-jason:
@winoros @eurekaka PTAL

ast/expressions.go
```diff
@@ -715,6 +715,13 @@ func (n *PositionExpr) Accept(v Visitor) (Node, bool) {
...
 		return v.Leave(newNode)
 	}
+	if n.N != nil {
+		node, ok := n.N.Accept(v)
+		if !ok {
+			return n, false
+		}
+		n.N = node.(ExprNode)
+	}
 	n = newNode.(*PositionExpr)
```
eurekaka: Put this line before the added check?

dbjoa: Thank you for the comment. I've updated the PR.

eurekaka: I mean like this:
```go
n = newNode.(*PositionExpr)
if n.N != nil {
    node, ok := n.N.Accept(v)
    if !ok {
        return n, false
    }
    n.N = node.(ExprNode)
}
```
we should modify `newNode.N` instead of `n.N`?

dbjoa: Thank you for the correction. I've updated the PR.

PR #24, opened 2018-11-07
Revert "parser,ast: fix the TiDB issue #8153 (#20)"

dbjoa:
This reverts commit 80638229692790da5b0152eb4d446738611f6efd.


tiancaiamao:
LGTM

We can add it back later, when it's ready.
PTAL @zz-jason

PR #32, opened 2018-11-08
support sql digest

lysu:

Now, we are hard to identity a group of similar SQLs, but in real product most SQLs execute by program is some similar SQLs that only different in parameter part(yes they maybe not use prepare too).

This PR  want to generate a digest for sql group, so later we can do some useful sth base on it.

- online add some hint without modify user's code.
- add some sql into blacklist(e.g. some sql are too slow, but client continue send to tidb, we need protect other vital sql's execution, so add them to blacklist)
- help user do prepare in slient..this is complex and current code is far away from that.. - -

- use lex to replace intLit, stringLit and so on...
- sha256 result

### Check List 

Tests 

 - Unit test

Code changes

 - Add exported method

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/32)


kennytm:
> md5

could we not use MD5 if not for backward compatibility? 😂

morgo:
> could we not use MD5 if not for backward compatibility? 😂

Yes, I think it is important for compatibility.  Although I will clarify that I think it should be a non-requirement to have matching digests to MySQL.  It sounds just too difficult.

morgo:
Two functions to be aware of:

- [STATEMENT_DIGEST(statement)](https://dev.mysql.com/doc/refman/8.0/en/encryption-functions.html#function_statement-digest) - return the digest of a statement
- [STATEMENT_DIGEST_TEXT(statement)](https://dev.mysql.com/doc/refman/8.0/en/encryption-functions.html#function_statement-digest-text) - return the normalized sql query

Although they are from 8.0, you might find adding them useful for writing tests.

kennytm:
Interesting. On MySQL 8.0 the hash function is changed from MD5 to SHA-256, but the output definitely isn't the SHA-256 of the normalized string. 

```
mysql> select statement_digest_text('Select 1');
+-----------------------------------+
| statement_digest_text('Select 1') |
+-----------------------------------+
| SELECT ?                          |
+-----------------------------------+
1 row in set (0.00 sec)

mysql> select statement_digest('Select 1');
+------------------------------------------------------------------+
| statement_digest('Select 1')                                     |
+------------------------------------------------------------------+
| d1b44b0c19af710b5a679907e284acd2ddc285201794bc69a2389d77baedddae |
+------------------------------------------------------------------+
1 row in set (0.00 sec)
```

```python
>>> hashlib.sha256(b'SELECT ?').hexdigest()
'66cbb3a40d4bbd150b75825ad291a6545399f3098fc1079e4d8b5bb061a6a481'
```

Apparently MySQL hashes the parsed token stream directly (we could do the same here 😉), meaning we're never going to match the digest [as discussed before](https://github.com/pingcap/parser/pull/32#issuecomment-437164554).

lysu:
@gregwebs @kennytm thanks, I will take a look~

morgo:
@kennytm wow, looks like it did change to `sha256`.  The second sentence in these manual pages is different:
https://dev.mysql.com/doc/refman/8.0/en/performance-schema-statement-digests.html
https://dev.mysql.com/doc/refman/5.7/en/performance-schema-statement-digests.html

normalize.go
```diff
@@ -0,0 +1,62 @@
+package parser
```
lamxTyler: Add license?

normalize_test.go
```diff
@@ -0,0 +1,46 @@
+package parser_test
```
lamxTyler: Ditto.

normalize_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/pingcap/parser"
+	"testing"
```
lamxTyler: Put it in the front?

normalize.go
```diff
@@ -0,0 +1,75 @@
...
+	"reflect"
+	"sync"
+	"unicode"
+	"unsafe"
+)
+
+// Digest generates a digest(or sql-id) for a SQL.
+// the purpose of digest is to identity a group of similar SQLs, then we can do other logic base on it.
+func Digest(sql string) string {
+	d := digesterPool.Get().(*digester)
```
zz-jason: how about:
```go
d := digesterPool.Get().(*digester)
defer func() {
	d.buffer.Reset()
	digesterPool.Put(d)
}() 
```

and remove line 48~49?

normalize.go
```diff
@@ -0,0 +1,75 @@
...
+	sqlLen := len(sql)
+	d.lexer.reset(sql)
+	for {
+		tok, pos, lit := d.lexer.scan()
+		if tok == unicode.ReplacementChar && d.lexer.r.eof() {
+			break
+		}
+		if pos.Offset == sqlLen {
```
XuHuaiyu: s/ sqlLen/ len(sql)

normalize.go
```diff
@@ -0,0 +1,75 @@
...
+
+var digesterPool = sync.Pool{
+	New: func() interface{} {
+		return &digester{
+			lexer: NewScanner(""),
+		}
+	},
+}
+
+type digester struct {
```
XuHuaiyu: We need a comment for this .

normalize.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		d.buffer.WriteRune(' ')
+		switch tok {
+		case intLit, stringLit, decLit, floatLit, bitLit, hexLit:
+			d.buffer.WriteRune('?')
+		default:
+			d.buffer.WriteString(lit)
+		}
+	}
+	result := md5.Sum(d.buffer.Bytes())
```
XuHuaiyu: when will len(result) == 0?

normalize.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	},
+}
+
+type digester struct {
+	buffer bytes.Buffer
+	lexer  *Scanner
+}
+
+func byte2str(b []byte) (s string) {
```
zz-jason: could you add a comment about this function to notice the user that it makes a shallow copy of the contents of byte slice to construct a string?

normalize.go
lysu: if we should better don't use `defer` as we can in hot-code path and I have do some refactor PTAL :D 
https://medium.com/i0exception/runtime-overhead-of-using-defer-in-go-7140d5c40e32

lysu:
@zz-jason @XuHuaiyu PTAL PTAL thx~

lysu:
@zz-jason @XuHuaiyu PTAL~

digester.go
```diff
@@ -83,17 +88,122 @@ func (d *sqlDigester) normalize(sql string) {
...
+		if i != len(d.tokens)-1 {
+			d.buffer.WriteRune(' ')
+		}
+	}
+	d.tokens = d.tokens[:0]
+}
+
+func (d *sqlDigester) genericLit(currTok token) token {
+	// "?, ?, ?, ?" => "..."
+	last2 := d.tokens.back(2)
```
kennytm: This means `(1,2,3,x,5,6,7)` and `(1,2,3,4,x,6,7)` will both become `(...,x,...)`?

digester_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, test := range tests {
+		actual := parser.DigestText(test.input)
+		c.Assert(actual, Equals, test.expect)
+	}
+}
+
+func (s *testSQLDigestSuite) TestDigestHashEqForSimpleSQL(c *C) {
+	sqlGroups := [][]string{
+		{"select * from b where id = 1", "select * from b where id = '1'", "select * from b where id =2"},
+		{"select 2 from b, c where b.id = c.id where c.id > 1", "select 4 from b, c where b.id = c.id where c.id > 23"},
```
kennytm: There are two `where` clauses here... is it supposed to work? 😅

digester.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	// 2 => ?
+	currTok.tok = genericSymbol
+	currTok.lit = "?"
+	return currTok
+}
+
+func (d *sqlDigester) isPrefixByUnary(currTok int) (isUnary bool) {
+	if !isNumLit(currTok) {
```
XuHuaiyu: It seems that isNumLit(-1) will return false. So isPrefixByUnary(-1) return false?

```
SignedLiteral:
	Literal
	{
		$$ = ast.NewValueExpr($1)
	}
|	'+' NumLiteral
	{
		$$ = &ast.UnaryOperationExpr{Op: opcode.Plus, V: ast.NewValueExpr($2)}
	}
|	'-' NumLiteral
	{
		$$ = &ast.UnaryOperationExpr{Op: opcode.Minus, V: ast.NewValueExpr($2)}
	}

NumLiteral:
	intLit
|	floatLit
|	decLit
```

digester.go
```diff
@@ -0,0 +1,253 @@
...
+	"crypto/sha256"
+	"fmt"
+	hash2 "hash"
+	"strings"
+	"sync"
+	"unicode"
+)
+
+// DigestHash generates the digest of statements.
+func DigestHash(sql string) (result string) {
```
XuHuaiyu: We need more comments for DigestHash and DigestText.
An example may be better.

digester.go
```diff
@@ -0,0 +1,253 @@
...
+func (d *sqlDigester) doDigestText(sql string) (result string) {
+	d.normalize(sql)
+	result = string(d.buffer.Bytes())
+	d.lexer.reset("")
+	d.buffer.Reset()
+	return
+}
+
+const (
+	genericSymbol     = -1
```
XuHuaiyu: We need to add comments for these 2 constants.

digester.go
```diff
@@ -0,0 +1,253 @@
...
+	isComma = tok == 44
+	return
+}
+
+type token struct {
+	tok int
+	lit string
+}
+
+type tokenDeque []token
```
XuHuaiyu: `Queue` is enough?

lysu: it will be push/pop in back and  foreach from front, maybe deque more suitable

digester.go
```diff
@@ -0,0 +1,267 @@
...
+	"strings"
+	"sync"
+	"unicode"
+)
+
+// DigestHash generates the digest of statements.
+// it will generate a hash on normalized form of statement text
+// which removes general property of a statement but keeps specific property.
+//
+// for example: both DigestHash('select 1') and DigestHash('SelEct 2') => e1c71d1661ae46e09b7aaec1c390957f0d6260410df4e4bc71b9c8d681021471
```
eurekaka: s/selEct/select

digester.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+const (
+	// genericSymbol presents parameter holder ("?") in statement
+	// it can be any value as long as it is not repeated with other tokens.
+	genericSymbol = -1
+	// genericSymbolList presents parameter holder lists ("?, ?, ...") in statement
+	// it can be any value as long as it is not repeated with other tokens.
+
+	// it can be any value as long as it is not repeated with other tokens.
```
eurekaka: redundant lines?

digester.go
```diff
@@ -0,0 +1,267 @@
...
+	for {
+		token := d.tokens.popBack(1)
+		if len(token) == 0 {
+			return
+		}
+		if token[0].tok == hintBegin {
+			return
+		}
+	}
+}
```
eurekaka: Why should the hint be removed during normalization?

lexer.go
```diff
@@ -257,7 +257,8 @@ func (s *Scanner) scan() (tok int, pos Pos, lit string) {
 		}
 		node = node.childs[ch0]
 		if node.fn != nil {
-			return node.fn(s)
+			tok, pos, lit = node.fn(s)
+			return
```
eurekaka: This is unnecessary change?

digester.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	if last2[1].lit != "by" {
+		return false
+	}
+	orderBy = last2[0].lit == "order"
+	return
+}
+
+func isComma(tok int) (isComma bool) {
+	isComma = tok == 44
```
eurekaka: Substitute magic numbers with constant?

digester.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	switch last2[0].tok {
+	case genericSymbol, genericSymbolList:
+		generic = true
+	default:
+	}
+	return
+}
+
+func isOrderBy(last2 []token) (orderBy bool) {
```
eurekaka: How about `group by`? for example:
```
mysql> select count(a), b from t group by (2);
+----------+------+
| count(a) | b    |
+----------+------+
|        1 | NULL |
|        0 |    2 |
+----------+------+
2 rows in set (0.00 sec)
```
here `group by 2` means `group by b`.

digester.go
```diff
@@ -0,0 +1,267 @@
...
+		if len(token) == 0 {
+			return
+		}
+		if token[0].tok == hintBegin {
+			return
+		}
+	}
+}
+
+func (d *sqlDigester) reduceLit(currTok token) token {
```
eurekaka: Seems both `count(*)` and `count(1)` should be reduced to `count(?)`, so we need to check if `*` is in `count` context?

digester.go
kennytm: Pretty sure the strange casing is intended, to show that the digest is case-insensitive.

digester.go
lysu: "*" is an special token "42:    264, // '*' (483x)" and will not go to here, and I added a new test for this case :)

digester.go
lysu: I'm not sure this too, maybe remove hint make SQL binding easier when we want to force plan no matter what hint in current sql?

eurekaka: Is this behavior compatible with MySQL? @lamxTyler any comments on this behavior?

eurekaka: /cc @morgo

digester.go
eurekaka: Oh, I didn't mean we should add check in this specific function, I mean `count(1)` or `count(2)` would be reduced to `count(?)` by this PR, while `count(*)` is still `count(*)`, but they should have same fingerprint actually?

lysu: so  `count(a)`  and `count(1)` should be same to ?

digester.go
lysu: no, it's too hard compatible mysql value, but on other side, we do more things than mysql, mysql's digest doesn't handle logic like `group/order by`, because our digest is used for sql plan binding.

digester.go
lysu: in practice, maybe one program(not human typing) should not often send count(*) in frist time and later send count(1),  it is too hard to exhaustive all equivalence plan in here..

lysu: @eurekaka my fault "count(*)" neq "count(a)" :rofl: , comment addressed- -

digester.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+type tokenDeque []token
+
+func (s *tokenDeque) pushBack(t token) {
+	*s = append(*s, t)
+}
+
+func (s *tokenDeque) popBack(n int) (t []token) {
+	t = (*s)[len(*s)-n:]
```
zz-jason: should we check `len(*s)-n < 0`?

digester.go
lysu: ping @lamxTyler, should plan manage need that? if so I will take that :rofl:

digester.go
```diff
@@ -0,0 +1,301 @@
...
+	digesterPool.Put(d)
+	return
+}
+
+// DigestText generates the normalized statements.
+// it will get normalized form of statement text
+// which removes general property of a statement but keeps specific property.
+//
+// for example: DigestText('select 1 from b where a = 1') => 'select ? from b where a = ?'
+func DigestText(sql string) (result string) {
```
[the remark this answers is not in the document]
lysu: it take care this name too, but mysql use `statement_digest_text` as function name

digester.go
lamxTyler: It would be good it if we can do that...

eurekaka: > Should we also ignore the `force index()`/`use index()`/`ignore index()`/`straight_join` key words?

Then we should remove these as well?

lysu: removed

lysu:
@zz-jason @XuHuaiyu  PTAL

lysu:
@zz-jason @XuHuaiyu @eurekaka PTAL

PR #37, opened 2018-11-12
parser: support window function ast

lamxTyler:
This PR supports all window function ast.
PTAL @zz-jason @lzmhhh123 @winoros


parser.y
```diff
@@ -3844,48 +3843,88 @@ TrimDirection:
 SumExpr:
 	"AVG" '(' BuggyDefaultFalseDistinctOpt Expression ')'  OptWindowingClause
 	{
-		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
+		if $6 != nil {
+			$$ = &ast.WindowFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool), Spec: *($6.(*ast.WindowSpec)),}
+		} else {
+			$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
```
lzmhhh123: Is it better to merge `AggregateFuncExpr` in `WindowFuncExpr`?

lamxTyler: I think it is better not to, because we need to first execute the `group-by` clause, and only after that we can execute the window expression.

tiancaiamao:
Would you please refer the TiDB PR ? @lamxTyler 
https://github.com/pingcap/parser#how-to-update-parser-for-tidb

tiancaiamao:
Would you please refer the TiDB PR ? @lamxTyler 
https://github.com/pingcap/parser#how-to-update-parser-for-tidb

tiancaiamao:
You can use the parser branch from @lamxTyler 's repo before this PR is merged:

```
GO111MODULE=on go replace mod edit github.com/pingcap/parser=github.com/lamxTyler/parser@window-func-ast
```

@lzmhhh123

lamxTyler:
@lzmhhh123 It depends on you. If the planner needs to know if the select stmt contains any window function expression, then I will add it for you.

PR #44, opened 2018-11-19
parser: add driver import check & example

### LINKED ISSUE #43: some test case failed
```go
func main(){

    sql := "select * from wn where id = 2"

    pa := parser.New()

    //set charsetInfo and collation
    charsetInfo, collation := "utf8", "utf8_general_ci" 

    pa.SetSQLMode(mysql.ModeNone)

    stmtNodes, err := pa.Parse( sql, charsetInfo, collation)

    fmt.Println(stmtNodes[0], err)
}
```

It works when I use sql="select * from wn"

But it failed when I use sql="select * from wn where id = 2"

The golang version is 1.11.1 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5e3ef8]

goroutine 1 [running]:
github.com/pingcap/parser.yyParse(0x715e60, 0xc000154048, 0xc000154000, 0xc00010bf68)
	/root/wn/go/pkg/mod/github.com/pingcap/parser@v0.0.0-20181119030059-e8ee63ef2cb2/parser.go:8654 +0x16ac8
github.com/pingcap/parser.(*Parser).Parse(0xc000154000, 0x7fffc6f056eb, 0x1b, 0x6b7519, 0x4, 0x6bd81d, 0xf, 0x0, 0x0, 0xc00010bf10, ...)
	/root/wn/go/pkg/mod/github.com/pingcap/parser@v0.0.0-20181119030059-e8ee63ef2cb2/yy_parser.go:101 +0x15f
main.main()
	/root/wn/go/test2/hello.go:26 +0x2a9
```

lysu:
fixes #43

1. add driver import check when create parser, give better panic message.

2. add parser example, so [godoc](https://godoc.org/github/pingcap/parser) will have example just like this 
https://godoc.org/gopkg.in/lysu/parser.v1 after this PR merged

@tiancaiamao

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/44)

### DIFF
--- a/parser_example_test.go
+++ b/parser_example_test.go
@@ -1,0 +1,35 @@
+// Copyright 2018 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser_test
+
+import (
+	"fmt"
+	"github.com/pingcap/parser"
+	// 0. import parser_driver implemented by TiDB(user also can implement own driver by self).
+	_ "github.com/pingcap/tidb/types/parser_driver"
+)
+
+// This example show how to parse a text sql into ast.
+func Example_parseSQL() {
+
+	// 1. Create a parser, this is a NOT thread-safe but heavy object,
+	// it is better to reuse it in thread-safe way as possible  as we can.
+	p := parser.New()
+
+	// 2. Parse a text SQL into AST([]ast.StmtNode)
+	stmtNodes, err := p.Parse("select * from tbl where id = 1", "", "")
+
+	// 3. Use AST to do cool things~
+	fmt.Println(stmtNodes[0], err)
+}
--- a/parser_example_test.go
+++ b/parser_example_test.go
@@ -23,6 +23,9 @@ import (
 // This example show how to parse a text sql into ast.
 func Example_parseSQL() {
 
+	// 0. make sure import parser_driver implemented by TiDB(user also can implement own driver by self).
+	// import _ "github.com/pingcap/tidb/types/parser_driver" in import part.
+
 	// 1. Create a parser, this is a NOT thread-safe but heavy object,
 	// it is better to reuse it in thread-safe way as possible  as we can.
 	p := parser.New()
--- a/parser_example_test.go
+++ b/parser_example_test.go
@@ -25,13 +25,15 @@ func Example_parseSQL() {
 	// 0. make sure import parser_driver implemented by TiDB(user also can implement own driver by self).
 	// and add `import _ "github.com/pingcap/tidb/types/parser_driver"` in the head of file.
 
-	// 1. Create a parser, this is a NOT thread-safe but heavy object,
-	// it is better to reuse it in thread-safe way as possible as we can.
+	// 1. Create a parser. The parser is NOT goroutine safe and should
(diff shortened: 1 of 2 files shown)


lysu:
orz, godoc doesn't show my import..

parser_example_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/pingcap/parser"
+	// 0. import parser_driver implemented by TiDB(user also can implement own driver by self).
+	_ "github.com/pingcap/tidb/types/parser_driver"
+)
+
+// This example show how to parse a text sql into ast.
+func Example_parseSQL() {
+
+	// 0. make sure import parser_driver implemented by TiDB(user also can implement own driver by self).
+	// import _ "github.com/pingcap/tidb/types/parser_driver" in import part.
```
kennytm: These two lines can be removed

parser_example_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// 1. Create a parser, this is a NOT thread-safe but heavy object,
+	// it is better to reuse it in thread-safe way as possible  as we can.
```
kennytm: I don't understand, do you mean e.g. use a `sync.Mutex` to protect it if you want to access the parser from multiple threads?

tiancaiamao: Each goroutine should use it's own parser @kennytm 

Don't do this:
```
p := parser.New()
go func() { p.Parse(...) } ()
go func() { p.Parse(...) } ()
```

parser_example_test.go
```diff
@@ -0,0 +1,38 @@
...
+	p := parser.New()
+
+	// 2. Parse a text SQL into AST([]ast.StmtNode)
+	stmtNodes, err := p.Parse("select * from tbl where id = 1", "", "")
+
+	// 3. Use AST to do cool things~
```
tiancaiamao: s/things~/things

parser_example_test.go
tiancaiamao: Step 0 doesn't take effect ? @lysu

parser_example_test.go
kennytm: Thanks @tiancaiamao. The comment should be rephrased to not use the word "reuse" then. Maybe you mean "move"?

parser_example_test.go
kennytm: If we need to display these two lines on godoc.org, then the comment on the actual `import` on line 19 should be shortened as this is repeating the same information...

(The actual `import` not showing is a bug of godoc: golang/go#8223, golang/gddo#433.)

lysu: @kennytm  yes..I tried and godoc doesn't show that so repeat it in here...- -

parser_example_test.go
lysu: @kennytm  I mean "reuse it in thread-local level", but how to say it :rofl:

kennytm: @lysu  I would split it into two sentences.

```go
// 1. Create a parser. The parser is NOT goroutine safe and should
// not be shared among multiple goroutines. However, parser is also
// heavy, so each goroutine should reuse its own local instance if 
// possible.
```

parser_example_test.go
```diff
@@ -0,0 +1,37 @@
...
+	// and add `import _ "github.com/pingcap/tidb/types/parser_driver"` in the head of file.
+
+	// 1. Create a parser, this is a NOT thread-safe but heavy object,
+	// it is better to reuse it in thread-safe way as possible as we can.
+	p := parser.New()
+
+	// 2. Parse a text SQL into AST([]ast.StmtNode)
+	stmtNodes, err := p.Parse("select * from tbl where id = 1", "", "")
+
+	// 3. Use AST to do cool thing~
```
kennytm: you should remove the `~` not the `s` 🤣

tiancaiamao:
LGTM

PR #45, opened 2018-11-19
mysql: update sql-mode default

morgo:
Change TiDB to use the same default as MySQL 5.7

Depends on https://github.com/pingcap/tidb/pull/8278 (merged)


morgo:
See also: https://github.com/pingcap/docs/pull/728 -- it turns out TiDB has great support for MySQL's SQL modes.  It just doesn't enable them all by default like mysql does :-)

tiancaiamao:
LGTM @zimulala

tiancaiamao:
Ping @zimulala

zz-jason:
should we file a PR in TiDB to test this change? @morgo @tiancaiamao

morgo:
@zz-jason The corresponding TiDB PR is https://github.com/pingcap/tidb/pull/8278

NO_ZERO_DATE did break the set suite, but it's since been fixed.

tiancaiamao:
LGTM
PTAL @zz-jason @zimulala

PR #47, opened 2018-11-20
Add test case about format "a % b"

hicqu:
There is a bug about format `a % b` caused by go interprets `%` as a special things. The PR fixes this and add a test case.


ast/format_test.go
```diff
@@ -65,6 +65,9 @@ func (ts *testAstFormatSuite) TestAstFormat(c *C) {
 		{`"abc" not regexp '.*bc?'`, `"abc" NOT REGEXP ".*bc?"`},
 		{`-  4`, `-4`},
 		{`- ( - 4 ) `, `-(-4)`},
+		{`a%b`, "`a` % `b`"},
```
tiancaiamao: I test use the old parser, and this test case passes @hicqu

ast/expressions.go
```diff
@@ -137,7 +138,13 @@ func (n *BinaryOperationExpr) Format(w io.Writer) {
 	fmt.Fprint(w, " ")
 	n.Op.Format(w)
```
tiancaiamao: WriteString will not consider the % escape, it must be something in fmt.Sprint that %% => %

hicqu:
The bug is already fixed, so just keep test cases. PTAL @tiancaiamao .

tiancaiamao:
LGTM
b.t.w, this should already been covered here https://github.com/pingcap/parser/blob/master/opcode/opcode_test.go#L33

tiancaiamao:
PTAL  @XuHuaiyu

ast/format_test.go
```diff
@@ -65,6 +65,9 @@ func (ts *testAstFormatSuite) TestAstFormat(c *C) {
...
+		{`a%b+6`, "`a` % `b` + 6"},
```
ngaut: Use ``` instead of %

ast/format_test.go
```diff
@@ -65,6 +65,9 @@ func (ts *testAstFormatSuite) TestAstFormat(c *C) {
...
+		{`a%(b+6)`, "`a` % (`b` + 6)"},
```
ngaut: Ditto

ast/format_test.go
hicqu: Do you mean put "`a` % `b`" in ``? I have tried but I meet another problem: `a` and `b` are already surround by ``, so we can't use `` to surround them.

PR #48, opened 2018-11-21
add from_tso function

yu34po:
open issue in tidb:
https://github.com/pingcap/tidb/issues/8384
just  add from_tso  builtin-time function


zz-jason:
LGTM

zz-jason:
@jackysp @XuHuaiyu PTAL
/cc @shenli, @disksing

ast/functions.go
```diff
@@ -171,6 +171,7 @@ const (
 	Year             = "year"
 	YearWeek         = "yearweek"
 	LastDay          = "last_day"
+	Tso              = "tidb_parse_tso"
```
lamxTyler: Better to add a test in `parser_test.go`

PR #49, opened 2018-11-22
fix error message print verbs

lysu:
current error message has some c-style fmt verbs and cannot be using in go, so change them to go's

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/49)


tiancaiamao:
LGTM

tiancaiamao:
I'm considering move those things out from this repo, they are specific to TiDB.

PR #51, opened 2018-11-25
ast: Fix FrameBound.Accept

spongedu:
`FrameBound.Unit` should be processed instead of `FrameBound.Expr` here.


tiancaiamao:
PTAL @zz-jason

zz-jason:
@spongedu Could you add some UT for this bug?

kennytm:
UT = Unit Tests?

spongedu:
@zz-jason  Add some unittests for this.  PTAL

PR #53, opened 2018-11-27
*: support new aggregation function var_pop and var_samp

mccxj:
ref #12


zz-jason:
@XuHuaiyu PTAL

zz-jason:
@tiancaiamao @winkyao PTAL

PR #56, opened 2018-11-29
*: support AST to SQL text

leoppro:
1、add Recoverable interface 
2、add Recoverable implement example(CreateDatabaseStmt,DropDatabaseStmt,UnaryOperationExpr)
3、add stmt Recoverable test frame
4、replace the tidb dependency to github.com/leoppro/tidb temporarily,
github.com/leoppro/tidb forked from pingcap/tidb and impl the Recoverable interface for ValueExpr and ParamMarkerExpr, when it is stable,i will pull it to pingcap/tidb


leoppro:
PTAL @GregoryIan  @tiancaiamao

ast/util.go
```diff
@@ -0,0 +1,171 @@
...
+	s.insertSentence(sentence, at)
+	s.insertText(around, at)
+	sentence.Reset()
+}
+
+// LeftBracket append a "(" to SQLSentence,
+// and after that all append operations(Text, TextAround, Sentence, SentenceAround) will automatically append a comma.
+func (s *SQLSentence) LeftBracket() *SQLSentence {
+	s.Text("(")
+	s.status = sqlSentenceNearBrackets
```
kennytm: This is gonna break when there are nested call to `LeftBracket()`. 

```go
ss.LeftBracket().Text("w").Text("x")
    .LeftBracket().Text("a").Text("b").RightBracket()
    .Text("y").Text("z").RightBracket()
// "(w, x, (a, b)yz)"
```

I'm not sure who will use `LeftBracket()` but I don't see much point when this can be easily solved by an `if` in a `for` loop.

ast/util.go
```diff
@@ -0,0 +1,171 @@
...
+
+const (
+	sqlSentenceNormal sqlSentenceStatus = iota
+	sqlSentenceNearBrackets
+	sqlSentenceInBrackets
+)
+
+// SQLSentence express a sql text.
+// In order to avoid copying when string splicing occurs,
+// SQLSentence uses a linked list structure to store strings.
```
kennytm: When we will insert a Sentence in the **middle** of another Sentence, instead of the end? Appending Sentences can be easier done using a `strings.Builder` directly.

ast/util_test.go
```diff
@@ -0,0 +1,43 @@
...
+		Text("FROM").Space().TextAround("table1", "`").TextAround("WHERE", " ").
+		Text("col2=1")
+	c.Assert(ss.String(), Equals, "SELECT `col1` FROM `table1` WHERE col2=1")
+	ss2 := NewSQLSentence()
+	ss2.Text("SELECT").Space().LeftBracket().TextAround("col3", "`").TextAround("col5", "`").
+		TextAround("col6", "`").RightBracket().TextAround("FROM", " ").
+		TextAround("table2", "`").TextAround("WHERE", " ").Text("col4=").
+		Text("(").Sentence(ss).Text(")")
+	c.Assert(ss2.String(), Equals,
+		"SELECT (`col3`, `col5`, `col6`) FROM `table2` WHERE col4=(SELECT `col1` FROM `table1` WHERE col2=1)")
```
kennytm: FYI this is not a valid SQL syntax (which is another reason I question the `LeftBracket()` function).

ast/ast.go
```diff
@@ -22,9 +22,16 @@ import (
 	"github.com/pingcap/parser/types"
 )
 
+// Recoverable can be restored to sql text
+type Recoverable interface {
+	// Restore returns the sql text from ast tree
+	Restore() *SQLSentence
```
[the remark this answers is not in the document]
leoppro: make the sentence (which looks like a string buffer) as input will be hard to insert a text in the middle of another Sentence.but under your reminding,I realized that it is not necessary to insert a text in the middle,i observed the ast node structure,appending to the string will be enough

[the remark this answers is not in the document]
leoppro: if only append to string,SQLSentence is of little significance,strings.Builder is enough

leoppro:
@kennytm PTAL again

ast/ddl.go
```diff
@@ -71,6 +84,21 @@ type CreateDatabaseStmt struct {
...
 
+// Restore implements Recoverable interface.
+func (n *CreateDatabaseStmt) Restore(sb *strings.Builder) {
+	sb.WriteString("CREATE DATABASE ")
+	if n.IfNotExists {
+		sb.WriteString("IF NOT EXISTS ")
+	}
+	sb.WriteString("`")
+	sb.WriteString(n.Name)
+	sb.WriteString("`")
```
kennytm: 1. Will it properly handle names with `` ` `` in it e.g.:

    ```sql
    CREATE DATABASE `foo``bar`;
    ```

2. Do we want to handle ANSI_QUOTES (which the syntax becomes ``CREATE DATABASE "foo`bar"``) or is it considered out-of-scope?

go.mod1
```diff
@@ -14,3 +15,5 @@ require (
 	golang.org/x/net v0.0.0-20181029044818-c44066c5c816
 	golang.org/x/text v0.3.0
 )
+
+replace github.com/pingcap/tidb => github.com/leoppro/tidb v0.0.0-00000000000000-812fd5698f7b
```
kennytm: Ensure to revert this before merging?

ast/ddl.go
kennytm: @leoppro In `tidb-tools` we have https://github.com/pingcap/tidb-tools/blob/c106fbb7c074df812df90e9035a6dc59a1853565/pkg/dbutil/common.go#L396 to escape the backquotes, but I'm not sure if `n.Name` will unescape the backquotes in the first place.

For MySQL even if ANSI_QUOTES is enabled the `` `...` `` syntax can still be used, so no function is lost if we ignore ANSI_QUOTES. It's just a question of aesthetic.

tiancaiamao:
LGTM
Don't forget to address comment from kenny @leoppro

ast/ddl.go
leoppro: what about now?
https://github.com/pingcap/parser/pull/56/files#diff-ed076aa6b0e8db7b5db0c844dcc1cb33R5

go.mod1
leoppro: i will do it after parser_driver of tidb  implement  `Recoverable` interface

leoppro:
@tiancaiamao  PTAL again, I moved `IsReadOnly` into `util.go` https://github.com/pingcap/parser/pull/56/files#diff-ed076aa6b0e8db7b5db0c844dcc1cb33R21

ast/util.go
```diff
@@ -1,6 +1,35 @@
...
+// limitations under the License.
+
 package ast
 
-import "strings"
+import (
+	"strings"
+)
+
+type NodeTextCleaner struct {
```
tiancaiamao: Is `NodeTextCleaner` used in any other packages ?

leoppro: no,only string

ast/ddl.go
```diff
@@ -14,8 +14,10 @@
 package ast
 
 import (
+	"github.com/pingcap/errors"
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/types"
+	"strings"
```
XuHuaiyu: put this at line 17
and leave a blank line between the native library and the third party library

ast/ast.go
```diff
@@ -17,14 +17,22 @@ package ast
...
+	"strings"
 
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/types"
 )
 
+// Recoverable can be restored to sql text
+type Recoverable interface {
+	// Restore returns the sql text from ast tree
+	Restore(sb *strings.Builder) error
```
XuHuaiyu: It seems that Restore does not need to return an error.

We can panic in the unimplemented Restore functions.

ast/util.go
```diff
@@ -59,3 +67,29 @@ func (checker *readOnlyChecker) Enter(in Node) (out Node, skipChildren bool) {
 func (checker *readOnlyChecker) Leave(in Node) (out Node, ok bool) {
 	return in, checker.readOnly
 }
+
+type nodeTextCleaner struct {
```
XuHuaiyu: Add a comment for this struct.

ast/util.go
```diff
@@ -34,6 +36,12 @@ func IsReadOnly(node Node) bool {
 	}
 }
 
+// CleanNodeText set the text of node and all child node empty.
+func CleanNodeText(node Node) {
```
XuHuaiyu: Seems this is only used during test.
Why not put this function in the test file?

ast/stats.go
```diff
@@ -13,7 +13,11 @@
 
 package ast
 
-import "github.com/pingcap/parser/model"
+import (
+	"github.com/pingcap/errors"
+	"github.com/pingcap/parser/model"
+	"strings"
```
XuHuaiyu: ditto

ast/misc.go
```diff
@@ -16,6 +16,7 @@ package ast
 import (
 	"bytes"
 	"fmt"
+	"github.com/pingcap/errors"
```
XuHuaiyu: ditto

ast/functions.go
```diff
@@ -15,7 +15,9 @@ package ast
 
 import (
 	"fmt"
+	"github.com/pingcap/errors"
```
XuHuaiyu: ditto

ast/expressions_test.go
```diff
@@ -15,8 +15,10 @@ package ast_test
 
 import (
 	. "github.com/pingcap/check"
+	"github.com/pingcap/parser"
 	. "github.com/pingcap/parser/ast"
 	_ "github.com/pingcap/tidb/types/parser_driver"
+	"strings"
```
XuHuaiyu: ditto

ast/dml.go
```diff
@@ -14,9 +14,11 @@
 package ast
 
 import (
+	"github.com/pingcap/errors"
 	"github.com/pingcap/parser/auth"
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
+	"strings"
```
XuHuaiyu: ditto

parser_test.go
```diff
@@ -256,12 +256,28 @@ func (s *testParserSuite) RunTest(c *C, table []testCase) {
...
+		if !t.ok {
 			c.Assert(err, NotNil, comment)
+			continue
+		}
+		c.Assert(err, IsNil, comment)
+		// restore correctness test
+		for _, stmt := range stmts {
+			var sb strings.Builder
+			err := stmt.Restore(&sb)
+			if err != nil {
```
XuHuaiyu: 1. When we panic in Restore, we can use type assertion here to check whether the Node has implemented the Restore function.
2. We also need a test function like `TestRestore` to test whether the AstToSqlText result is as expected.

ast/ast.go
leoppro: For engineering consideration, we shouldn't painc to crash the process. and because of community cooperation, we add a error return to prevent unit test crash when some `Restore` is unimplemented

ast/util.go
leoppro: yes,it is only used during test,but i don't want put this function here and there

parser_test.go
leoppro: 1. I'm afraid that this approach will lead to problems in community cooperation.
2. We put it in `parser_test.go/RunTest`, and we add a unit test for one node.
like: https://github.com/pingcap/parser/pull/56/files#diff-77f9c5478bd9b2f3d33089e2ef1f4d90R106

ast/util.go
XuHuaiyu: I do not really understand what does `put this function here and there` mean?
If you mean you want to avoid duplicate definition, we'd better add a comment for this function and nodeTextCleaner to declare that they are only used for testing.

leoppro:
> It'll be nice if you can create an issue in the tidb repo to trace the process of this task.
> We can list the sub-tasks in the issue, thus the community can choose their interested tasks to participate.
> 
> This can be referred. [pingcap/tidb#7623](https://github.com/pingcap/tidb/issues/7623)

the issue in pingcap/tidb#8532, PTAL @XuHuaiyu

ast/ast.go
XuHuaiyu: We can use type assertion as I mentioned below.
From the engineering point of view, the definition of this useless return argument is unfriendly to the invoker since we must check the nonexistent error every time.

ast/util.go
leoppro: yes, we need use `CleanNodeText` during `parser_test`,`ddl_test`,`dml_test`,etc ,not only `parser_test`, i'd already add comment

tiancaiamao: That's easy:
```
util.go:
func cleanNodeText(node Node) ...
parser_test.go:
var CleanNodeText = cleanNodeText
```

`CleanNodeText` will be public to any test packages
@leoppro

ast/expressions_test.go
```diff
@@ -101,3 +104,31 @@ func (tc *testExpressionsSuite) TestExpresionsVisitorCover(c *C) {
...
+		expectSQL string
+	}{{"select ++1", "SELECT ++1"},
+		{"select --1", "SELECT --1"},
+		{"select -+1", "SELECT -+1"},
+		{"select -1", "SELECT -1"}}
+
+	for _, node := range testNodes {
+		stmt, err := parser.ParseOneStmt(node.sourceSQL, "", "")
+		comment := Commentf("source %#v", node)
+		c.Assert(err, IsNil, comment)
```
XuHuaiyu: It seems that we do not need line120 ~ line127.
We can define restoreSql as stmt.Restore().

ast/ddl.go
```diff
@@ -61,6 +64,19 @@ type DatabaseOption struct {
 	Value string
 }
 
+// Restore implements Recoverable interface.
+func (n *DatabaseOption) Restore(sb *strings.Builder) error {
+	switch n.Tp {
+	case DatabaseOptionCharset:
```
XuHuaiyu: check error here.

XuHuaiyu:
PTAL @tiancaiamao @zz-jason

ast/expressions_test.go
leoppro: i'm sorry,i don't understand you

leoppro:
> How about using the visitor model? Implement a visitor visits the AST tree, complement the sql during the traverse? Thus we don't need to introduce a `Recoverable` interface and change the AST node implementations. @leoppro

I have considered this way, it's hard to implement.
Firstly, we should use type assertion to handle different `ast.Node`, it will be complicated;
Secondly, we can‘t control visiting order for child node.

ast/util.go
tiancaiamao: For test only function should not be exported and pollute the package API.

ast/dml.go
```diff
@@ -169,6 +182,11 @@ type DeleteTableList struct {
 	Tables []*TableName
 }
 
+// Restore implements Recoverable interface.
+func (n *DeleteTableList) Restore(sb *strings.Builder) error {
+	panic("Not implemented")
```
tiancaiamao: Don't panic!

zz-jason:
> > How about using the visitor model? Implement a visitor visits the AST tree, complement the sql during the traverse? Thus we don't need to introduce a `Recoverable` interface and change the AST node implementations. @leoppro
> 
> I have considered this way, it's hard to implement.
> Firstly, we should use type assertion to handle different `ast.Node`, it will be complicated;
> Secondly, we can‘t control visiting order for child node.

Got it!

leoppro:
> I think Restore interface if necessary to implement AST to SQL
> Maybe we can add a method to Node, rather than define a Restorable interface.

some struct which isn't a Node need implement Restorable also, such as DatabaseOption

tiancaiamao:
And if you define the `Recover()` method in the node struct (the conceptual  base class), it would save you many codes to define

```
func (xx XXNode) Recover() error {
   return errors.New("not implemented")
}
``` 
@leoppro

ast/expressions_test.go
```diff
@@ -101,3 +104,40 @@ func (tc *testExpressionsSuite) TestExpresionsVisitorCover(c *C) {
...
+
+func (tc *testExpressionsSuite) createTestCase4UnaryOperationExpr() []exprTestCase {
+	return []exprTestCase{
+		{"select ++1", "SELECT ++1"},
+		{"select --1", "SELECT --1"},
+		{"select -+1", "SELECT -+1"},
+		{"select -1", "SELECT -1"},
+	}
+}
+func (tc *testExpressionsSuite) TestExpresionsRestore(c *C) {
```
XuHuaiyu: leave a blank line here.

ast/expressions_test.go
XuHuaiyu: I mean we do not need to make up `restoreSql` manually
We replace line129~136 to
``` go
restoreSql := stmt.Restore()
```

XuHuaiyu:
LGTM

XuHuaiyu:
PTAL @zz-jason @tiancaiamao

ast/ddl.go
```diff
@@ -62,6 +66,21 @@ type DatabaseOption struct {
...
+		sb.WriteString("CHARACTER SET = ")
+		sb.WriteString(n.Value)
+	case DatabaseOptionCollate:
+		sb.WriteString("COLLATE = ")
+		sb.WriteString(n.Value)
+	default:
+		return errors.New(fmt.Sprintf("invalid DatabaseOptionType: %d", n.Tp))
```
```suggestion
		return errors.Errorf("invalid DatabaseOptionType: %d", n.Tp)
```

PR #58, opened 2018-11-29
support `create fulltext index` syntax, but not that function

### LINKED ISSUE #54: support the `FULLTEXT INDEX` syntax
```go
package main

import (
	"fmt"

	"github.com/pingcap/parser"
)

func main() {
	sql := "CREATE FULLTEXT INDEX ft_ip ON tb (ip);"
	fmt.Println(parser.New().ParseOneStmt(sql, "", ""))
}
```

<nil> line 1 column 15 near " INDEX ft_ip ON tb (ip);" (total length 39)

### LINKED ISSUE #57: add .editorconfig back from tidb
https://github.com/pingcap/tidb/blob/master/.editorconfig

martianzhang:
fix #57 add .editorconfig back from pingcap/tidb
fix #54 create fulltext index syntax support


parser.y
```diff
@@ -1795,6 +1796,14 @@ CreateIndexStmtUnique:
 	{
 		$$ = true
 	}
+|	"FULLTEXT"
```
winkyao: I don't think add FULLTEXT to CreateIndexStmtUnique is proper. we need to add a new non-terminal symbol like `CreateIndexStmtFullText`

bytewatch: Yes. It's better to add new  non-terminal symbols. This can expose what kind of index the statement is creating.

martianzhang: Yeah, agree that. Now it's a trick for bypassing syntax error, not full function support.

Add new symbol needs more discussion. Except `FULLTEXT`, there are also other key options like `SPATIAL` or MySQL/TiDB will support other types soon or later.

tiancaiamao: Add FULLTEXT and SPATIAL is ok, but put them under `CreateIndexStmtUnique` is not a good idea.
Maybe we can change `CreateIndexStmtUnique` to `CreateIndexStmtOption`, and the optional value is UNIQUE/FULLTEXT/SPATIAL ?

See also https://github.com/twitter-forks/mysql/blob/master/sql/sql_yacc.yy#L2090

martianzhang: Enn, use `CreateIndexStmtOption` is much more better. But TiDB is highly depended on `pingcap/parser`, change existed struct should take an attention.

I'll close this PR, let's find a better way so or later.

PR #60, opened 2018-12-03
parser: support disable window function parser

lamxTyler:
Change the `EnableWindowFunc` to `SetWindowFunc`, so we can disable it dynamically.


lexer.go
```diff
@@ -195,9 +195,9 @@ func (s *Scanner) GetSQLMode() mysql.SQLMode {
 	return s.sqlMode
 }
 
-// EnableWindowFunc enables the scanner to recognize the keywords of window function.
-func (s *Scanner) EnableWindowFunc() {
-	s.supportWindowFunc = true
+// SetWindowFunc controls whether the scanner recognize the keywords of window function.
+func (s *Scanner) SetWindowFunc(val bool) {
```
zz-jason: s/SetWindowFunc/SetSupportWindowFunc/?

ngaut: EnableWIndowFunc should be better.

zz-jason:
LGTM

zz-jason:
@XuHuaiyu PTAL

zz-jason:
please add this PR in https://github.com/pingcap/tidb/projects/18

PR #62, opened 2018-12-04
ast: implement the Restore function of UseStmt

leoppro:
Firstly, we implement `Restore` function of `ast.UseStmt`.

Secondly, we add a test for `Restore`. For `ast.StmtNode`, we can test them in `parser_test.go` because they can restore to a complete SQL. We just need to add expect SQL to `testCase`, the `RunTest` function will help us to test it.

See: [RunTest Function](https://github.com/pingcap/parser/blob/ce4d755a8937ee6bc0e851fafdcd042ab5b1a1c1/parser_test.go#L255)

Issue: pingcap/tidb#8532

### DIFF
--- a/ast/misc.go
+++ b/ast/misc.go
@@ -341,7 +341,9 @@ type UseStmt struct {
 
 // Restore implements Recoverable interface.
 func (n *UseStmt) Restore(sb *strings.Builder) error {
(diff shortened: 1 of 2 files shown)


leoppro:
@zz-jason PTAL

PR #63, opened 2018-12-04
ast: impl Restore of ColunmNameExpr

leoppro:
Firstly, we implement `Restore` function of `ast.ColunmNameExpr` and `ast.ColunmName`.

Secondly, we add a test for `Restore`. In this case, `ast.ColunmNameExpr` can't restore to a complete SQL, so we can splice a sql to test it. We have already implemented a test function to simplify our work.

See: [TestExpresionsRestore Function](https://github.com/pingcap/parser/blob/ce4d755a8937ee6bc0e851fafdcd042ab5b1a1c1/ast/expressions_test.go#L122)

Issue: pingcap/tidb#8532

### DIFF
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -371,7 +371,16 @@ type ColumnName struct {
 
 // Restore implements Recoverable interface.
 func (n *ColumnName) Restore(sb *strings.Builder) error {
-	return errors.New("Not implemented")
+	if n.Schema.L != "" {
+		WriteName(sb, n.Schema.L)
+		sb.WriteString(".")
+	}
+	if n.Table.L != "" {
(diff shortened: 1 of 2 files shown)


ast/expressions.go
```diff
@@ -371,7 +371,16 @@ type ColumnName struct {
 
 // Restore implements Recoverable interface.
 func (n *ColumnName) Restore(sb *strings.Builder) error {
-	return errors.New("Not implemented")
+	if n.Schema.L != "" {
```
kennytm: Why `L` instead of `O`?

kennytm: I prefer `O`.

leoppro:
@zz-jason PTAL

PR #66, opened 2018-12-04
charset: make collation case insensitive

winkyao:
fix https://github.com/pingcap/tidb/issues/8304

before this PR, below case will failed.:

```
mysql> create table test(id int) ENGINE=InnoDB DEFAULT CHARSET=UTF8 COLLATE=UTF8_BIN;
ERROR 1115 (42000): Unknown character set: 'UTF8'
```


tiancaiamao:
LGTM

PR #71, opened 2018-12-07
parser: implement Restore for BetweenExpr

haplone:
* implement BetweenExpr#Restore by call Format
* add expressions_test.go#createTestCase4BetweenExpr 

pingcap/tidb#8532


haplone:
I've implement BetweenExpr Restore @leoppro

ast/expressions_test.go
```diff
@@ -131,11 +131,19 @@ func (tc *testExpressionsSuite) createTestCase4ColumnNameExpr() []exprTestCase {
 	}
 }
 
+func (tc *testExpressionsSuite) createTestCase4BetweenExpr() []exprTestCase {
+	return []exprTestCase{
+		{"select b between 1 and 2", "SELECT `b` BETWEEN 1 AND 2"},
```
leoppro: add test case:
`select b between a and b`, `select b between '' and 'b'`,

leoppro:
@tiancaiamao PTAL

ast/expressions_test.go
haplone: ok
I have added string and time test @leoppro

ast/expressions_test.go
```diff
@@ -131,6 +131,16 @@ func (tc *testExpressionsSuite) createTestCase4ColumnNameExpr() []exprTestCase {
...
+		{"select b not between 1 and 2", "SELECT `b` NOT BETWEEN 1 AND 2"},
+		{"select b between a and b", "SELECT `b` BETWEEN `a` AND `b`"},
+		{"select b between '' and 'b'", "SELECT `b` BETWEEN \"\" AND \"b\""},
```
tiancaiamao: Why the recovered `''` becomes `""`
Have you ever consider the ANSI_QUOTES sql_mode @leoppro

leoppro: it depending on the Format function implement of exprNode, what is this function used for?should we use it in Restore？

kennytm: IMO strings should always be formatted using single quotes. The current implementation uses `strconv.Quote` (https://github.com/pingcap/tidb/blob/e3f3ac24f65a89d4bfc3bba0957b2d4cfbc11e6a/types/parser_driver/value_expr.go#L109) which I think is not correct since Go's rule of escape ≠ MySQL's rule of escape.

Anyway, it is not possible to have a correct `Format()`/`Restore()` implementation across all SQL_MODEs, because [NO_BACKSLASH_ESCAPES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_no_backslash_escapes) will change how a string is interpreted (and yes TiDB supports this mode).

```sql
mysql> set sql_mode = 'NO_BACKSLASH_ESCAPES';
Query OK, 0 rows affected (0.05 sec)

mysql> select '\\\\';
+------+
| \\\\ |
+------+
| \\\\ |
+------+
1 row in set (0.00 sec)

mysql> set sql_mode = '';

mysql> select '\\\\';
+----+
| \\ |
+----+
| \\ |
+----+
1 row in set (0.00 sec)
```

leoppro: wait a minute, What effect does sql_mode have on parser? Does parser's behavior change because of sql_mode?

leoppro: if we handle sql_mode in other layers of tidb reather than parser, we don't need to consider that

leoppro: i saw the SetSQLMode in parser, should we restore node considering different modes?

kennytm: I think `Restore()` and `Format()` should carry a `FormatContext` parameter to encapsulate the SQL mode, but this kind of refactoring is out-of-scope for *this* PR.

leoppro: i have two suggest:
1. `Restore()` ouput sql with a changeless sql_mode,and we set this sql_mode when we excute it.
2. we use `SetSQLMode` to appoint sql_mode and handle it in `Restore()`

haplone:
@leoppro PTAL

leoppro:
LGTM
@kennytm PTAL

PR #74, opened 2018-12-08
parser,ast: parse ALTER TABLE .. TRUNCATE PARTITION

tiancaiamao:
Parse the ALTER TABLE .. TRUNCATE PARTITION syntax for TiDB

https://github.com/pingcap/tidb/pull/8624


parser.y
```diff
@@ -2364,7 +2371,7 @@ DropViewStmt:
 |
 	"DROP" "VIEW" "IF" "EXISTS" TableNameList RestrictOrCascadeOpt
 	{
-		$$ = &ast.DropTableStmt{IfExists: true, Tables: $5.([]*ast.TableName), IsView: true}
+		$$ = &ast.DoStmt{}
```
kennytm: Is this change expected?

crazycs520: Why change this?

model/ddl.go
```diff
@@ -80,6 +81,7 @@ var actionMap = map[ActionType]string{
 	ActionDropTablePartition:           "drop table partition",
 	ActionCreateView:                   "create view",
 	ActionModifyTableCharsetAndCollate: "modify table charset and collate",
+	ActionTruncateTablePartition:       "truncate table partition",
```
ciscoxll: s / `truncate table partition` / `alter table truncate partition` ?

PR #75, opened 2018-12-10
parser: optimize restore unit test

leoppro:
optimize restore unit test, add generic restore test logic


leoppro:
PTAL @XuHuaiyu @kennytm

leoppro:
@zz-jason PTAL

PR #76, opened 2018-12-10
ast, mysql: add window function flag and error message

lamxTyler:
Add a flag to check if an expression contains window function, and add error messages for window function.


zz-jason:
@XuHuaiyu @winoros @eurekaka PTAL

PR #77, opened 2018-12-11
types: refactor the PR #8619 by introducing the FieldType.Clone() member function

dbjoa:
Related with the TiDB PR pingcap/tidb#8641


types/field_type.go
```diff
@@ -50,6 +50,12 @@ func NewFieldType(tp byte) *FieldType {
 	}
 }
 
+// Clone returns a copy of itself.
+func (ft *FieldType) Clone() *FieldType {
```
ngaut: Do we need to do deep clone of Elems field?

dbjoa: Yes, we should do if we need to change ``Elems`` field independently. However, like ``ColumnInfo.Clone()``, we might not need in the current use case.

https://github.com/pingcap/parser/blob/4e6d047fcaae221376638de5f44c07cb6bf3eb44/model/model.go#L82-L85

PR #80, opened 2018-12-13
ignore hint when meet unknow token in hint

lysu:
Current, when we meet unknow hint token whole sql will be failure.

This PR will ignore whole hint when meet unknow hint to compatible to SQL that generate by tool and target to other database.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/80)


parser_test.go
```diff
@@ -1812,6 +1812,20 @@ func (s *testParserSuite) TestDDL(c *C) {
...
 
+func (s *testParserSuite) TestHintError(c *C) {
+	parser := New()
+	stmt, err := parser.Parse("select /*+ tidb_unknow(T1,t2) */ c1, c2 from t1, t2 where t1.c1 = t2.c1", "", "")
+	c.Assert(err, IsNil)
+	c.Assert(len(stmt[0].(*ast.SelectStmt).TableHints), Equals, 0)
+	stmt, err = parser.Parse("select /*+ tidb_unknow(T1,t2, 1) TIDB_INLJ(t1, T2) */ c1, c2 from t1, t2 where t1.c1 = t2.c1", "", "")
+	c.Assert(len(stmt[0].(*ast.SelectStmt).TableHints), Equals, 0)
+	c.Assert(err, IsNil)
+	stmt, err = parser.Parse("select1 /*+ TIDB_INLJ(t1, T2) */ c1, c2 from t1, t2 where t1.c1 = t2.c1", "", "")
```
jackysp: select1 -> select

parser_test.go
```diff
@@ -1812,6 +1812,20 @@ func (s *testParserSuite) TestDDL(c *C) {
...
+	c.Assert(err, NotNil)
+	stmt, err = parser.Parse("select /*+ TIDB_INLJ(t1, T2) */ c1, c2 fromt t1, t2 where t1.c1 = t2.c1", "", "")
```
jackysp: fromt -> from

lysu: those two tests just want to test whether error still can happen

jackysp: oh, i see.

tiancaiamao:
LGTM

parser_test.go
```diff
@@ -1812,6 +1812,20 @@ func (s *testParserSuite) TestDDL(c *C) {
 	s.RunTest(c, table)
 }
 
+func (s *testParserSuite) TestHintError(c *C) {
+	parser := New()
+	stmt, err := parser.Parse("select /*+ tidb_unknow(T1,t2) */ c1, c2 from t1, t2 where t1.c1 = t2.c1", "", "")
```
zz-jason: I think we should return a warning to tell the user that the hint is unknown and takes no effect.

lysu: I take it too, but it seems there are no elegance way to pass warn out of parser that in current situation....

so we should choose:

-  add a new error code then back code to 1064 (in mysql both error warn are 1064)
-  add a return value
-  using string.contains just like `try again later` :laughing: 

which way should we use? I think this category question maybe better disccuss before writing code :rocket:

parser_test.go
```diff
@@ -1812,9 +1812,27 @@ func (s *testParserSuite) TestDDL(c *C) {
 	s.RunTest(c, table)
 }
 
+func (s *testParserSuite) TestHintError(c *C) {
+	parser := New()
+	stmt, warns, err := parser.Parse("select /*+ tidb_unknow(T1,t2) */ c1, c2 from t1, t2 where t1.c1 = t2.c1", "", "")
+	c.Assert(err, IsNil)
+	c.Assert(len(warns), Equals, 1)
```
zz-jason: it'e better to further check the warning message

lysu: added, but it maybe useless, because current parser error message is unreadable and fix that will break many testcase, so I fire another issue to track message content problem at https://github.com/pingcap/parser/issues/86

zz-jason: got it.

PR #81, opened 2018-12-17
parser: implement Restore for BinaryOperation, WhenClause, CaseExpr and ParenthesesExpr

haplone:
* implement BinaryOperation#Restore
* implement WhenClause#Restore
* implement CaseExpr#Restore 
* implement ParenthesesExpr#Restore
* add expressions_test.go#TestBinaryOperationExpr
* add expressions_test.go#TestCaseExpr
* add expressions_test.go#TestWhenClause
* add expressions_test.go#TestParenthesesExpr

pingcap/tidb#8532


leoppro:
thank for your contribution

ast/expressions_test.go
```diff
@@ -155,3 +155,32 @@ func (tc *testExpressionsSuite) TestBetweenExprRestore(c *C) {
 	}
 	RunNodeRestoreTest(c, testCases, "select %s", extractNodeFunc)
 }
+
+func (tc *testExpressionsSuite) TestCaseExpr(c *C) {
```
leoppro: we'd better add a test for WhenClause

haplone: I've added test for WhenClause

leoppro:
LGTM @kennytm PTAL

ast/expressions.go
```diff
@@ -238,7 +256,30 @@ type CaseExpr struct {
...
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("CASE ")
+	if n.Value != nil {
+		if err := n.Value.Restore(ctx); err != nil {
+			return errors.Annotate(err, "An error occurred while restore CaseExpr.Value")
+		}
+		ctx.WritePlain(" ")
+	}
+	for idx, clause := range n.WhenClauses {
+		if idx != 0 {
```
exialin: I think here the `if` condition can be avoided by changing the position of some spaces. For example, change `"CASE "` to `"CASE“`, change `"WHEN "` of `WhenClause` to `" WHEN "` (it doesn't matter since `WhenClause` only appears in `CaseExpr`), and move the above `ctx.WritePlain(" ")` before the `if`.
This is not a big case anyway, but you can fix the bug of `CaseExpr.Format` :-)

exialin: I think here the `if` condition can be avoided by changing the position of some spaces. For example, change `"CASE "` to `"CASE“`, change `"WHEN "` of `WhenClause` to `" WHEN "` (it doesn't matter since `WhenClause` only appears in `CaseExpr`), and move the above `ctx.WritePlain(" ")` before the `if`.
This is not a big case anyway, but you can fix the bug of `CaseExpr.Format` :-)

kennytm:
@haplone  Sorry, would you mind implementing `Restore` for `ParenthesesExpr` as well? I'd like to ensure in the test cases that expressions disambiguated by precedence are properly handled:

```sql
(1 + 2) * 3
1 + 2 * 3
```

Thanks~~~ >_<

ast/expressions_test.go
```diff
@@ -155,3 +155,44 @@ func (tc *testExpressionsSuite) TestBetweenExprRestore(c *C) {
...
+	extractNodeFunc := func(node Node) Node {
+		return node.(*SelectStmt).Fields.Fields[0].Expr
+	}
+	RunNodeRestoreTest(c, testCases, "select %s", extractNodeFunc)
+}
+
+func (tc *testExpressionsSuite) TestBinaryOperationExpr(c *C) {
+	testCases := [] NodeRestoreTestCase{
+		{"'a'!=1", "'a'!=1"},
+		{"a!=1", "`a`!=1"},
```
kennytm: Please add a test case for `a <> 1`.

ast/expressions.go
haplone: I've changed CaseExpr and WhenClause
I will make another PR for CaseExpr.Format, as I want to add a test for it.

ast/expressions_test.go
haplone: I've changed as your suggests
@kennytm PTAL

ast/expressions.go
```diff
@@ -200,7 +210,15 @@ type WhenClause struct {
 
 // Restore implements Node interface.
 func (n *WhenClause) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord(" WHEN ")
```
leoppro:
```suggestion
 	ctx.WriteKeyWord("WHEN ")
```

ast/expressions.go
```diff
@@ -238,7 +256,27 @@ type CaseExpr struct {
...
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("CASE")
+	if n.Value != nil {
+		ctx.WritePlain(" ")
+		if err := n.Value.Restore(ctx); err != nil {
+			return errors.Annotate(err, "An error occurred while restore CaseExpr.Value")
+		}
+	}
+	for _, clause := range n.WhenClauses {
+		if err := clause.Restore(ctx); err != nil {
```
leoppro:
```suggestion
 		ctx.WritePlain(" ")
 		if err := clause.Restore(ctx); err != nil {
```

ast/expressions.go
leoppro: We need to comply that the output of `Restore` is a trimed string.If we don't do this, we can only read the code to confirm whether there is a space

ast/expressions.go
haplone: I've changed as your suggests
@leoppro PTAL

PR #87, opened 2018-12-19
parser: implement Restore for VariableExpr

exialin:
Implement `Restore` and unit test for `VariableExpr`.

By the way, the literal string of operator `EQ` should be `=` instead of `==`, which is fixed.


opcode/opcode.go
```diff
@@ -109,7 +109,7 @@ var opsLiteral = map[Op]string{
 	RightShift: ">>",
 	GE:         ">=",
 	LE:         "<=",
-	EQ:         "==",
+	EQ:         "=",
```
leoppro: @tiancaiamao  PTAL
Will there be a problem with this modification?

ast/expressions.go
```diff
@@ -1075,7 +1075,29 @@ type VariableExpr struct {
...
+				ctx.WriteKeyWord("GLOBAL")
+			} else {
+				ctx.WriteKeyWord("SESSION")
+			}
+			ctx.WritePlain(".")
+		}
+	} else {
+		ctx.WritePlain("@")
+	}
+	ctx.WritePlain(n.Name)
```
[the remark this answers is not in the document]
exialin: @kennytm Thanks for the clarification, but I am confused by MySQL's behaviour:
```sql
mysql> set @select = 1;

mysql> select @select;
+---------+
| @select |
+---------+
|       1 |
+---------+
1 row in set (0.00 sec)
```
It looks like reserved key words can be used as variable names. So we just wrap the variable names with backquotes?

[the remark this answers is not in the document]
kennytm: @exialin  `WriteName` already wraps them in backquotes, and adding the backquotes won't affect the result.

```sql
mysql> set @select = 1;

mysql> select @select;
+---------+
| @select |
+---------+
|       1 |
+---------+
1 row in set (0.00 sec)

mysql> select @`select`;
+-----------+
| @`select` |
+-----------+
|         1 |
+-----------+
1 row in set (0.00 sec)
```

exialin:
@kennytm Hi, I want to discuss two problems with you. After changing to `WriteName`, some tests will fail.

1. The generated ASTs of ``select @`a` `` and `select @a` are different.

The AST of ``select @`a` ``:
```
*ast.SelectStmt	&{{{{select @`a`}}} {[]} 0xc00000f590 false <nil> <nil> 0xc00000f5c0 <nil> <nil> [] <nil> <nil> none [] false false}
  *ast.FieldList	&{{} [0xc000056660]}
    *ast.SelectField	&{{`a`} 8 <nil> 0xc0000e6960  false}
      *ast.VariableExpr	&{{{} {0 0 0 0   []} 64} a false false false <nil>}
```

The AST of `select @a`:
```
*ast.SelectStmt	&{{{{select @a}}} {[]} 0xc00000f590 false <nil> <nil> 0xc00000f5c0 <nil> <nil> [] <nil> <nil> none [] false false}
  *ast.FieldList	&{{} [0xc000056660]}
    *ast.SelectField	&{{@a} 7 <nil> 0xc00009caa0  false}
      *ast.VariableExpr	&{{{} {0 0 0 0   []} 64} a false false false <nil>}
```

So they are not deeply equal to each other. In the unit test function, I have to use ``@`a` `` but not `@a` as the source SQL text. But the restored SQL text will work.

2. TiDB doesn't support queries like this while MySQL supports:
```sql
tidb> select @@global.`sql_mode`;
ERROR 1193 (HY000): Unknown system variable ''
```
The AST of `select @@global.sql_mode`:
```
*ast.SelectStmt	&{{{{select @@global.sql_mode}}} {[]} 0xc000083500 false <nil> <nil> 0xc000083530 <nil> <nil> [] <nil> <nil> none [] false false}
  *ast.FieldList	&{{} [0xc00007e600]}
    *ast.SelectField	&{{@@global.sql_mode} 7 <nil> 0xc000092aa0  false}
      *ast.VariableExpr	&{{{} {0 0 0 0   []} 64} sql_mode true true true <nil>}
```
The AST of ``select @@global.`sql_mode` ``:
```
ast.SelectStmt	&{{{{select @@global.`sql_mode`}}} {[]} 0xc000083500 false <nil> <nil> 0xc000083530 <nil> <nil> [] <nil> <nil> none [] false false}
  *ast.FieldList	&{{} [0xc00007e600]}
    *ast.SelectField	&{{} 7 <nil> 0xc000092a00 sql_mode false}
      *ast.VariableExpr	&{{{} {0 0 0 0   []} 64}  true true true <nil>}
```
As you can see the parse currently doesn't handle such case, so `Variable.Name=''`. Maybe this syntax should be added, and the test can pass only when it is supported.

kennytm:
@exialin Thanks for the throughout tests. I think both are bugs in the parser. I've filed #94 and #95. 

So we have two paths forward:

1. Use `WriteKeyWord` instead of `WriteName` and ignores variables with quoted names, or
2. Wait until those two issues are fixed

I think we should do (1) (use `WriteKeyWord`) to avoid blocking this PR on these edge cases. WDYT @tiancaiamao @leoppro

opcode/opcode.go
tiancaiamao: It fix a bug, I think.

exialin: @leoppro Maybe I'd better create another pull request to fix this first?

leoppro:
#112 id merged, please resolve conflicts and run test again @exialin

exialin:
@leoppro PTAL. 
Here's one more problem. The name of system variables are converted to lower cases:
https://github.com/pingcap/parser/blob/f4f461469cfddff6693f9a2c65379ff9fc29ea7a/parser.y#L5649-L5651

But user variables are not. So in the test cases, the names have to be case sensitive for user variables, while insensitive for system variables. Should they be consistent?

leoppro:
> @leoppro PTAL.
> Here's one more problem. The name of system variables are converted to lower cases:
> [parser/parser.y](https://github.com/pingcap/parser/blob/f4f461469cfddff6693f9a2c65379ff9fc29ea7a/parser.y#L5649-L5651)
> 
> Lines 5649 to 5651 in [f4f4614](/pingcap/parser/commit/f4f461469cfddff6693f9a2c65379ff9fc29ea7a)
> 
>  	doubleAtIdentifier 
>  	{ 
>  		v := strings.ToLower($1) 
> But user variables are not. So in the test cases, the names have to be case sensitive for user variables, while insensitive for system variables. Should they be consistent?

it doesn't matter, ignore it will be fine

leoppro:
LGTM,  PTAL @kennytm

ast/expressions_test.go
```diff
@@ -301,3 +301,22 @@ func (tc *testExpressionsSuite) TestMaxValueExprRestore(c *C) {
...
 	RunNodeRestoreTest(c, testCases, "alter table posts add partition ( partition p1 values less than %s)", extractNodeFunc)
 }
+
+func (tc *testExpressionsSuite) TestVariableExpr(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"@a>1", "@`a`>1"},
+		// {"@`aB`+1", "@`ab`+1"},
+		{"@'a':=1", "@`a`:=1"},
+		{"@`a``b`=4", "@`a``b`=4"},
+		// {`@"aBC">1`, "@`abc`>1"},
```
leoppro:
```suggestion
 		{`@"aBC">1`, "@`aBC`>1"},
```

ast/expressions_test.go
```diff
@@ -301,3 +301,22 @@ func (tc *testExpressionsSuite) TestMaxValueExprRestore(c *C) {
 	}
 	RunNodeRestoreTest(c, testCases, "alter table posts add partition ( partition p1 values less than %s)", extractNodeFunc)
 }
+
+func (tc *testExpressionsSuite) TestVariableExpr(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"@a>1", "@`a`>1"},
+		// {"@`aB`+1", "@`ab`+1"},
```
leoppro:
```suggestion
 		{"@`aB`+1", "@`aB`+1"},
```

leoppro:
@kennytm  PTAL

ast/expressions_test.go
```diff
@@ -301,3 +301,22 @@ func (tc *testExpressionsSuite) TestMaxValueExprRestore(c *C) {
...
+		{"@a>1", "@`a`>1"},
+		{"@`aB`+1", "@`aB`+1"},
+		{"@'a':=1", "@`a`:=1"},
+		{"@`a``b`=4", "@`a``b`=4"},
+		{`@"aBC">1`, "@`aBC`>1"},
+		{"@`a`+1", "@`a`+1"},
+		{"@@var", "@@`var`"},
+		{"@@global.b='foo'", "@@GLOBAL.`b`='foo'"},
+		{"@@session.'C'", "@@SESSION.`c`"},
+		{`@@local."aBc"`, "@@SESSION.`abc`"},
```
kennytm: Please add a test case for an empty variable name:.

```
@``
```

and

```
@
```

exialin: @kennytm  Done

PR #88, opened 2018-12-20
parser: implement Restore for IndexOption

feloxx:
1. implement Restore for IndexOption
2. Add indexOption goyacc code
3. Add indexOptionList goyacc code

---

Issue: [https://github.com/pingcap/tidb/issues/8532](https://github.com/pingcap/tidb/issues/8532)

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -363,28 +363,29 @@ type IndexOption struct {
 
 // Restore implements Node interface.
 func (n *IndexOption) Restore(ctx *RestoreCtx) error {
+	hasPrevOption := false
 	if n.KeyBlockSize > 0 {
 		ctx.WriteKeyWord("KEY_BLOCK_SIZE")
-		ctx.WritePlain("=")
-		ctx.WritePlainf("%d", n.KeyBlockSize)
-	}
-
-	if n.KeyBlockSize > 0 && n.Tp != model.IndexTypeInvalid {
-		ctx.WritePlain(" ")
+		ctx.WritePlainf("=%d", n.KeyBlockSize)
+		hasPrevOption = true
 	}
 
 	if n.Tp != model.IndexTypeInvalid {
+		if hasPrevOption {
+			ctx.WritePlain(" ")
+		}
 		ctx.WriteKeyWord("USING ")
 		ctx.WritePlain(n.Tp.String())
-	}
-
-	if n.Tp != model.IndexTypeInvalid && n.Comment != "" {
-		ctx.WritePlain(" ")
+		hasPrevOption = true
 	}
 
 	if n.Comment != "" {
+		if hasPrevOption {
+			ctx.WritePlain(" ")
+		}
 		ctx.WriteKeyWord("COMMENT ")
 		ctx.WriteString(n.Comment)
+		hasPrevOption = true
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


ast/ddl.go
```diff
@@ -363,7 +364,19 @@ type IndexOption struct {
 
 // Restore implements Node interface.
 func (n *IndexOption) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	if n.KeyBlockSize > 0 {
+		ctx.WriteKeyWord(" KEY_BLOCK_SIZE = ")
```
kennytm: As explained in https://github.com/pingcap/parser/pull/81#discussion_r242415210, we expect the output of `Restore()` is a trimmed string (no leading or trailing spaces).

ast/ddl.go
```diff
@@ -363,7 +364,19 @@ type IndexOption struct {
...
+		ctx.WritePlain(strconv.FormatUint(n.KeyBlockSize, 10))
```
```suggestion
		ctx.WritePlainf("%d", n.KeyBlockSize)
```

ast/ddl.go
```diff
@@ -363,7 +364,19 @@ type IndexOption struct {
...
+	}
+	if n.Tp == model.IndexTypeBtree || n.Tp == model.IndexTypeHash {
```
```suggestion
	if n.Tp != model.IndexTypeInvalid {
```

or change to a `switch` statement, so it won't silently fail even if there are suddenly new index types.

ast/ddl.go
```diff
@@ -363,7 +363,30 @@ type IndexOption struct {
 
 // Restore implements Node interface.
 func (n *IndexOption) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	if n.KeyBlockSize > 0 {
+		ctx.WriteKeyWord("KEY_BLOCK_SIZE")
+		ctx.WritePlain("=")
+		ctx.WritePlainf("%d", n.KeyBlockSize)
```
kennytm: These two can be combined

```go
ctx.WritePlainf("=%d", n.KeyBlockSize)
```

ast/ddl.go
```diff
@@ -363,7 +363,30 @@ type IndexOption struct {
...
+	}
+
+	if n.KeyBlockSize > 0 && n.Tp != model.IndexTypeInvalid {
```
kennytm: 🤔 This will repeat the condition 3 times. How about encode this in a variable?

```go
hasPrevOption := false 
if n.KeyBlockSize > 0 {
    write(...)
    hasPrevOption = true
}
if n.Tp != model.IndexTypeInvalid {
    if hasPrevOption {
        ctx.WritePlain(" ")
    }
    write(...)
    hasPrevOption = true
}
if n.Comment != "" {
    if hasPrevOption {
        ctx.WritePlain(" ")
    }
    write(...)
}
```

feloxx: I see what you mean. Let me see

parser.y
```diff
@@ -2937,7 +2939,7 @@ IndexOption:
 	{
 		$$ = &ast.IndexOption{
 			// TODO bug should be fix here!
```
leoppro: PTAL @XuHuaiyu 
what should the bug fix here?
Do we support `KeyBlockSize` function?

leoppro:
LGTM

PR #102, opened 2018-12-21
parser: implement Restore for FuncCallExpr and AggregateFuncExpr

buzzers:
implement FuncCallExpr#Restore
implement FuncCastExpr#Restore
implement AggregateFuncExpr#Restore
add functions_test.go#TestFuncCallExprRestore
add functions_test.go#TestFuncCastExprRestore
add functions_test.go#TestAggregateFuncExprRestore

pingcap/tidb#8532


leoppro:
thanks for your contribution, make sure the `make test` successful please

ast/functions.go
```diff
@@ -330,7 +330,17 @@ type FuncCallExpr struct {
 
 // Restore implements Node interface.
 func (n *FuncCallExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WritePlainf("%s(", n.FnName.O)
```
leoppro: use `WriteName` to append `FnName.O`

ast/functions.go
```diff
@@ -520,7 +562,20 @@ type AggregateFuncExpr struct {
 
 // Restore implements Node interface.
 func (n *AggregateFuncExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WritePlainf("%s(", n.F)
```
leoppro: ditto

ast/functions.go
```diff
@@ -407,7 +418,38 @@ type FuncCastExpr struct {
 
 // Restore implements Node interface.
 func (n *FuncCastExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	switch n.FunctionType {
+	case CastFunction:
+		ctx.WritePlain("CAST(")
```
leoppro: use `WriteKeyword`

parser.y
```diff
@@ -3543,10 +3543,12 @@ SimpleExpr:
 |	"CONVERT" '(' Expression "USING" StringName ')'
 	{
 		// See https://dev.mysql.com/doc/refman/5.7/en/cast-functions.html#function_convert
-		charset1 := ast.NewValueExpr($5)
-		$$ = &ast.FuncCallExpr{
-			FnName: model.NewCIStr($1),
-			Args: []ast.ExprNode{$3, charset1},
+		x := types.NewFieldType(mysql.TypeString)
```
leoppro: PTAL @tiancaiamao

ast/functions.go
buzzers: `WriteName` will wrapped in quotes, so `CONV('a',16,2)` will become `'CONV'('a',16,2)`. Is the right?

leoppro: it use "`" in default flags  =>`CONV`('a',16,2)

ast/functions.go
```diff
@@ -407,7 +418,38 @@ type FuncCastExpr struct {
 
 // Restore implements Node interface.
 func (n *FuncCastExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	switch n.FunctionType {
+	case CastFunction:
+		ctx.WriteKeyWord("CAST(")
```
leoppro:
```suggestion
 		ctx.WriteKeyWord("CAST")
 		ctx.WritePlain("(")
```

ast/functions.go
```diff
@@ -407,7 +418,38 @@ type FuncCastExpr struct {
...
+		if err := n.Expr.Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCastExpr.Expr")
+		}
+		ctx.WriteKeyWord(" AS ")
+		n.Tp.FormatAsCastType(ctx.In)
+		ctx.WritePlain(")")
+	case CastConvertFunction:
+		ctx.WriteKeyWord("CONVERT(")
```
leoppro: ditto

ast/functions.go
```diff
@@ -520,7 +562,20 @@ type AggregateFuncExpr struct {
 
 // Restore implements Node interface.
 func (n *AggregateFuncExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WritePlainf("%s(", n.F)
+	if n.Distinct {
+		ctx.WritePlain("DISTINCT ")
```
leoppro: ditto

ast/functions.go
buzzers: I tried to replace `WritePlain` with `WriteName`, but the test failed. The restored statements and source statements are not equal, and their AST is different. `CONV('a',16,2)` and ``` `CONV('a',16,2)` ``` not equal.

leoppro: it's seem like a bug of parser , i will test it and create a issue for it

buzzers: And I think there are some bug:
1. BinaryExpr here that are different from the original statements, mainly `=` and `<>`, which led to the test failure. 
2. In addition, ValueExpr string type in the original statement with encoding, the result of the restore does not contain encoding, which also led to the test failure, such as `_latin1'Müller'` will be restored to `'Müller'`.

leoppro: and reply the bug you report:
1. could you show me a fail case?
2. tidb only support utf8mb4, no matter what charset write in sql, tidb always handle them in utf8mb4, so parser ignore the charset

leoppro: > Use `AVG(test_score)` and `` `AVG`(`test_score`) ``.

got it

buzzers:
```
func (ts *testFunctionsSuite) TestXXX(c *C) {
	extractNodeFunc := func(node Node) Node {
		return node.(*SelectStmt).Fields.Fields[0].Expr
	}

	parser := parser.New()
	stmt, err := parser.ParseOneStmt("SELECT CONVERT(_latin1'Müller' USING utf8);", "", "")
	if err != nil {
		c.Fatal(err.Error())
	}
	var sb strings.Builder
	extractNodeFunc(stmt).Restore(NewRestoreCtx(DefaultRestoreFlags, &sb))
	restore := sb.String()
	if restore != "CONVERT(_latin1'Müller' USING utf8)" {
		c.Fatal(sb.String())
	}
}
```

And I get output

```
functions_test.go:117:
    c.Fatal(sb.String())
... Error: CONVERT('Müller' USING utf8)
```

parser.y
tiancaiamao: Ident.

tiancaiamao: It seems a bug. Would you please file a PR to the TiDB repo and cover this bug fix? @buzzers
and PTAL @XuHuaiyu

tiancaiamao: Fix this one `SELECT CONVERT(_latin1'Müller' USING utf8)` in TiDB if you'd like to.
https://github.com/pingcap/tidb/blob/master/expression/builtin_cast.go
A dummy implementation is acceptable as long as the new introduced `CastConvertUsingFunction` doesn't make tidb panic

tiancaiamao: Take a look here https://github.com/pingcap/parser#how-to-update-parser-for-tidb

buzzers: OK, and other problem, `AVG(test_score)` and ``` `AVG`(`test_score`) ```, is this really a bug?

ast/functions.go
leoppro: > `CONV` parse result is right, but `AVG(test_score)` and `` `AVG`(`test_score`) `` parse result is not equal. But they should be the same. @tiancaiamao

yes, it seems like a bug

parser.y
buzzers: `CastConvertUsingFunction` what is it used for, and why do you have this function?

ast/functions.go
buzzers: Well, it looks like I need to wait for this bug fix to move on to the next step.

leoppro: > Well, it looks like I need to wait for this bug fix to move on to the next step.

yes

parser.y
buzzers: @tiancaiamao This modification seems to have led to Tidb's TestInferType test case error, but I do not know how to fix it, can you give me an opinion?
Error out now typeinfer_test.go:459 `{ "convert (c_double_d using c_text_d) ", MySQL. Typelongblob, CharSet. CharsetUTF8MB4, 0, MySQL. Maxblobwidth, types. Unspecifiedlength},`.
Expected RetType not equal.

ast/functions.go
buzzers: I think I found out the reason for this bug, but I don't know how to fix it.
`parser/ast/misc.go:622`, look back here for a character to confirm whether you need to check the BUILTIN function name, so ``` `AVG`(`test_score`) ``` peek() will get "'" instead of "(".
But I don't know how to fix it, because I didn't find a way to view two characters backwards.

buzzers: I think I found out the reason for this bug, but I don't know how to fix it.
``` `AVG` ``` will be recognized as `-identifier`, so `isTokenIdentifier` method will not be called.

File: ast/functions_test.go [hunk 8]
```diff
@@ -37,3 +39,65 @@ func (ts *testFunctionsSuite) TestFunctionsVisitorCover(c *C) {
 		stmt.Accept(visitor1{})
 	}
 }
+
+func (ts *testFunctionsSuite) TestFuncCallExprRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"JSON_ARRAYAGG(attribute)", "`JSON_ARRAYAGG`(`attribute`)"},
```
leoppro: and remove the backquotes, I think the unit test will pass successfully

ast/functions_test.go
```diff
@@ -37,3 +39,65 @@ func (ts *testFunctionsSuite) TestFunctionsVisitorCover(c *C) {
...
+		{"RAND()", "`RAND`()"},
+	}
+	extractNodeFunc := func(node Node) Node {
+		return node.(*SelectStmt).Fields.Fields[0].Expr
+	}
+	RunNodeRestoreTest(c, testCases, "select %s", extractNodeFunc)
+}
+
+func TestXXXX(t *testing.T) {
+}
```
kennytm: Remove this function?

ast/functions_test.go
```diff
@@ -37,3 +39,65 @@ func (ts *testFunctionsSuite) TestFunctionsVisitorCover(c *C) {
...
+func (ts *testFunctionsSuite) TestFuncCallExprRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"JSON_ARRAYAGG(attribute)", "`JSON_ARRAYAGG`(`attribute`)"},
+		{"JSON_OBJECTAGG(attribute, value)", "`JSON_OBJECTAGG`(`attribute`, `value`)"},
+		{"ABS(-1024)", "`ABS`(-1024)"},
+		{"ACOS(3.14)", "`ACOS`(3.14)"},
+		{"CONV('a',16,2)", "`CONV`('a', 16, 2)"},
+		{"COS(PI())", "`COS`(`PI`())"},
+		{"RAND()", "`RAND`()"},
+	}
```
kennytm: Please add the TODO test cases for the following:

```sql
ADDDATE('2000-01-01', 1)
DATE_ADD('2000-01-01', INTERVAL 1 DAY)
EXTRACT(DAY FROM '2000-01-01')
GET_FORMAT(DATE, 'EUR')
POSITION('a' IN 'abc')
TRIM(BOTH 'x' FROM 'xxxyxxx')
```

parser.y
```diff
@@ -3544,9 +3544,9 @@ SimpleExpr:
 		// See https://dev.mysql.com/doc/refman/5.7/en/cast-functions.html#function_convert
 		charset1 := ast.NewValueExpr($5)
 		$$ = &ast.FuncCallExpr{
-			FnName: model.NewCIStr($1),
-			Args: []ast.ExprNode{$3, charset1},
-		}
+        	FnName: model.NewCIStr($1),
```
leoppro: could you revert this file

ast/functions.go
```diff
@@ -390,6 +515,7 @@ type CastFunctionType int
 const (
 	CastFunction CastFunctionType = iota + 1
 	CastConvertFunction
+	CastConvertUsingFunction
```
leoppro: i think we needn't add this type now

ast/functions.go
```diff
@@ -407,7 +533,46 @@ type FuncCastExpr struct {
...
+	case CastConvertFunction:
+		ctx.WriteKeyWord("CONVERT")
+		ctx.WritePlain("(")
+		if err := n.Expr.Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCastExpr.Expr")
+		}
+		ctx.WritePlain(", ")
+		n.Tp.FormatAsCastType(ctx.In)
+		ctx.WritePlain(")")
+	case CastConvertUsingFunction:
```
leoppro: ditto

ast/functions_test.go
```diff
@@ -37,3 +37,77 @@ func (ts *testFunctionsSuite) TestFunctionsVisitorCover(c *C) {
...
+		{"JSON_OBJECTAGG(attribute, value)", "JSON_OBJECTAGG(`attribute`, `value`)"},
+		{"ABS(-1024)", "ABS(-1024)"},
+		{"ACOS(3.14)", "ACOS(3.14)"},
+		{"CONV('a',16,2)", "CONV('a', 16, 2)"},
+		{"COS(PI())", "COS(PI())"},
+		{"RAND()", "RAND()"},
+
+		{"ADDDATE('2000-01-01', 1)", "ADDDATE('2000-01-01', 1)"},
+		{"DATE_ADD('2000-01-01', INTERVAL 1 DAY)", "DATE_ADD('2000-01-01', INTERVAL 1 DAY)"},
+		{"DATE_ADD('2000-01-01', INTERVAL '1 1:12:23.100000' DAY_MICROSECOND)", "DATE_ADD('2000-01-01', INTERVAL '1 1:12:23.100000' DAY_MICROSECOND)"},
```
leoppro: add test
```
DATE_ADD('2008-01-02', INTERVAL INTERVAL(1, 0, 1) DAY)
BENCHMARK(1000000, AES_ENCRYPT('text',UNHEX('F3229A0B371ED2D9441B830D21A390C3')))
SUBSTRING('Quadratically' FROM 5 FOR 3)
MASTER_POS_WAIT(@log_name, @log_pos, @timeout, @channel_name)
bit_and(distinctrow c1)
bit_and(distinct c1)
bit_and(all c1)
JSON_TYPE('[123]')
```

ast/util_test.go
```diff
@@ -63,7 +63,14 @@ type nodeTextCleaner struct {
 
 // Enter implements Visitor interface.
 func (checker *nodeTextCleaner) Enter(in Node) (out Node, skipChildren bool) {
+	//in.SetText("")
```
leoppro: remove these two lines will be fun

leoppro:
@kennytm  PTAL

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
 
 // Restore implements Node interface.
 func (n *FuncCallExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	switch n.FnName.O {
```
kennytm: Please `switch` using `n.FnName.L`, and add a test case showing that

```sql
extract(day from '1999-01-01')
```

restores correctly

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+	case "CONVERT":
+		ctx.WriteKeyWord("CONVERT")
+		ctx.WritePlain("(")
+		if err := n.Args[0].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCastExpr.Expr")
+		}
+		ctx.WritePlain(" ")
+		ctx.WriteKeyWord("USING")
+		ctx.WritePlain(" ")
```
kennytm: These 3 lines can be combined

```go
ctx.WriteKeyWord(" USING ")
```

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
 
 // Restore implements Node interface.
 func (n *FuncCallExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	switch n.FnName.O {
+	case "CONVERT":
+		ctx.WriteKeyWord("CONVERT")
```
kennytm: Move these outside the `switch`. (Similar for all cases)

```go
ctx.WriteKeyWord(n.FnName.O)
ctx.WritePlain("(")
switch n.FnName.L {
   ...
}
ctx.WritePlain(")")
return nil
```

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+		ctx.WritePlain(")")
+	case "POSITION":
+		ctx.WriteKeyWord(n.FnName.O)
+		ctx.WritePlain("(")
+		if err := n.Args[0].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCallExpr")
+		}
+		ctx.WritePlain(" ")
+		ctx.WriteKeyWord("IN")
+		ctx.WritePlain(" ")
```
kennytm: ditto combine these 3 lines

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+		ctx.WriteKeyWord(n.Args[2].(ValueExpr).GetString())
+		ctx.WritePlain(")")
+	case "EXTRACT":
+		ctx.WriteKeyWord(n.FnName.O)
+		ctx.WritePlain("(")
+		ctx.WriteKeyWord(n.Args[0].(ValueExpr).GetString())
+		ctx.WritePlain(" ")
+		ctx.WriteKeyWord("FROM")
+		ctx.WritePlain(" ")
```
kennytm: ditto combine these 3 lines

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+		ctx.WritePlain(")")
+	case "DATE_ADD":
+		ctx.WriteKeyWord(n.FnName.O)
+		ctx.WritePlain("(")
+		if err := n.Args[0].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCallExpr")
+		}
+		ctx.WritePlain(", ")
+		ctx.WriteKeyWord("INTERVAL")
+		ctx.WritePlain(" ")
```
kennytm: combine these **2** lines

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+			if err := n.Args[0].Restore(ctx); err != nil {
+				return errors.Annotatef(err, "An error occurred while restore FuncCallExpr")
+			}
+		case 2:
+			if err := n.Args[1].Restore(ctx); err != nil {
+				return errors.Annotatef(err, "An error occurred while restore FuncCallExpr")
+			}
+			ctx.WritePlain(" ")
+			ctx.WriteKeyWord("FROM")
+			ctx.WritePlain(" ")
```
kennytm: combine these 3 lines

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+		case 3:
+			switch fmt.Sprint(n.Args[2].(ValueExpr).GetValue()) {
+			case "3":
+				ctx.WriteKeyWord("TRAILING")
+			case "2":
+				ctx.WriteKeyWord("LEADING")
+			case "0", "1":
+				ctx.WriteKeyWord("BOTH")
+			}
+			ctx.WritePlain(" ")
```
kennytm: move the space into `WriteKeyWord`

ast/functions.go
```diff
@@ -330,7 +330,132 @@ type FuncCallExpr struct {
...
+			if n.Args[1].(ValueExpr).GetValue() != nil {
+				if err := n.Args[1].Restore(ctx); err != nil {
+					return errors.Annotatef(err, "An error occurred while restore FuncCallExpr")
+				}
+				ctx.WritePlain(" ")
+			}
+			ctx.WriteKeyWord("FROM")
+			ctx.WritePlain(" ")
```
kennytm: combine these 2 lines

ast/functions.go
```diff
@@ -407,7 +532,35 @@ type FuncCastExpr struct {
...
+	switch n.FunctionType {
+	case CastFunction:
+		ctx.WriteKeyWord("CAST")
+		ctx.WritePlain("(")
+		if err := n.Expr.Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCastExpr.Expr")
+		}
+		ctx.WritePlain(" ")
+		ctx.WriteKeyWord("AS")
+		ctx.WritePlain(" ")
```
kennytm: combine these 3 lines

ast/functions.go
```diff
@@ -407,7 +532,35 @@ type FuncCastExpr struct {
 
 // Restore implements Node interface.
 func (n *FuncCastExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	switch n.FunctionType {
+	case CastFunction:
+		ctx.WriteKeyWord("CAST")
+		ctx.WritePlain("(")
```
kennytm: move the `WriteKeyWord` of function name outside the `switch`

ast/functions.go
```diff
@@ -520,7 +673,22 @@ type AggregateFuncExpr struct {
 
 // Restore implements Node interface.
 func (n *AggregateFuncExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord(n.F)
+	ctx.WritePlain("(")
+	if n.Distinct {
+		ctx.WriteKeyWord("DISTINCT")
+		ctx.WritePlain(" ")
```
kennytm: combine these 2 lines

ast/functions.go
buzzers: Here I may not be able to move the function name out of switch, because the function name of the statement is not saved in the expression.

kennytm: Okay.

leoppro:
PTAL @kennytm

ast/functions.go
```diff
@@ -330,7 +330,100 @@ type FuncCallExpr struct {
...
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord(n.FnName.O)
+	ctx.WritePlain("(")
+	switch n.FnName.L {
+	case "convert":
+		if err := n.Args[0].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCastExpr.Expr")
+		}
+		ctx.WriteKeyWord(" USING ")
+		ctx.WritePlain(n.Args[1].GetType().Charset)
```
kennytm: @leoppro Should this use `WriteKeyWord`? The character set is case-insensitive.

```sql
select convert('1234' using UtF8Mb4);
```

leoppro: ok, @buzzers 
please revise this line then we will merge the pr.

buzzers: PTAL @kennytm

PR #103, opened 2018-12-21
ci: update ci script for checking code format

leoppro:
make ci check code format


leoppro:
@kennytm  @zz-jason PTAL

PR #106, opened 2018-12-21
opcode: fix literal string of operator `EQ`

exialin:
The literal string of operator `EQ` should be `=` instead of `==`.


tiancaiamao:
LGTM

PR #111, opened 2018-12-22
parser: implement Restore for TableSource and Join

exialin:
Implement `Restore` and unit tests for `TableSource` and `Join`.

The `Source` of `TableSource` can be a `SelectStmt` or a `UnionStmt`, which haven't been implemented yet, so they are not covered in the test.

---


ast/dml_test.go
```diff
@@ -139,3 +139,32 @@ func (tc *testDMLSuite) TestLimitRestore(c *C) {
...
+	RunNodeRestoreTest(c, testCases, "select * from %s", extractNodeFunc)
+}
+
+func (tc *testDMLSuite) TestJoinRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"t1 natural join t2", "`t1` NATURAL JOIN `t2`"},
+		{"t1 natural left join t2", "`t1` NATURAL LEFT JOIN `t2`"},
+		{"t1 natural right outer join t2", "`t1` NATURAL RIGHT JOIN `t2`"},
+		{"t1 straight_join t2", "`t1` STRAIGHT_JOIN `t2`"},
+		{"t1 straight_join t2 on t1.a=t2.a", "`t1` STRAIGHT_JOIN `t2` ON `t1`.`a`=`t2`.`a`"},
```
kennytm: Perhaps use an operator which allows us not to rely on #106?

ast/dml.go
```diff
@@ -85,7 +85,51 @@ type Join struct {
 
 // Restore implements Node interface.
 func (n *Join) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	if err := n.Left.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while restore Join.Left")
+	}
+	ctx.WritePlain(" ")
```
kennytm: You could skip this extra call if you attach the space to the `JOIN`/`STRAIGHT_JOIN` keyword.

```go
if condition {
    ctx.WriteKeyWord(" NATURAL")
}
if condition {
    ctx.WriteKeyWord(" LEFT")
}
ctx.WriteKeyWord(" JOIN ")
```

ast/dml_test.go
```diff
@@ -139,3 +139,32 @@ func (tc *testDMLSuite) TestLimitRestore(c *C) {
...
+		{"t1 cross join t2", "`t1` JOIN `t2`"},
+		{"t1 cross join t2 on t1.a=t2.a", "`t1` JOIN `t2` ON `t1`.`a`=`t2`.`a`"},
+		{"t1 join t2 using (b)", "`t1` JOIN `t2` USING (`b`)"},
+		{"t1 join t2 using (b,c) left join t3 on t1.a=t3.a", "`t1` JOIN `t2` USING (`b`,`c`) LEFT JOIN `t3` ON `t1`.`a`=`t3`.`a`"},
+		{"t1 natural join t2 right join t3 using (b,c)", "`t1` NATURAL JOIN `t2` RIGHT JOIN `t3` USING (`b`,`c`)"},
```
kennytm: Please add a test case involving `INNER JOIN`.

opcode/opcode.go
```diff
@@ -109,7 +109,7 @@ var opsLiteral = map[Op]string{
 	RightShift: ">>",
 	GE:         ">=",
 	LE:         "<=",
-	EQ:         "==",
+	EQ:         "=",
```
kennytm: (This should stay as part of #106, not in this PR.)

ast/dml_test.go
```diff
@@ -139,3 +139,34 @@ func (tc *testDMLSuite) TestLimitRestore(c *C) {
...
+		{"t1 natural right outer join t2", "`t1` NATURAL RIGHT JOIN `t2`"},
+		{"t1 straight_join t2", "`t1` STRAIGHT_JOIN `t2`"},
+		{"t1 straight_join t2 on t1.a>t2.a", "`t1` STRAIGHT_JOIN `t2` ON `t1`.`a`>`t2`.`a`"},
+		{"t1 cross join t2", "`t1` JOIN `t2`"},
+		{"t1 cross join t2 on t1.a>t2.a", "`t1` JOIN `t2` ON `t1`.`a`>`t2`.`a`"},
+		{"t1 inner join t2 using (b)", "`t1` JOIN `t2` USING (`b`)"},
+		{"t1 join t2 using (b,c) left join t3 on t1.a>t3.a", "`t1` JOIN `t2` USING (`b`,`c`) LEFT JOIN `t3` ON `t1`.`a`>`t3`.`a`"},
+		{"t1 natural join t2 right join t3 using (b,c)", "`t1` NATURAL JOIN `t2` RIGHT JOIN `t3` USING (`b`,`c`)"},
+		// {"t1, t2", "`t1` JOIN `t2`"},
+		// {"t1, t2, t3", "`t1` JOIN `t2` JOIN `t3`"},
```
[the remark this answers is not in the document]
exialin: @kennytm  Oh I didn't consider the parentheses. The above test case is not handled correctly. I'll fix it.
 By the way, the AST of `select * from t1, t2` is the same as that of `select * from (t1) join t2`.

exialin:
@kennytm I added a variable in `RestoreCtx` to indicate whether paretheses should be added when restoring `Join`. Now complicated cases can be restored. Please take a look if I missed something, or are there better solutions?

ast/dml.go
```diff
@@ -85,7 +85,64 @@ type Join struct {
...
+	}
+	ctx.JoinLevel++
+	if err := n.Left.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while restore Join.Left")
+	}
+	ctx.JoinLevel--
+	if n.Right == nil {
+		if ctx.JoinLevel != 0 {
+			ctx.WritePlain(")")
+		}
```
kennytm: Put these into `defer`.

```go
if ctx.JoinLevel != 0 {
    ctx.WritePlain("(")
    defer ctx.WritePlain(")")
}
```

ast/dml.go
```diff
@@ -85,7 +85,59 @@ type Join struct {
 
 // Restore implements Node interface.
 func (n *Join) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	if ctx.JoinLevel != 0 {
+		ctx.WritePlain("(")
+		defer ctx.WritePlain(")")
+	}
+	ctx.JoinLevel++
```
leoppro: could we use type asserts to decide whether to add brackets

ast/dml_test.go
```diff
@@ -139,3 +139,36 @@ func (tc *testDMLSuite) TestLimitRestore(c *C) {
 	RunNodeRestoreTest(c, testCases, "SELECT 1 %s", extractNodeFunc)
 }
 
+func (tc *testDMLSuite) TestTableSourceRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"tbl", "`tbl`"},
+		{"tbl as t", "`tbl` AS `t`"},
+	}
```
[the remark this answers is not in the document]
exialin: @leoppro This is the case for statements like `select * from (select * from a) as t` and `select * from (select * from a union select * from b) as t`. But `SelectStmt` and `UnionStmt` haven't been implemented yet, so I didn't figure out how to test these cases. Do you have any ideas?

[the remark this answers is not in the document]
leoppro: add a `TODO` comment will be nice

ast/dml.go
exialin: Actually I tried as you suggest, but it is hard to handle the following case. For example, based on current parser, the ASTs of `select * from t1, t2` and `select * from (t1) join t2` are the same:
```
*ast.SelectStmt	&{{{{select * from t1,t2}}} {[]} 0xc00000f590 false 0xc00000a720 <nil> 0xc00000f5c0 <nil> <nil> [] <nil> <nil> none [] false false}
  *ast.TableRefsClause	&{{} 0xc000148100}
    *ast.Join	&{{} {[]} 0xc000148080 0xc000052380 1 <nil> [] false false}
      *ast.Join	&{{} {[]} 0xc000052340 <nil> 0 <nil> [] false false}
        *ast.TableSource	&{{} 0xc000146090 }
          *ast.TableName	&{{} {[]}  t1 <nil> <nil> []}
      *ast.TableSource	&{{} 0xc000146120 }
        *ast.TableName	&{{} {[]}  t2 <nil> <nil> []}
  *ast.FieldList	&{{} [0xc000056660]}
    *ast.SelectField	&{{} 7 0xc0000a4d20 <nil>  false}
```
But the AST of `select * from (t1) join (t2)` is different:
```
...
    *ast.Join	&{{} {[]} 0xc000164080 0xc000164100 1 <nil> [] false false}
      *ast.Join	&{{} {[]} 0xc0000a2300 <nil> 0 <nil> [] false false}
        *ast.TableSource	&{{} 0xc000162090 }
          *ast.TableName	&{{} {[]}  t1 <nil> <nil> []}
      *ast.Join	&{{} {[]} 0xc0000a2340 <nil> 0 <nil> [] false false}
        *ast.TableSource	&{{} 0xc000162120 }
          *ast.TableName	&{{} {[]}  t2 <nil> <nil> []}
  ...
```
It is natural to find that `Join`s under the top `Join` should be wrapped with parentheses, and we don't have to think about other cases. If not, we have to consider the actual type of `TableSource`, which could be a normal table, a `SelectStmt` or a `UnionStmt`. I think that is more complicated.

leoppro: ok, detailed explanation, good job!

PR #113, opened 2018-12-23
parser: implement Restore for RowExpr

o-oops:
Issue: pingcap/tidb#8532


ast/expressions.go
```diff
@@ -999,12 +999,30 @@ type RowExpr struct {
...
 
 // Format the ExprNode into a Writer.
 func (n *RowExpr) Format(w io.Writer) {
-	panic("Not implemented")
+	fmt.Fprint(w, "ROW(")
+	for i, v := range n.Values {
+		v.Format(w)
+		if i < len(n.Values)-1 {
+			fmt.Fprint(w, ",")
+		}
```
kennytm: You could use 

```go
if i != 0 {
    fmt.Fprint(w, ",")
}
v.Format(w)
```

to avoid the subtraction (similar for `Restore`).

o-oops: Yes, I update it

ast/expressions.go
```diff
@@ -999,12 +999,30 @@ type RowExpr struct {
...
 func (n *RowExpr) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("ROW")
+	ctx.WritePlain("(")
+	for i, v := range n.Values {
+		if i != 0 {
+			ctx.WritePlain(",")
+		}
+		if err := v.Restore(ctx); err != nil {
+			return errors.Annotate(err, fmt.Sprintf("An error occurred when restore RowExpr.Values[%v]", i))
```
leoppro: use `errors.Annotatef`

ast/expressions.go
```diff
@@ -999,12 +999,30 @@ type RowExpr struct {
...
+		}
+	}
+	ctx.WritePlain(")")
+	return nil
 }
 
 // Format the ExprNode into a Writer.
 func (n *RowExpr) Format(w io.Writer) {
```
leoppro: We don't have to implement `Format`, after the `Restore` is perfected,we will remove `Format`

leoppro:
resolve conflicts please.

PR #118, opened 2018-12-23
parser: implement Restore for WildCardField, SelectField and FieldList

exialin:
Implement `Restore` and unit tests for `WildCardField`, `SelectField` and `FieldList`.

---


ast/dml_test.go
```diff
@@ -139,3 +139,41 @@ func (tc *testDMLSuite) TestLimitRestore(c *C) {
...
+	}
+	RunNodeRestoreTest(c, testCases, "SELECT %s", extractNodeFunc)
+}
+
+func (tc *testDMLSuite) TestSelectFieldRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"*", "*"},
+		{"t.*", "`t`.*"},
+		{"testdb.t.*", "`testdb`.`t`.*"},
+		{"col as a", "`col` AS `a`"},
```
kennytm: Also add a test case for expressions with the "as" e.g. `col + 1 a`.

exialin: Done.

PR #120, opened 2018-12-24
fix hint handle bug in subquery

lysu:
###  question

```
 	_, _, err = parser.Parse("SELECT 1 FROM DUAL WHERE 1 IN (SELECT /*+ DEBUG_HINT3 */ 1)", "", "")
 	c.Assert(err, IsNil)
```

will be fail in current master code

### why 

parser are try its best to reuse yySymType (https://github.com/pingcap/parser/blob/947d4ab3052f1ca608820e8dd5dd1d8c0c838fdf/yy_parser.go#L68)

what happen in here is ` /*+ DEBUG_HINT3 */` reuse `1` of `1 IN ` and doesn't reset.

we should never forgot set `$ = nil` if we want it be `nil` :rofl: 

### fix

reset `$=nil`

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -10480,6 +10480,7 @@ yynewstate:
 		{
 			yyerrok()
 			parser.lastErrorAsWarn()
+			parser.yyVAL.item = nil
 		}
 	case 979:
 		{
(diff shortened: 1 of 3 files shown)


lysu:
@zz-jason @tiancaiamao  PTAL

PR #122, opened 2018-12-24
parser: implement Restore for TruncateTableStmt

o-oops:
Issue: pingcap/tidb#8532

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -1087,7 +1087,11 @@ type TruncateTableStmt struct {
 
 // Restore implements Node interface.
 func (n *TruncateTableStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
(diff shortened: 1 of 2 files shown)


ast/ddl.go
```diff
@@ -1087,7 +1087,14 @@ type TruncateTableStmt struct {
 
 // Restore implements Node interface.
 func (n *TruncateTableStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("TRUNCATE")
+	ctx.WritePlain(" ")
+	ctx.WriteKeyWord("TABLE")
+	ctx.WritePlain(" ")
```
kennytm: These four statements can be combined

```go
ctx.WriteKeyWord("TRUNCATE TABLE ")
```

o-oops: I modified it

PR #128, opened 2018-12-26
parser: implement Restore for OrderByClause

SwanSpouse:
Implement Restore and unit tests for OrderByClause

Issue: pingcap/tidb#8532

ps: 
I'm confused about the meaning of field `OrderByClause.ForUnion`.
Does it mean using order by clause in a union query?

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -674,7 +674,16 @@ type OrderByClause struct {
 
 // Restore implements Node interface.
 func (n *OrderByClause) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("ORDER BY ")
+	for i, item := range n.Items {
+		if i != 0 {
+			ctx.WritePlain(",")
+		}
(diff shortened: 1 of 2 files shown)


kennytm:
The `ForUnion` was introduced in pingcap/tidb#491 when merging two arrays `[]*OrderByItem` in a `SelectStmt` into a single `[]*OrderByClause`, and before that PR there was no `UnionStmt`. It seems no code reads or writes to this field however. I think it is safe to ignore it in `Restore()`.

leoppro:
LGTM

ast/dml.go
```diff
@@ -674,7 +674,15 @@ type OrderByClause struct {
 
 // Restore implements Node interface.
 func (n *OrderByClause) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	for i, item := range n.Items {
```
leoppro: add keyword `ORDER BY` here

SwanSpouse: DONE PTAL

leoppro:
LGTM

PR #129, opened 2018-12-26
parser: implement Restore for HavingClause

SwanSpouse:
Implement Restore and unit tests for HavingClause

Issue: pingcap/tidb#8532

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -647,7 +647,11 @@ type HavingClause struct {
 
 // Restore implements Node interface.
 func (n *HavingClause) Restore(ctx *RestoreCtx) error {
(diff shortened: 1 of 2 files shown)


ast/dml.go
```diff
@@ -647,7 +647,10 @@ type HavingClause struct {
 
 // Restore implements Node interface.
 func (n *HavingClause) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	if err := n.Expr.Restore(ctx); err != nil {
```
leoppro: add keyword `HAVING` here

SwanSpouse: DONE PTAL

PR #131, opened 2018-12-26
add drop view ddl action

AndrewDi:
Part of View feature implement, add drop view action to handle drop view func.

Because most of `DropTable` and `DropView` code logical is the same, so I think we can use `ActionTruncateTablePartition` to pass operation type. and we can use `ActionDropTable` and `ActionDropView` for different logical part.

### DIFF
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -53,6 +53,8 @@ const (
 	ActionCreateView                   ActionType = 21
 	ActionModifyTableCharsetAndCollate ActionType = 22
 	ActionTruncateTablePartition       ActionType = 23
+	ActionDropTableOrView              ActionType = 24
+	ActionDropView                     ActionType = 25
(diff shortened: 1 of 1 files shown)


AndrewDi:
@winkyao @XuHuaiyu PTAL

model/ddl.go
```diff
@@ -53,6 +53,8 @@ const (
 	ActionCreateView                   ActionType = 21
 	ActionModifyTableCharsetAndCollate ActionType = 22
 	ActionTruncateTablePartition       ActionType = 23
+	ActionDropTableOrView              ActionType = 24
```
winkyao: Why we need ActionDropTableOrView?

AndrewDi: DropTable and DropView's common part would use this flag.

PR #135, opened 2018-12-28
parser: implement Restore for FrameBound

exialin:
Implement `Restore` and unit test for `FrameBound`.

I changed two lines in `parser.y`, though not sure it is completely right.

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -14,6 +14,8 @@
 package ast
 
 import (
+	"strings"
+
 	"github.com/pingcap/errors"
 	"github.com/pingcap/parser/auth"
 	"github.com/pingcap/parser/model"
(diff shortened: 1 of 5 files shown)


ast/dml.go
```diff
@@ -1519,7 +1519,41 @@ type FrameBound struct {
...
+			if err := n.Expr.Restore(ctx); err != nil {
+				return errors.Annotate(err, "An error occurred while restore FrameBound.Expr")
+			}
+		}
+		if n.Unit != nil {
+			// Here the Unit string should not be quoted
+			singleQuotesFlag := ctx.Flags & RestoreStringSingleQuotes
+			doubleQuotesFlag := ctx.Flags & RestoreStringDoubleQuotes
+			ctx.Flags &= ^RestoreStringSingleQuotes
+			ctx.Flags &= ^RestoreStringDoubleQuotes
```
kennytm: This doesn't seem like the correct approach. For instance, the following function call shouldn't parse either:

```sql
select date_add(current_timestamp, interval 1 'year');
```

which means the same logic got to be duplicated for FuncCallExpr.

----

I have no idea why `Unit` is considered a *string literal* expression?! It should be made a special "unit expression" based on an enumerated constant. That involves changing TiDB itself though.

exialin: Yes, I think it is reasonable. And without this change, it is harder for `FuncCallExpr` to handle this kind of cases.

ast/dml.go
```diff
@@ -1519,7 +1519,41 @@ type FrameBound struct {
...
+			ctx.WriteKeyWord("INTERVAL ")
+		}
+		if n.Expr != nil {
+			if err := n.Expr.Restore(ctx); err != nil {
+				return errors.Annotate(err, "An error occurred while restore FrameBound.Expr")
+			}
+		}
+		if n.Unit != nil {
+			// Here the Unit string should not be quoted
+			singleQuotesFlag := ctx.Flags & RestoreStringSingleQuotes
```
leoppro: My temporary plan:
```
var sb string.Builder
n.Unit.Restore(NewRestoreCtx(DefaultRestoreFlags, &sb))
ctx.WriteKeyWord(sb.String())
```

kennytm: Both is fine as a workaround, and we need to add a TODO comment improving this no matter which we choose 🙃

leoppro: TODO what? make a unit expression and change the TiDB?

kennytm: @leoppro Yes.

leoppro:
LGTM, I think we should merge this pr to advance progress. And it will be another issue about "Keyword expression". What do you think about it? @kennytm

kennytm:
@leoppro The keyword expression issue has already been addressed in #142. This PR is not blocked by keyword expressions, I'd just like a TODO to make it not forgotten.

PR #139, opened 2018-12-29
parser: implement Restore for CreateViewStmt

AndrewDi:
implement Restore for CreateViewStmt

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -821,7 +821,34 @@ type CreateViewStmt struct {
 
 // Restore implements Node interface.
 func (n *CreateViewStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("CREATE ")
+	if n.OrReplace {
+		ctx.WriteKeyWord("OR REPLACE ")
+	}
+	ctx.WritePlainf("ALGORITHM = %s ", n.Algorithm)
+	if n.Definer.CurrentUser {
+		ctx.WritePlain("DEFINER = current_user ")
+	} else {
+		ctx.WritePlain("DEFINER = ")
+		ctx.WriteName(n.Definer.Username)
+		if n.Definer.Hostname != "" {
+			ctx.WritePlain("@")
+			ctx.WriteName(n.Definer.Hostname)
+		}
+	}
+	ctx.WritePlainf(" SQL SECURITY %s ", n.Security)
+	ctx.WriteKeyWord("VIEW ")
+	ctx.WriteName(n.ViewName.Schema.O)
+	ctx.WritePlain(".")
+	ctx.WriteName(n.ViewName.Name.O)
+	ctx.WriteKeyWord(" AS ")
+
+	if err := n.Select.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while create index")
+	}
+
+	ctx.WritePlainf(" WITH %s CHECK OPTION", n.CheckOption)
+	return nil
 }
 
 // Accept implements Node Accept interface.


ast/ddl.go
```diff
@@ -821,7 +821,34 @@ type CreateViewStmt struct {
...
+	ctx.WriteName(n.ViewName.Name.O)
+	ctx.WriteKeyWord(" AS ")
+
+	if err := n.Select.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while create index")
+	}
+
+	ctx.WritePlainf(" WITH %s CHECK OPTION", n.CheckOption.String())
+	return nil
 }
```
leoppro: switch the test case on in `parser_test.go -> TestView`, eg:
```
{"create view v as select * from t", true, "CREATE VIEW `v` AS SELECT * FROM `t`"},
```

leoppro: and I think we should wait for the implement of `SelectStmt` .

leoppro: what about trying to implement `Restore` of `AlterTableStmt`?

ast/ddl.go
```diff
@@ -821,7 +821,34 @@ type CreateViewStmt struct {
 
 // Restore implements Node interface.
 func (n *CreateViewStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("CREATE ")
+	if n.OrReplace {
+		ctx.WriteKeyWord("OR REPLACE ")
+	}
+	ctx.WritePlainf("ALGORITHM = %s ", n.Algorithm.String())
```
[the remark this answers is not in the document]
AndrewDi: @kennytm View Algorithm part is particular, in MySQL, if `Algorithm == AlgorithmUndefined`, it just keep this `AlgorithmUndefined` value in view definition, If we show create a view, we will see `Algorithm=Undefined` from output.

ast/ddl.go
AndrewDi: @leoppro I think you can merge this PR after `SelectStmt` is complete.

AndrewDi: And wildcard in `SelectStmt` should be explain with column and column names,

leoppro: > And wildcard in `SelectStmt` should be explain with column and column names, If we can explain this in AST level?

no, we use `WildCardField` to mark it.

AndrewDi: That would be a problem for me...

AndrewDi:
@leoppro @kennytm PTAL

ast/ddl.go
```diff
@@ -952,7 +954,53 @@ type CreateViewStmt struct {
...
+	}
+
+	ctx.WriteKeyWord(" AS ")
+
+	if err := n.Select.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while create CreateViewStmt.Select")
+	}
+
+	if n.CheckOption != model.CheckOptionCascaded {
+		ctx.WriteKeyWord(fmt.Sprintf(" WITH %s CHECK OPTION", n.CheckOption.String()))
```
leoppro:
```suggestion
		ctx.WriteKeyWord(" WITH ", n.CheckOption.String())
		ctx.WriteKeyWord(n.CheckOption.String())
		ctx.WriteKeyWord(" CHECK OPTION")
```

AndrewDi: OK

ast/ddl.go
```diff
@@ -952,7 +952,57 @@ type CreateViewStmt struct {
...
+	ctx.WriteKeyWord("CREATE ")
+	if n.OrReplace {
+		ctx.WriteKeyWord("OR REPLACE ")
+	}
+	ctx.WriteKeyWord("ALGORITHM")
+	ctx.WritePlain(" = ")
+	ctx.WriteKeyWord(n.Algorithm.String())
+	ctx.WriteKeyWord(" DEFINER")
+	ctx.WritePlain(" = ")
+	if n.Definer.CurrentUser {
```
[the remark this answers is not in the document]
AndrewDi: OK

Issue #147, opened 2019-01-03
MySQL compatibility: `show create database if not exists` syntax


lonng:
See: https://dev.mysql.com/doc/refman/5.7/en/show-create-database.html

MySQL

```
MariaDB [(none)]> show create database b2;
+----------+----------------------------------------------------------------+
| Database | Create Database                                                |
+----------+----------------------------------------------------------------+
| b2       | CREATE DATABASE `b2` /*!40100 DEFAULT CHARACTER SET utf8mb4 */ |
+----------+----------------------------------------------------------------+
1 row in set (0.002 sec)

MariaDB [(none)]> show create database if not exists b2;
+----------+-----------------------------------------------------------------------------------------+
| Database | Create Database                                                                         |
+----------+-----------------------------------------------------------------------------------------+
| b2       | CREATE DATABASE /*!32312 IF NOT EXISTS*/ `b2` /*!40100 DEFAULT CHARACTER SET utf8mb4 */ |
+----------+-----------------------------------------------------------------------------------------+
1 row in set (0.000 sec)
```

TiDB
```
MySQL [b1]> show create batabase b2;
ERROR 1105 (HY000): line 1 column 20 near " b2" (total length 23)
MySQL [b1]> show create database b2;
+----------+-----------------------------------------------------------------+
| Database | Create Database                                                 |
+----------+-----------------------------------------------------------------+
| b2       | CREATE DATABASE `b2` /* !40100 DEFAULT CHARACTER SET utf8mb4 */ |
+----------+-----------------------------------------------------------------+
1 row in set (0.000 sec)

MySQL [b1]> show create database if not exists b2;
ERROR 1105 (HY000): line 1 column 23 near " not exists b2" (total length 37)
```

PR #148, opened 2019-01-03
parser: support `show create database if not exists` syntax

lonng:
Via: https://github.com/pingcap/parser/issues/147

See: https://dev.mysql.com/doc/refman/5.7/en/show-create-database.html


winkyao:
@crazycs520 PTAL

PR #134, opened 2018-12-28
parser: implement Restore for ColumnDef and ColumnOption

feloxx:
1. implement Restore for ColumnDef
2. implement Restore for ColumnOption
(todo Waiting for FuncCallExpr)

---

Issue: [https://github.com/pingcap/tidb/issues/8532](https://github.com/pingcap/tidb/issues/8532)

PR #149, opened 2019-01-04
parser: change restore location and implement Restore for ColumnDef and ColumnOption

feloxx:
The reason for this pr:

> Because some elements in `node` do not implement the `Restore` method.
> When they are implemented in the `Restore` mode, because of the location of the `Restore`, it is easy to have circular dependencies, so adjust.

---

1. change `Restore` loc -> util (Refer to the idea of tidb2.0.7).
2. Split `Restore test` and `Ast related Restore test`.
3. Implement the `FieldType Restore` method to prepare for future replacement of `String` and `Format` methods.

---

This pr contains https://github.com/pingcap/parser/pull/134

---

Issue: [https://github.com/pingcap/tidb/issues/8532](https://github.com/pingcap/tidb/issues/8532)


leoppro:
In `opcode\opcode.go#Op.Restore`:
```
// Restore the Op into a Writer
func (o Op) Restore(w io.Writer) error {
	if v, ok := opsLiteral[o]; ok {
		fmt.Fprint(w, v)
		return nil
	}
	return errors.Errorf("Invalid opcode type %d during restoring AST to SQL text", o)
}
```
This is a remaining problem of circular dependencies, could you rewrite this function using RestoreCtx.

ast/ddl.go
```diff
@@ -368,8 +369,52 @@ type ColumnOption struct {
...
+		ctx.WriteKeyWord("UNIQUE KEY")
+	case ColumnOptionNull:
+		ctx.WriteKeyWord("NULL")
+	case ColumnOptionOnUpdate:
+		ctx.WriteKeyWord("ON UPDATE ")
+		if err := n.Expr.Restore(ctx); err != nil {
+			return errors.Annotate(err, "An error occurred while splicing ColumnOption ON UPDATE Expr")
+		}
+	case ColumnOptionFulltext:
+		// TiDB Parser ignore the `ColumnOptionFulltext` type now
```
leoppro: what about throwing an error?

types/field_type.go
```diff
@@ -195,6 +196,87 @@ func (ft *FieldType) String() string {
 	return strings.Join(strs, " ")
 }
 
+func (ft *FieldType) Restore(ctx *restore.RestoreCtx) error {
```
leoppro: add a test function for `FieldType.Restore`

types/field_type.go
```diff
@@ -195,6 +196,87 @@ func (ft *FieldType) String() string {
...
+		if ft.Charset != "" && ft.Charset != charset.CharsetBin {
+			strs = append(strs, fmt.Sprintf("CHARACTER SET %s", ft.Charset))
+		}
+		if ft.Collate != "" && ft.Collate != charset.CharsetBin {
+			strs = append(strs, fmt.Sprintf("COLLATE %s", ft.Collate))
+		}
+	}
+
+	result := strings.Join(strs, " ")
+	ctx.WriteKeyWord(result)
```
leoppro: WriteKeyWord one by one, and i think we shouldn't `Join` strs

types/field_type.go
```diff
@@ -195,6 +196,87 @@ func (ft *FieldType) String() string {
...
+
+		switch ft.Tp {
+		case mysql.TypeEnum, mysql.TypeSet:
+			// Format is ENUM ('e1', 'e2') or SET ('e1', 'e2')
+			es := make([]string, 0, len(ft.Elems))
+			for _, e := range ft.Elems {
+				e = format.OutputFormat(e)
+				es = append(es, e)
+			}
+			suffix = fmt.Sprintf("('%s')", strings.Join(es, "','"))
```
leoppro: use RestoreCtx

types/field_type.go
```diff
@@ -195,6 +196,87 @@ func (ft *FieldType) String() string {
...
+		case mysql.TypeTimestamp, mysql.TypeDatetime, mysql.TypeDuration:
+			if isDecimalNotDefault {
+				suffix = fmt.Sprintf("(%d)", displayDecimal)
```
leoppro: ditto

leoppro:
@kennytm PTAL

ast/ddl.go
feloxx: thank you

go.mod1
```diff
@@ -14,3 +14,5 @@ require (
 	github.com/sirupsen/logrus v1.2.0
 	golang.org/x/text v0.3.0
 )
+
+replace github.com/pingcap/tidb => github.com/feloxx/tidb v0.0.0-00000000000000-4df31ec070b4c0971059a02ed576802a334378b8
```
kennytm: Please file the corresponding PR on TiDB for this.

cc @tiancaiamao This PR causes a "cyclic dependency" between TiDB and Parser, both need to be updated together. What is our conventional sequence to make it work while keeping the CI passing? (or is it too much a hassle which the whole PR should be rejected?)

opcode/opcode.go
```diff
@@ -140,9 +141,9 @@ func (o Op) Format(w io.Writer) {
 }
 
 // Restore the Op into a Writer
```
kennytm: Update the comment

parser_test.go
```diff
@@ -292,7 +293,7 @@ func (s *testParserSuite) RunRestoreTest(c *C, sourceSQLs, expectSQLs string) {
 		}
 		restoreSQLs += restoreSQL
 	}
-	comment = Commentf("restore %v; expect %v", restoreSQLs, expectSQLs)
+	comment = Commentf("fmtsql %v; expect %v", restoreSQLs, expectSQLs)
```
kennytm: Revert this replacement.

```suggestion
	comment = Commentf("restore %v; expect %v", restoreSQLs, expectSQLs)
```

types/field_type.go
```diff
@@ -195,6 +196,77 @@ func (ft *FieldType) String() string {
...
+
+	switch ft.Tp {
+	case mysql.TypeEnum, mysql.TypeSet:
+		// Format is ENUM ('e1', 'e2') or SET ('e1', 'e2')
+		es := make([]string, 0, len(ft.Elems))
+		for _, e := range ft.Elems {
+			e = format.OutputFormat(e)
+			es = append(es, e)
+		}
+		ctx.WritePlainf("('%s')", strings.Join(es, "','"))
```
kennytm: You don't need the intermediate array.

```go
ctx.WritePlain("(")
for i, e := range ft.Elems {
    if i != 0 {
        ctx.WritePlain(",")
    }
    ctx.WriteString(e)
}
ctx.WritePlain(")")
```

types/field_type.go
```diff
@@ -195,6 +196,77 @@ func (ft *FieldType) String() string {
...
+		ctx.WriteKeyWord(" ZEROFILL")
+	}
+	if mysql.HasBinaryFlag(ft.Flag) && ft.Tp != mysql.TypeString {
+		ctx.WriteKeyWord(" BINARY")
+	}
+
+	if IsTypeChar(ft.Tp) || IsTypeBlob(ft.Tp) {
+		if ft.Charset != "" && ft.Charset != charset.CharsetBin {
+			ctx.WriteKeyWord(" CHARACTER SET ")
+			ctx.WritePlainf("%s", ft.Charset)
```
kennytm: We should still use `WriteKeyWord` for character set and collation

parser_test.go
feloxx: thank you

opcode/opcode.go
feloxx: thank you

PR #159, opened 2019-01-08
parser: upload tidb version

leoppro:
upload tidb version to pingcap/tidb#971629b9477a

PR #162, opened 2019-01-08
parser: fix tidb version

leoppro:
See #159 
the bug in `testify` still remains unresolved, but I tried to construct `gomod` cache structure manually, and deceive `gomod` that `testify` is exist in my local cache, so I can run `go get xxx/tidb@master` successfully.


leoppro:
@amyangfei PTAL

amyangfei:
LGTM

PR #164, opened 2019-01-09
support select partition syntax

crazycs520:
support below syntax:
```sql
select * from t_partition partition (p0,p1)
```

this is for https://github.com/pingcap/tidb/pull/8990


crazycs520:
@tiancaiamao @ciscoxll  PTAL

PR #165, opened 2019-01-09
parser: support alter database

mia0x75:


tiancaiamao:
Maybe we can parse this statement and do nothing? @shenli @winkyao

PR #169, opened 2019-01-10
parser: fix window function interval expression type

lamxTyler:
We should not store the interval expression as a value expression, it should be kept as a `ExprNode`.
For https://github.com/pingcap/tidb/pull/9004


parser.y
```diff
@@ -4719,7 +4719,7 @@ WindowFrameStart:
 	}
 |	"INTERVAL" Expression TimeUnit "PRECEDING"
 	{
-		$$ = ast.FrameBound{Type: ast.Preceding, Expr: ast.NewValueExpr($2), Unit: ast.NewValueExpr($3),}
+		$$ = ast.FrameBound{Type: ast.Preceding, Expr: $2, Unit: ast.NewValueExpr($3),}
```
eurekaka: This is for cases where `interval` can be an expression? like `1 + 1`?

lamxTyler: Yes.

PR #172, opened 2019-01-12
 support restore table by table name syntax.

crazycs520:
eg:
```sql
admin restore table table_name
```
This is for TiDB PR: pingcap/tidb#7937


ast/misc.go
```diff
@@ -799,6 +799,9 @@ func (n *AdminStmt) Accept(v Visitor) (Node, bool) {
 
 	n = newNode.(*AdminStmt)
 	for i, val := range n.Tables {
+		if n.Tp == AdminRestoreTable {
```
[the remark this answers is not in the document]
crazycs520: In TiDB planner/core/preprocess.go, preprocessor.Leave will `handleTableName`, that is to say get `tableInfo` from `infoschema`, but in `admin restore table [table_name]` syntax, the `table_name` table already dropd, then preprocessor.Leave `handleTableName` return error that table not exists.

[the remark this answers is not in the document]
zimulala: Maybe we can add some comments

parser.y
```diff
@@ -5764,11 +5764,26 @@ AdminStmt:
...
+		}
+	}
+|	"ADMIN" "RESTORE" "TABLE" TableName
+	{
+		$$ = &ast.AdminStmt{
+			Tp: ast.AdminRestoreTable,
+			Tables: []*ast.TableName{$4.(*ast.TableName)},
+		}
+	}
+|	"ADMIN" "RESTORE" "TABLE" TableName NUM
```
zimulala: What does `NUM` mean? I think if we want use job number, we can use line5767.

crazycs520: @zimulala 
`admin restore table [table_name] [NUM]`, NUM means search how many recent ddl history jobs. Such as `admin restore table t1 20`, we will only search the recent 20 history ddl jobs to find the job which drop table t1.

crazycs520: But this check has not yet implemented in tidb. Because we currently always get all ddl history job, after implement only get [NUM] ddl history jobs, I will implement this in TiDB.

PR #175, opened 2019-01-14
improve syntax error message

lysu:
refine the syntax error message to more compatible MySQL's behavior.

try best to substring from the first token meet error.

because the parser is an independent library, it just outputs a part of the message, the user can pre-append other messages to construct own message.

e.g. In TiDB will pre-append

```
You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use,
```
before

```
line 1 column 29 near \"from t1.a = t2.a;\"
```

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/175)


lysu:
@winkyao @tiancaiamao

tiancaiamao:
LGTM

PR #177, opened 2019-01-15
parser: add show create view grammar

AndrewDi:
add show create view grammar support
ref [SHOW CREATE VIEW Syntax](https://dev.mysql.com/doc/refman/8.0/en/show-create-view.html)


AndrewDi:
@zz-jason @leoppro PTAL

PR #178, opened 2019-01-15
parser: implement Restore for ShowStmt

GunaYX:
implement Restore for `ShowStmt`
implement Restore for `UserIdentity`

While implementing `ShowStmt`, i might have found a bug in the `parser.y` file:
 https://github.com/pingcap/parser/blob/f562e1af651f8b9985ccf8c368ec89fe493db057/parser.y#L6185-L6199

If we have a "LIKE" keyword here, a `PatternLikeExpr` will be constructed and passed to `$$`. And a "WHERE" keyword will only pass `Expression` to `$$`.

And as the code written here:
https://github.com/pingcap/parser/blob/f562e1af651f8b9985ccf8c368ec89fe493db057/parser.y#L5914-L5925

A `PatternLikeExpr` will fill the `Pattern` field of the `ShowStmt` ast node, and the others(the `Expression` after "WHERE") will be assigned to `Where`. The problem appears when the `ExprNode` interface present Expression behind "WHERE" has type `PatternLikeExpr`.  The Expression will be assigned `Pattern` which may lead to problems in the future process.

pingcap/tidb#8532


GunaYX:
sorry i don't know why my branch is outdated but i did fetch upstream.

ast/dml.go
```diff
@@ -1558,7 +1558,159 @@ type ShowStmt struct {
 
 // Restore implements Node interface.
 func (n *ShowStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	RestoreOptFull := func() {
+		if n.Full {
+			ctx.WriteKeyWord("FULL ")
+		}
+	}
+	RestoreFromOrIn := func() {
```
kennytm: I don't think we need this function as we'll always write `IN`.

ast/dml.go
```diff
@@ -1558,7 +1558,159 @@ type ShowStmt struct {
 
 // Restore implements Node interface.
 func (n *ShowStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	RestoreOptFull := func() {
```
kennytm: Better use lower-case here, following the naming convention of local variables.

```suggestion
	restoreOptFull := func() {
```

ast/dml.go
```diff
@@ -1558,7 +1558,159 @@ type ShowStmt struct {
...
+		ctx.WriteKeyWord("STATS_HEALTHY")
+		if err := RestoreShowLikeOrWhereOpt(); err != nil {
+			return err
+		}
+	case ShowProfiles:
+		ctx.WriteKeyWord("PROFILES")
+	case ShowPrivileges:
+		ctx.WriteKeyWord("PRIVILEGES")
+	// ShowTargetFilterable
+	case ShowEngines, ShowDatabases, ShowCharset, ShowTables, ShowTableStatus, ShowIndex, ShowColumns, ShowWarnings, ShowErrors, ShowVariables, ShowStatus, ShowCollation, ShowTriggers, ShowProcedureStatus, ShowEvents, ShowPlugins:
```
kennytm: Maybe just use `default:` here?

tiancaiamao:
> The problem appears when the ExprNode interface present Expression behind "WHERE" has type PatternLikeExpr. The Expression will be assigned Pattern which may lead to problems in the future process.

I'm not quite understand.
Do you mean SQL in this form:

`SHOW XXX WHERE YYY LIKE ZZZ`
`SHOW XXX LIKE YYY WHERE ZZZ`? 

@GunaYX

GunaYX:
yeah, SQL in like `SHOW XXX WHERE YYY LIKE ZZZ` will lead to this bug.
sorry for my terrible description.

On 1/17/2019 11:48，tiancaiamao<notifications@github.com><mailto:notifications@github.com> wrote：

The problem appears when the ExprNode interface present Expression behind "WHERE" has type PatternLikeExpr. The Expression will be assigned Pattern which may lead to problems in the future process.

I'm not quite understand.
Do you mean SQL in this form:

SHOW XXX WHERE YYY LIKE ZZZ
SHOW XXX LIKE YYY WHERE ZZZ?

@GunaYX<https://github.com/GunaYX>

leoppro:
ci has been fixed now @GunaYX

tiancaiamao:
LGTM

GunaYX:
@leoppro  fixed the bug

ast/dml.go
```diff
@@ -1558,7 +1558,154 @@ type ShowStmt struct {
...
+			ctx.WriteKeyWord("TABLES")
+			restoreShowDatabaseNameOpt()
+		case ShowTableStatus:
+			ctx.WriteKeyWord("TABLE STATUS")
+			restoreShowDatabaseNameOpt()
+		case ShowIndex:
+			// here can be INDEX INDEXES KEYS
+			ctx.WriteKeyWord("INDEX")
+			// FROM or IN
+			ctx.WriteKeyWord(" IN ")
```
kennytm: Combine these two statements

```go
ctx.WriteKeyWord("INDEX IN ")
```

ast/dml.go
```diff
@@ -1558,7 +1558,154 @@ type ShowStmt struct {
...
+			restoreShowDatabaseNameOpt()
+		case ShowProcedureStatus:
+			ctx.WriteKeyWord("PROCEDURE STATUS")
+		case ShowEvents:
+			ctx.WriteKeyWord("EVENTS")
+			restoreShowDatabaseNameOpt()
+		case ShowPlugins:
+			ctx.WriteKeyWord("PLUGINS")
+		default:
+			return errors.New("Unkown ShowStmt type")
```
kennytm:
```suggestion
			return errors.New("Unknown ShowStmt type")
```

PR #182, opened 2019-01-16
ci: fix integration test of ci

leoppro:
download parser by commit hash in integration test.


leoppro:
PTAL
@kennytm @GregoryIan

PR #183, opened 2019-01-17
model: add version number to Column

crazycs520:
add `Version` field to `ColumnInfo` struct.
This is for https://github.com/pingcap/tidb/pull/9115


model/model.go
```diff
@@ -76,6 +83,12 @@ type ColumnInfo struct {
 	types.FieldType     `json:"type"`
 	State               SchemaState `json:"state"`
 	Comment             string      `json:"comment"`
+	// Version means the version of the column info.
+	// Version = 0: For OriginDefaultValue and DefaultValue of timestamp column will stores the default time in system time zone.
+	//              There will be bug if multiple TiDB server in different system time zone.
```
winkyao: `That is a bug if multiple TiDB servers in different system time zone.`

PR #191, opened 2019-01-27
parser: support GOARCH=386 build

hueypark:
Fix Int overflow when build on 32 bit(https://github.com/pingcap/tidb/issues/3227)

Explicit use of int64 instead of int

### DIFF
--- a/ast/misc.go
+++ b/ast/misc.go
@@ -80,8 +80,8 @@ type TypeOpt struct {
 // FloatOpt is used for parsing floating-point type option from SQL.
 // See http://dev.mysql.com/doc/refman/5.7/en/floating-point-types.html
 type FloatOpt struct {
-	Flen    int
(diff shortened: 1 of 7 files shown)


hueypark:
The reason the test fails in the CI is that the tidb and parser dependencies are circular references.
I do not have the right to solve this problem, can someone fix it?

kennytm:
Thanks for the contribution!

To fix the CI you need to submit a PR to pingcap/tidb too. Our [README](https://github.com/pingcap/parser/blob/master/README.md) has instruction about how to break the circular dependency.

hueypark:
@kennytm 
Thank you for your kind explanation.
However, while I was doing additional work on the tidb, I got a negative idea about 36bit support.
For more information, go to https://github.com/pingcap/tidb/issues/3227

PR #192, opened 2019-01-28
parser: support OPTIONALLY keyword in LOAD DATA

### LINKED ISSUE #179: Support OPTIONALLY ENCLOSED BY for LOAD DATA INFILE
It looks like the keyword `OPTIONALLY` is optional, because `LOAD DATA` seems to understand how to parse csv data that does not have enclosed quotes (this test case works the same in MySQL and TIDB):
```
echo "a,b,c" > /tmp/tmp.csv
echo '"1",2,"3"' >> /tmp/tmp.csv

DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (
 id INT NOT NULL PRIMARY KEY,
 b INT,
 c varchar(10)
);
LOAD DATA LOCAL INFILE '/tmp/tmp.csv' INTO TABLE test.t1 FIELDS TERMINATED BY ',' ENCLOSED BY '\"' IGNORE 1 LINES;

mysql> SELECT * FROM t1;
+----+------+------+
| id | b    | c    |
+----+------+------+
|  1 |    2 | 3    |
+----+------+------+
1 row in set (0.00 sec)
```

However, the following does not work in TiDB:
```
LOAD DATA LOCAL INFILE '/tmp/tmp.csv' INTO TABLE test.t1 FIELDS TERMINATED BY ',' OPTIONALLY ENCLOSED BY '\"' IGNORE 1 LINES;
```

Because of the above test-case, it looks like `OPTIONALLY` can be parsed but ignored unless someone discovers a test-case to the contrary.

imtbkcat:
fix #179 
add `OPTIONALLY` keyword in `Enclosed` statement. 

tidb PR: https://github.com/pingcap/tidb/pull/9200


tiancaiamao:
LGTM

Will there be a `Restore` method for the load data statement?  @leoppro 
There is no way to distinguish `optionally enclosed by` and `enclosed by` in this PR

leoppro:
yes, it's fine to ignore the `optionally`

parser.y
```diff
@@ -7675,7 +7676,7 @@ LocalOpt:
 	}
 
 Fields:
-     	{
+    {
```
jackysp: Please keep the align.

jackysp:
Rest LGTM

parser.y
imtbkcat: @jackysp ok

PR #193, opened 2019-01-30
parser: implement Restore for PrepareStmt ExplainStmt ExecuteStmt Dea…

eiantee:
…llocateStmt


ast/misc_test.go
```diff
@@ -203,3 +203,50 @@ func (ts *testMiscSuite) TestTableOptimizerHintRestore(c *C) {
 	}
 	RunNodeRestoreTest(c, testCases, "select /*+ %s */ * from t1 join t2", extractNodeFunc)
 }
+
+func (ts *testMiscSuite) TestExplainStmtRestore(c *C) {
```
leoppro: test `StmtNode` in `parser_test.go`

ast/misc.go
```diff
@@ -228,7 +277,20 @@ type ExecuteStmt struct {
...
+	ctx.WriteKeyWord("EXECUTE ")
+	ctx.WriteName(n.Name)
+	if n.UsingVars != nil && len(n.UsingVars) > 0 {
+		ctx.WriteKeyWord(" USING ")
+		for i, val := range n.UsingVars {
+			if i != 0 {
+				ctx.WritePlain(",")
+			}
+			if err := val.Restore(ctx); err != nil {
+				return errors.Annotate(err, fmt.Sprintf("An error occurred while restore ExecuteStmt.UsingVars index %d", i))
```
leoppro:
```suggestion
				return errors.Annotatef(err, "An error occurred while restore ExecuteStmt.UsingVars index %d", i)
```

eiantee: thank you for suggestion，it has been fixed！

ast/misc.go
```diff
@@ -134,7 +134,34 @@ type ExplainStmt struct {
...
+				return errors.Annotate(err, "An error occurred while restore ExplainStmt.ShowStmt.Column")
+			}
+		}
+		return nil
+	}
+	ctx.WriteKeyWord("EXPLAIN ")
+	if n.Analyze {
+		ctx.WriteKeyWord("ANALYZE ")
+	} else {
+		if strings.ToLower(n.Format) == ExplainFormatDOT {
```
kennytm: Change the condition to

```suggestion
		if strings.ToLower(n.Format) != ExplainFormatROW {
```

or just let `Restore()` always write out the FORMAT. MySQL supports two entirely different formats (`JSON` and `TRADITIONAL`) and I don't think there's any reason we cannot expand the list of formats in the future.

ast/misc.go
```diff
@@ -165,7 +192,23 @@ type PrepareStmt struct {
 
 // Restore implements Node interface.
 func (n *PrepareStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("PREPARE ")
+	if n.Name == "" {
+		return errors.New("An error occurred while restore PrepareStmt.Name")
```
kennytm: What's wrong with an empty name? The following is valid on MySQL 8:

```sql
PREPARE `` FROM 'select 1';
EXECUTE ``;
```

ast/misc.go
```diff
@@ -195,7 +238,13 @@ type DeallocateStmt struct {
 
 // Restore implements Node interface.
 func (n *DeallocateStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	if n.Name == "" {
+		return errors.New("An error occurred while restore DeallocateStmt.Name")
+	}
+	ctx.WriteKeyWord("DEALLOCATE ")
+	ctx.WriteKeyWord("PREPARE ")
```
kennytm: Combine these two statements.

```go
ctx.WriteKeyWord("DEALLOCATE PREPARE ");
```

ast/misc_test.go
leoppro: i think we can remove tests in `misc_test.go`

eiantee:
I have modified those as you suggested and added some empty-name test case

ast/misc.go
```diff
@@ -228,7 +268,20 @@ type ExecuteStmt struct {
 
 // Restore implements Node interface.
 func (n *ExecuteStmt) Restore(ctx *RestoreCtx) error {
-	return errors.New("Not implemented")
+	ctx.WriteKeyWord("EXECUTE ")
+	ctx.WriteName(n.Name)
+	if n.UsingVars != nil && len(n.UsingVars) > 0 {
```
```suggestion
	if len(n.UsingVars) > 0 {
```

It is unnecessary to check `n.UsingVars != nil` since `len(nil) == 0`.

Issue #140, opened 2018-12-29
Support FLUSH LOGS, FLUSH HOSTS sql commands


morgo:
Both are safe to be a no-op in TiDB.

PR #195, opened 2019-01-31
parser: add FLUSH LOGS statement

c4pt0r:
`FLUSH LOGS` is a valid flush command in MySQL.
Ref: https://github.com/datacharmer/test_db/blob/master/employees.sql#L110

Partially addresses https://github.com/pingcap/parser/issues/140


kennytm:
#238 also added the FLUSH LOGS statement. I suggest we close this in favor of #238.

zz-jason:
@c4pt0r any update? As @kennytm pointed out, can we close this PR in favor of #238?

PR #197, opened 2019-02-03
parser: implement Restore for CreateUserStmt, AlterUserStmt and DropUserStmt

exialin:
### What problem does this PR solve? 
Implement `Restore` for `CreateUserStmt`, `AlterUserStmt` and `DropUserStmt`.
And `Restore`s for `UserSpec`, `UserIdentity` and `AuthOption` are implemented to accomplish this task.

Issue: https://github.com/pingcap/tidb/issues/8532

### Check List 

Tests 

 - Unit test

PR #198, opened 2019-02-03
parser: implement Restore for PrivElem, GrantStmt and RevokeStmt

exialin:
Implement `Restore` for `PrivElem`, `GrantStmt` and `RevokeStmt`.
Waiting for some helper `Restore`s implemented in #197 .

### Check List 

Tests 

 - Unit test


ast/misc.go
```diff
@@ -93,6 +93,18 @@ type AuthOption struct {
 	// TODO: support auth_plugin
 }
 
+// Restore implements Node interface.
+func (n *AuthOption) Restore(ctx *RestoreCtx) error {
```
tiancaiamao: Restore "IDENTIFIED WITH" will get "IDENTIFIED BY" ?

tiancaiamao: And there is no test to cover `AutoOption` ?

exialin: @tiancaiamao 
1. According to `parser.y`, it looks like syntax `IDENTIFIED WITH string` has no effect:
https://github.com/pingcap/parser/blob/63e2db7dbaddf946dc28c17077251b217fd5da26/parser.y#L7302-L7319

2. To test `Restore` of `AuthOption` (and `UserSpec`, `UserIdentity`), and to follow the testing routine of other types, it should be promoted as a `Node` and add `Accept` for it. Is that suitable?

tiancaiamao:
LGTM

tiancaiamao:
PTAL @leoppro

ast/misc.go
```diff
@@ -975,7 +1111,40 @@ type GrantStmt struct {
...
+	}
+	ctx.WriteKeyWord(" ON ")
+	switch n.ObjectType {
+	case ObjectTypeNone:
+		// do nothing
+	case ObjectTypeTable:
+		ctx.WriteKeyWord("TABLE ")
+	default:
+		return errors.New("Unsupported object type of GrantStmt")
+	}
```
kennytm: Please implement `Restore` on `ObjectTypeType` to avoid duplicating this `switch` twice.

exialin: Modified. PTAL

ast/misc.go
```diff
@@ -1250,6 +1308,19 @@ const (
 	ObjectTypeTable
 )
 
+// Restore implements Node interface.
+func (n ObjectTypeType) Restore(ctx *RestoreCtx) error {
+	switch n {
+	case ObjectTypeNone:
+		// do nothing
+	case ObjectTypeTable:
+		ctx.WriteKeyWord("TABLE ")
```
```suggestion
		ctx.WriteKeyWord("TABLE")
```

This also means line 1389 and 1447 would need to look like this 🤷‍♂️:

```go
if n.ObjectType != ObjectTypeNone {
	if err := n.ObjectType.Restore(ctx); err != nil {
		return errors.Annotate(err, "An error occurred while restore RevokeStmt.ObjectType")
	} 
	ctx.WritePlain(" ")
}
```

exialin: Yes, this is better. Should I create a new pull request to fix it? Sorry for the trouble.

kennytm: @exialin Yes please. Thanks.

PR #200, opened 2019-02-05
types: Fix description for year type

spongedu:
1. Fix issue https://github.com/pingcap/tidb/issues/9188, type `year` should be demonstrated with `display width`.

2. Migrate type related tests from TiDB repo to parser repo.

add an suffix for type `year` function `FieldType. CompactStr`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


tiancaiamao:
LGTM

types/field_type_test.go
```diff
@@ -0,0 +1,166 @@
+// Copyright 2015 PingCAP, Inc.
```
zimulala: s/2015/2019

spongedu: Done

spongedu: @zimulala  PTAL :)

tiancaiamao:
PTAL @zimulala

zz-jason:
@spongedu integration test failed, some test case in the TiDB repo should be changed to make it pass.

tiancaiamao:
Wait for this one https://github.com/pingcap/tidb/pull/9253

spongedu:
@zz-jason @tiancaiamao  This pr and https://github.com/pingcap/tidb/pull/9253 seems depend on each other. There seems be a deadlock.....

kennytm:
The proper action, according to the README, is 

1. Merge this PR
2. Update the `parser` dependency in pingcap/tidb#9253
3. Merge pingcap/tidb#9253

spongedu:
@kennytm Agree.  I suppose we should find a way to walk around the test fail.

PR #201, opened 2019-02-06
parser: check invalid column define option for generated columns 

spongedu:
As is described in [MySQL Reference](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html), generated column definitions have this syntax below:
```sql
col_name data_type [GENERATED ALWAYS] AS (expr)
  [VIRTUAL | STORED] [NOT NULL | NULL]
  [UNIQUE [KEY]] [[PRIMARY] KEY]
  [COMMENT 'string']

```
some column options such as `DEFAULT`, `AUTO_INCREMENT`, `ON UPDATE` are not supported for generate column definition.  A parser error should be raised with these options.

Add an check in parser when `ColumnDef` is created for generated columns.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


spongedu:
@kennytm @morgo PTAL

ast/ddl.go
```diff
@@ -354,6 +354,14 @@ const (
 	ColumnOptionCollate
 )
 
+var (
+	invalidOption4GeneratedColumn = map[ColumnOptionType]struct{}{
```
```suggestion
	invalidOptionForGeneratedColumn = map[ColumnOptionType]struct{}{
```

ast/ddl.go
```diff
@@ -672,6 +680,22 @@ func (n *ColumnDef) Accept(v Visitor) (Node, bool) {
 	return v.Leave(n)
 }
 
+func (n *ColumnDef) Validate() bool {
```
kennytm: Document the `Validate()` method. Or could this be a private method (rename as `validate()`)?

ast/ddl.go
```diff
@@ -672,6 +680,22 @@ func (n *ColumnDef) Accept(v Visitor) (Node, bool) {
...
+	generatedCol := false
+	illegalOpt4gc := false
+	for _, opt := range n.Options {
+		if opt.Tp == ColumnOptionGenerated {
+			generatedCol = true
+		}
+		_, found := invalidOption4GeneratedColumn[opt.Tp]
+		illegalOpt4gc = illegalOpt4gc || found
+		if illegalOpt4gc {
+			break
```
kennytm: Remove the `break`. The program flow is wrong if those "invalid options" appeared before GENERATED.

```sql
create table t1 (a int generated always as (10) default 5);
create table t2 (a int default 5 generated always as (10));
```

The latter is currently *accepted* by TiDB, because we don't distinguish GENERATED options. Alternatively we could separate the column definitions as [MySQL described](https://dev.mysql.com/doc/refman/8.0/en/create-table.html).

```
column_definition:
    data_type [NOT NULL | NULL] [DEFAULT {literal | (expr)} ]
      [AUTO_INCREMENT] [UNIQUE [KEY]] [[PRIMARY] KEY]
      [COMMENT 'string']
      [COLLATE collation_name]
      [COLUMN_FORMAT {FIXED|DYNAMIC|DEFAULT}]
      [STORAGE {DISK|MEMORY}]
      [reference_definition]
      [check_constraint_definition]
  | data_type
      [COLLATE collation_name]
      [GENERATED ALWAYS] AS (expr)
      [VIRTUAL | STORED] [NOT NULL | NULL]
      [UNIQUE [KEY]] [[PRIMARY] KEY]
      [COMMENT 'string']
      [reference_definition]
      [check_constraint_definition]
```

parser.y
```diff
@@ -1550,7 +1550,12 @@ ColumnDefList:
 ColumnDef:
 	ColumnName Type ColumnOptionListOpt
 	{
-		$$ = &ast.ColumnDef{Name: $1.(*ast.ColumnName), Tp: $2.(*types.FieldType), Options: $3.([]*ast.ColumnOption)}
+		colDef := &ast.ColumnDef{Name: $1.(*ast.ColumnName), Tp: $2.(*types.FieldType), Options: $3.([]*ast.ColumnOption)}
+		if !colDef.Validate() {
+			yylex.AppendError(yylex.Errorf(""))
```
kennytm: There should be a meaningful message instead of `""`

ast/ddl.go
spongedu: ok

ast/ddl.go
spongedu: OK. I'll remove the break first and separate the column definitions later. maybe in another pr

parser.y
spongedu: ok

ast/ddl.go
spongedu: ok

spongedu:
@kennytm  PTAL

kennytm:
@spongedu Thanks for contribution. For the build-integration failure, please file a PR to pingcap/tidb to update the parser dependency, and change [this test](https://github.com/pingcap/tidb/blob/7611a7dae8d9d1779b9ae3f88e224b33557a92c1/ddl/db_integration_test.go#L546-L562) to test for parse failure. cc @winkyao pingcap/tidb#8620.

spongedu:
@kennytm  ok

PR #202, opened 2019-02-12
parser: remove dependency on tidb/util/hack package

tiancaiamao:

Go compiler has a optimize that

```
map[string(b)].. will not cause byte to string allocation
```

So the util/hack here is useless.

Fix https://github.com/pingcap/tidb/pull/9230

remove tidb/util/hack.String

### Check List 

Tests 

 - No code


tiancaiamao:
PTAL @jackysp @lysu

PR #203, opened 2019-02-12
parser: minor fix of #198

exialin:
Minor fix remained of PR #198.


exialin:
@kennytm PTAL

PR #205, opened 2019-02-13
model: add new field to explict charset&collation in ColumnInfo

xiekeyi98:
To solve https://github.com/pingcap/tidb/pull/9113 , we need change ColumnInfo a little.

Add a field `ExplictedCollation` .

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -103,8 +103,8 @@ func (c *ColumnInfo) IsGenerated() bool {
 	return len(c.GeneratedExprString) != 0
 }
 
-// IsExplictedCollation returns true if the charset or collation are explicitly assigned
-func (c *ColumnInfo) IsExplictedCollation() bool {
+// HasExplictedCollation returns true if the charset or collation are explicitly assigned
+func (c *ColumnInfo) HasExplictedCollation() bool {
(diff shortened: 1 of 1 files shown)


model/model.go
```diff
@@ -83,6 +83,7 @@ type ColumnInfo struct {
 	types.FieldType     `json:"type"`
 	State               SchemaState `json:"state"`
 	Comment             string      `json:"comment"`
+	ExplictedCollation  bool        `json:"explicted_collation"`
```
xiekeyi98: Any other **better variable name** ? I think this name is not good enough, but I cannot find a better name(avoid too long).
And I modeled the variable name of MYSQL(In MYSQL, this name is `column_has_explicit_collation`).

kennytm: "Explicit" is an adjective not a verb, there's no need to add the -ed suffix.

```suggestion
	ExplicitCollation   bool        `json:"explicit_collation"`
```

xiekeyi98: Done,thanks for your help.

xiekeyi98:
I make something wrong.
Sorry for bother everyone.
This PR will be closed because it is not necessary now.

PR #206, opened 2019-02-13
parser: finish the restore function of all ast node

leoppro:
Most of the restore function has been completed. but some test cases didn't enable.

this pr switch on all of the test cases in `parser_test.go` and fix some bug exposed in testing.

**after pingcap/tidb#9302 merged, i will update the version of tidb dependency, then all the UT will pass**

Issue: pingcap/tidb#8532

### Check List 

Tests 

 - Integration test
switch on many tests in `parser_test.go`


leoppro:
@GregoryIan @kennytm PTAL

ast/ddl.go
```diff
@@ -1280,13 +1281,15 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
 		}
 	case TableOptionStatsPersistent:
 		// TODO: not support
+		ctx.WriteKeyWord("STATS_PERSISTENT = DEFAULT")
```
```suggestion
		ctx.WriteKeyWord("STATS_PERSISTENT")
		ctx.WritePlain(" = ")
		ctx.WriteKeyWord("DEFAULT")
```

ast/ddl.go
```diff
@@ -1280,13 +1281,15 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
...
 		ctx.WritePlain(" /* TableOptionStatsPersistent is not supported */ ")
 	case TableOptionShardRowID:
 		ctx.WriteKeyWord("SHARD_ROW_ID_BITS ")
 		ctx.WritePlain("= ")
 		ctx.WritePlainf("%d", n.UintValue)
 	case TableOptionPackKeys:
 		// TODO: not support
+		ctx.WriteKeyWord("PACK_KEYS = DEFAULT")
```
kennytm: (ditto)

ast/ddl.go
```diff
@@ -1280,13 +1281,15 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
 		}
 	case TableOptionStatsPersistent:
 		// TODO: not support
+		ctx.WriteKeyWord("STATS_PERSISTENT = DEFAULT")
 		ctx.WritePlain(" /* TableOptionStatsPersistent is not supported */ ")
```
kennytm: Do we still need this?

ast/misc.go
```diff
@@ -1232,9 +1246,7 @@ type PrivElem struct {
 func (n *PrivElem) Restore(ctx *RestoreCtx) error {
 	switch n.Priv {
 	case 0:
-		// Do nothing for types that have no effect.
-		// Actually this should not happen since there is no way to determine its type.
-		return errors.New("Cannot determine privilege type")
+		ctx.WritePlainf("/* UNSUPPORTED TYPE */")
```
```suggestion
		ctx.WritePlain("/* UNSUPPORTED TYPE */")
```

ast/ddl.go
leoppro: yes, it's necessary.
The following two sql corresponds to two different ast tree:
```
create table t (c int) STATS_PERSISTENT = default

create table t (c int);
```
To ensure that the two ast trees are equal, we have to add `STATS_PERSISTENT = default`

kennytm: I mean is the comment `/* TableOptionStatsPersistent is not supported */` still needed to be printed out?

leoppro: yes, It's used to remind people that `STATS_PERSISTENT = default` is not the right result.

GregoryIan:
LGTM

leoppro:
@kennytm  PTAL

PR #208, opened 2019-02-14
parser: fix lexer that treat 9eTSs as a float

tiancaiamao:

Scan 9eTSs should get an identifier token

Fix https://github.com/pingcap/tidb/issues/8983

### Check List 

Tests 

 - Unit test

### DIFF
--- a/lexer_test.go
+++ b/lexer_test.go
@@ -113,6 +113,7 @@ func (s *testLexerSuite) TestLiteral(c *C) {
 		{"132.313", decLit},
 		{"132.3e231", floatLit},
 		{"132.3e-231", floatLit},
+		{"001e-12", floatLit},
 		{"23416", intLit},
 		{"123test", identifier},
 		{"123" + string(unicode.ReplacementChar) + "xxx", identifier},
@@ -232,6 +233,8 @@ func (s *testLexerSuite) TestIdentifier(c *C) {
 		{"1_x", "1_x"},
 		{"0_x", "0_x"},
 		{replacementString, replacementString},
+		{"9e", "9e"},
+		{"9eTSs", "9eTSs"},
 		{fmt.Sprintf("t1%cxxx", 0), "t1"},
 	}
 	l := &Scanner{}
--- a/lexer_test.go
+++ b/lexer_test.go
@@ -127,7 +127,7 @@ func (s *testLexerSuite) TestLiteral(c *C) {
 		{"\\N", null},
 		{".*", int('.')},       // `.`, `*`
 		{".1_t_1_x", int('.')}, // `.`, `1_t_1_x`
-		{"9e9e", intLit},       // 9e9e = 9e9 + e
+		{"9e9e", floatLit},     // 9e9e = 9e9 + e
 		// Issue #3954
 		{".1e23", floatLit}, // `.1e23`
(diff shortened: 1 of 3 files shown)


tiancaiamao:
PTAL @winoros @lysu

lexer_test.go
```diff
@@ -232,6 +233,8 @@ func (s *testLexerSuite) TestIdentifier(c *C) {
 		{"1_x", "1_x"},
 		{"0_x", "0_x"},
 		{replacementString, replacementString},
+		{"9e", "9e"},
+		{"9eTSs", "9eTSs"},
```
lysu: maybe need a test-case for `9e9e` :shipit:

kennytm: `9e9e` is not recognized as an single identifier though.

```sql
mysql> select 1 9e9e;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '9e9e' at line 1

mysql> select 9e9e;
+------------+
| e          |
+------------+
| 9000000000 |
+------------+
1 row in set (0.00 sec)
```

By extension, could we also check the following:

```sql
mysql> select 1 0xaaahhh;
+----------+
| 0xaaahhh |
+----------+
|        1 |
+----------+
1 row in set (0.00 sec)

mysql> select 1 0b123;
+-------+
| 0b123 |
+-------+
|     1 |
+-------+
1 row in set (0.00 sec)

mysql> select 1 0x3p4;
+-------+
| 0x3p4 |
+-------+
|     1 |
+-------+
1 row in set (0.00 sec)
```

tiancaiamao: 9e9e = 9e9 + e @lysu

tiancaiamao:
```
mysql> select 9e9e;
+------------+
| e          |
+------------+
| 9000000000 |
+------------+
1 row in set (0.00 sec)

mysql> select 9e9 e;
+------------+
| e          |
+------------+
| 9000000000 |
+------------+
1 row in set (0.00 sec)
```

@kennytm

tiancaiamao: Test added @lysu

tiancaiamao:
```
mysql> create table t (0b123 int);
Query OK, 0 rows affected (0.19 sec)
```

You're right ... @kennytm 
0b123 is also an identifier!

tiancaiamao:
The rule is weird...

```
0x7fz != 0x7f + z
0b12 != 0b1 + 2
9e9a = 9e9 + a
```

I have some updates, PTAL @kennytm @lysu

parser_test.go
```diff
@@ -2040,7 +2040,7 @@ func (s *testParserSuite) TestType(c *C) {
 		// for bit
 		{"select 0b01, 0b0, b'11', B'11'", true, "SELECT b'1',b'0',b'11',b'11'"},
 		{"select 0B01", false, ""},
-		{"select 0b21", false, ""},
+		// {"select 0b21", false, ""}, 0b21 is an identifier, the statement could parse.
```
kennytm: 🤔 `0B01` is also an identifier. I think both lines 2042 and 2043 should have the same treatment (either both can be parsed or both need to be commented out).

tiancaiamao: Fixed

PR #209, opened 2019-02-14
plugin: add `flush tidb plugins [plugin-name,..]` support

lysu:

refer https://github.com/pingcap/tidb/pull/9136

For reasons of efficiency， plugin often caches some metadata / config in each tidb nodes, and we need a mechanism to update plugin related metadata / config for each of them.

After discuss with @tennix  @tiancaiamao @jackysp, 

we decide to use "mutator notify plugin data changed and plugin pull new configuration pattern".

- make plugin be flushable
- mutator will call `flush tidb plugins ${plugin-name,..}` in one of tidb-servers
- plugin framework notify all plugin in every tidb-server node via etcd
- plugin pull new configuration into cache.

so, this PR address (1) to add `flush tidb plugins [name,...]` in parser, old `flush status` has some special means in mysql, so we choosen new one.

### Check List 

Tests 

 - Unit test

Code changes

 - New tidb keyword

Side effects

 - N/A

Related changes

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/209)


tiancaiamao:
How about `flush tidb plugin [plugin-name]`, as we already have the `tidb` keywords?

tiancaiamao:
LGTM

PR #211, opened 2019-02-16
parser: parse the fraction of CURRENT_TIMESTAMP default value

### LINKED ISSUE #186: CURRENT_TIMESTAMP FuncCallExpr should support args
like the question on so :https://stackoverflow.com/questions/23671222/mysql-default-value-for-timestamp3

> mysql> create table tbl_so_q23671222_1( ts timestamp(3) default now() );
ERROR 1067 (42000): Invalid default value for 'ts'
mysql> create table tbl_so_q23671222_1( ts timestamp(3) default now(3) );
Query OK, 0 rows affected (0.59 sec)
mysql> create table tbl_so_q23671222_2( ts timestamp(3) default current_timestamp );
ERROR 1067 (42000): Invalid default value for 'ts'
mysql> create table tbl_so_q23671222_2( ts timestamp(3) default current_timestamp(3) );
Query OK, 0 rows affected (0.38 sec)

current_timestamp(**3**) should be parsed, otherwise it's not a valid ddl. 

version: 5958b6fcdb2d33614d9b2ab99f76b5c1c422b39f

example: 
```go
func TestDDL(t *testing.T) {
	s := `create table t(dt datetime(6) not null default current_timestamp(6) on update current_timestamp(6))`
	p := parser.New()
	stmt, err := p.ParseOneStmt(s, "", "")
	require.NoError(t, err)
	b := bytes.NewBufferString("")
	ctx := format.NewRestoreCtx(format.DefaultRestoreFlags, b)
	require.NoError(t, stmt.Restore(ctx))
	fmt.Println(b.String())
}
```
 outputs 
```sql
`CREATE TABLE `t` (`dt` DATETIME(6) NOT NULL DEFAULT CURRENT_TIMESTAMP() ON UPDATE CURRENT_TIMESTAMP())` 
```
which is invalid.

spongedu:
Fix #186

### Check List 

Tests 

Code changes

Side effects

Related changes
after this pr is merged, we should take another pr in `TiDB` to add a check in ddl. As is described in [MySQL Reference](https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html):

```
If a TIMESTAMP or DATETIME column definition includes an explicit fractional seconds precision value anywhere, the same value must be used throughout the column definition.
```


zz-jason:
@spongedu seems you should fix the restore functions.

spongedu:
@zz-jason ok, I'll take a look

spongedu:
@zz-jason some cases have wrong results before. I've fixed them. PTAL @zz-jason

zz-jason:
LGTM

tiancaiamao:
LGTM

PR #212, opened 2019-02-16
model: view  add quote support

AndrewDi:
Store two version stmt in View's metadata and use `func (v ViewInfo) SelectStmtByQuote(quote string) string` to obtain `Quote Stmt String`.

Change `SelectStmt` from `String` to `Array String`.

### Check List 

Tests 

 - Unit test (in TiDB)

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -537,7 +537,18 @@ type ViewInfo struct {
 	Security    ViewSecurity       `json:"view_security"`
 	SelectStmt  string             `json:"view_select"`
 	CheckOption ViewCheckOption    `json:"view_checkoption"`
-	Cols        []CIStr            `json:"view_cols"`
+	//todo Remove this attribute when upgrade to next release version
+	Cols            []CIStr `json:"view_cols"`
+	SelectStmtQuote string  `json:view_select_quote`
+}
+
+// SelectStmtByQuote provide SelectStmt with different quote
+func (v ViewInfo) SelectStmtBySQLMode(mode mysql.SQLMode) string {
+	if mode.HasANSIQuotesMode() {
+		return v.SelectStmtQuote
+	} else {
+		return v.SelectStmt
+	}
 }
 
 // PartitionType is the type for PartitionInfo
--- a/model/model.go
+++ b/model/model.go
@@ -537,7 +537,7 @@ type ViewInfo struct {
 	Security    ViewSecurity       `json:"view_security"`
 	SelectStmt  string             `json:"view_select"`
 	CheckOption ViewCheckOption    `json:"view_checkoption"`
-	//todo Remove this attribute when upgrade to next release version
+	// TODO: Remove this attribute when upgrade to next release version
 	Cols            []CIStr `json:"view_cols"`
 	SelectStmtQuote string  `json:view_select_quote`
 }
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -80,6 +80,21 @@ func (*testModelSuite) TestModelBasic(c *C) {
 		PKIsHandle:  true,
 	}
 
+	view := &TableInfo{
+		ID:          1,
+		Name:        NewCIStr("v"),
+		Charset:     "utf8",
+		Collate:     "utf8_bin",
+		Columns:     []*ColumnInfo{column},
+		Indices:     nil,
+		ForeignKeys: nil,
+		PKIsHandle:  true,
+		View: &ViewInfo{
+			SelectStmt:      "SELECT 1 AS `C` FROM `DUAL`",
+			SelectStmtQuote: "SELECT 1 AS \"C\" FROM \"DUAL\"",
+		},
+	}
+
 	dbInfo := &DBInfo{
 		ID:      1,
 		Name:    NewCIStr("test"),
@@ -124,6 +139,9 @@ func (*testModelSuite) TestModelBasic(c *C) {
 	}
 	no := anIndex.HasPrefixIndex()
 	c.Assert(no, Equals, false)
+
+	c.Assert(view.View.SelectStmtBySQLMode(mysql.ModeNone), Equals, "SELECT 1 AS `C` FROM `DUAL`")
+	c.Assert(view.View.SelectStmtBySQLMode(mysql.ModeANSIQuotes), Equals, "SELECT 1 AS \"C\" FROM \"DUAL\"")
 }
 
 func (*testModelSuite) TestJobStartTime(c *C) {
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -2173,7 +2173,7 @@ type WindowSpec struct {
 
 // Restore implements Node interface.
 func (n *WindowSpec) Restore(ctx *RestoreCtx) error {
-	if name := n.Name.String(); name != "" {
+	if name := n.Name.String(); name != "" && name != "<unnamed window>" {
 		ctx.WriteName(name)
 		if n.OnlyAlias {
 			return nil


AndrewDi:
@XuHuaiyu @winkyao @zz-jason PTAL

zz-jason:
@AndrewDi Could you add some UT in the parser repo?

AndrewDi:
> @AndrewDi Could you add some UT in the parser repo?

Done

AndrewDi:
@zz-jason PTAL

model/model.go
```diff
@@ -384,9 +384,19 @@ type ViewInfo struct {
 	Algorithm   ViewAlgorithm      `json:"view_algorithm"`
 	Definer     *auth.UserIdentity `json:"view_definer"`
 	Security    ViewSecurity       `json:"view_security"`
-	SelectStmt  string             `json:"view_select"`
+	SelectStmt  []string           `json:"view_select"`
 	CheckOption ViewCheckOption    `json:"view_checkoption"`
-	Cols        []CIStr            `json:"view_cols"`
```
zz-jason: why removing `Cols`?

AndrewDi: before select statement contains wildcard, I need this attr to store some info, I raise this pr to expand select statement wildcard, so this attr no more needed.

zz-jason: OK. Test failed, seems you should modify some tests in the tidb repo: https://circleci.com/gh/AndrewDi/parser/73?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

AndrewDi: PR https://github.com/pingcap/tidb/pull/9328 has fix this test.

model/model.go
```diff
@@ -404,9 +404,19 @@ type ViewInfo struct {
...
+}
+
+// SelectStmtByQuote provide SelectStmt with different quote
+func (v ViewInfo) SelectStmtByQuote(quote string) string {
```
XuHuaiyu: change the input argument to SQLMode

model/model.go
```diff
@@ -404,9 +404,17 @@ type ViewInfo struct {
 	Algorithm   ViewAlgorithm      `json:"view_algorithm"`
 	Definer     *auth.UserIdentity `json:"view_definer"`
 	Security    ViewSecurity       `json:"view_security"`
-	SelectStmt  string             `json:"view_select"`
+	SelectStmt  []string           `json:"view_select"`
```
XuHuaiyu: We need a detailed description here for
1. why we need to store strings here
2. what do they store individually

model/model.go
```diff
@@ -401,12 +401,21 @@ func (v *ViewCheckOption) String() string {
...
-	Definer     *auth.UserIdentity `json:"view_definer"`
-	Security    ViewSecurity       `json:"view_security"`
-	SelectStmt  string             `json:"view_select"`
-	CheckOption ViewCheckOption    `json:"view_checkoption"`
-	Cols        []CIStr            `json:"view_cols"`
+	Algorithm ViewAlgorithm      `json:"view_algorithm"`
+	Definer   *auth.UserIdentity `json:"view_definer"`
+	Security  ViewSecurity       `json:"view_security"`
+	//In order to support ansi_quotes sqlmode, we need to store SelectStmt with different quotes.
+	SelectStmt  []string        `json:"view_select"`
```
erjiaqing: Can we create a new field called SelectStmtANSIQuote instead of replace this into a slice?

This breaks backward compatible, since if we use a old version of TiDB to open the new database, it will throw:

```
2019/03/22 15:51:03.950 tidb.go:79: [error] [ddl] init domain failed json: cannot unmarshal array into Go struct field ViewInfo.view_select of type string
github.com/pingcap/errors.AddStack
        /home/ejq/go/pkg/mod/github.com/pingcap/errors@v0.11.1/errors.go:174
github.com/pingcap/errors.Trace
        /home/ejq/go/pkg/mod/github.com/pingcap/errors@v0.11.1/juju_adaptor.go:15
github.com/pingcap/tidb/meta.(*Meta).ListTables
        /home/ejq/go/src/github.com/pingcap/tidb/meta/meta.go:398
github.com/pingcap/tidb/domain.(*Domain).fetchSchemasWithTables
        /home/ejq/go/src/github.com/pingcap/tidb/domain/domain.go:177
runtime.goexit
        /usr/lib/go/src/runtime/asm_amd64.s:1337
```

When starting up.

Another suggestion is that since show is a low frequency command we can just store the select statement quoted by back quotes, and rebuild it with double quotes when we need to show the statement.

AndrewDi: It's not only used b `SHOW CREATE` statement, `SELECT` statement use this too.

erjiaqing: So maybe we should use two fields instead of replace it by a slice

erjiaqing: Or maybe we can force parser to use back quotes when parsing views in select.

model/model_test.go
```diff
@@ -80,6 +80,21 @@ func (*testModelSuite) TestModelBasic(c *C) {
...
+		ID:          1,
+		Name:        NewCIStr("v"),
+		Charset:     "utf8",
+		Collate:     "utf8_bin",
+		Columns:     []*ColumnInfo{column},
+		Indices:     nil,
+		ForeignKeys: nil,
+		PKIsHandle:  true,
+		View: &ViewInfo{
+			SelectStmt:      "SELECT 1 AS 'C' FROM 'DUAL'",
```
kennytm: 😕 `SELECT 1 AS 'C' FROM 'DUAL'` is invalid whether ANSI_QUOTES is enabled or not. Do you mean to use `` ` `` instead of `'`?

| Quotation mark | With ANSI_QUOTES | Without ANSI_QUOTES |
|:-:|:-:|:-:|
| `'…'` | String literal | String literal |
| `"…"` | Identifier | String literal |
| `` `…` `` | Identifier | Identifier |

AndrewDi: my mistake。。。

erjiaqing:
What if we operate on a view created in older versions, and do not have `SelectStmtQuoted` param?

model/model.go
```diff
@@ -432,7 +432,18 @@ type ViewInfo struct {
 	Security    ViewSecurity       `json:"view_security"`
 	SelectStmt  string             `json:"view_select"`
 	CheckOption ViewCheckOption    `json:"view_checkoption"`
-	Cols        []CIStr            `json:"view_cols"`
+	//todo Remove this attribute when upgrade to next release version
```
zz-jason:
```suggestion
	// TODO: Remove this attribute when upgrade to next release version
```

AndrewDi:
> What if we operate on a view created in older versions, and do not have `SelectStmtQuoted` param?

Empty string would return.

erjiaqing:
> Quotation mark |	With ANSI_QUOTES |	Without ANSI_QUOTES
> ---------------|------------|-------------
> `'…'` |	String literal |	String literal
> `"…"` |	Identifier |	String literal
> `` `…` `` |	Identifier |	Identifier

According to this, if we always use `'` for string literals and `` ` `` for identifiers and store the views, we won't have issues with or without `ANSI_QUOTES`. So I still think that we don't have to store the two  SelectStmts.

AndrewDi:
> > Quotation mark
> > With ANSI_QUOTES
> > Without ANSI_QUOTES
> > 
> > 
> > 
> > 
> > `'…'`
> > String literal
> > String literal
> > 
> > 
> > `"…"`
> > Identifier
> > String literal
> > 
> > 
> > `` `…` ``
> > Identifier
> > Identifier
> 
> According to this, if we always use `'` for string literals and `` ` `` for identifiers when store the views, we won't have issues with or without `ANSI_QUOTES`. So I still think that we don't have to store the two SelectStmts.

If sqlmod=ansi_quotes is enable, and user create a view with `"…"`, then what happened?

erjiaqing:
After we parse the statement, there will be no difference between `"` and `` ` ``
According to your PR https://github.com/pingcap/tidb/pull/9328 we can see that SelectStmt is generated by `Select.Restore` instead of user input (`Select.Text()`).

kennytm:
BTW I wonder how would this work regarding NO_BACKSLASH_ESCAPES (which changes the parse result of `'\n'`)

model/model.go
```diff
@@ -537,7 +537,18 @@ type ViewInfo struct {
 	Security    ViewSecurity       `json:"view_security"`
 	SelectStmt  string             `json:"view_select"`
 	CheckOption ViewCheckOption    `json:"view_checkoption"`
-	Cols        []CIStr            `json:"view_cols"`
+	// TODO: Remove this attribute when upgrade to next release version
+	Cols            []CIStr `json:"view_cols"`
+	SelectStmtQuote string  `json:view_select_quote`
```
zz-jason: Seems we don't need to add a new field. We can traverse `SelectStmt` and replace ` to " to obtain the ansi quoted statement.

AndrewDi: @zz-jason Can we remove attr `Cols` when upgrade to 4.0?

XuHuaiyu:
A new PR will be pushed.

PR #213, opened 2019-02-19
parser: add gramma for RBAC

imtbkcat:
MySQL has supported role-based access control since MySQL 8.0. We plan to support this function step by step. The first step is adding gramma for parser.

just add CNF gramma for role operations. 
reference: [https://dev.mysql.com/doc/refman/8.0/en/roles.html](https://dev.mysql.com/doc/refman/8.0/en/roles.html)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


go.mod1
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/pingcap/errors v0.11.0
 	github.com/pingcap/tidb v0.0.0-20190218065808-69472bd1a6e9
 	github.com/pingcap/tipb v0.0.0-20190107072121-abbec73437b7
+	github.com/pkg/errors v0.8.1
```
kennytm: This PR shouldn't need to add the `pkg/errors` dependency? (it already depends on `pingcap/errors`)

auth/auth.go
```diff
@@ -31,6 +31,7 @@ type UserIdentity struct {
 	CurrentUser  bool
 	AuthUsername string // Username matched in privileges system
 	AuthHostname string // Match in privs system (i.e. could be a wildcard)
+	IsRole       bool
```
tiancaiamao: How about defining a `RoleIdentity` ?
```
type RoleIdentity struct {
    Username string
    Hostname string
}
```

parser.y
```diff
@@ -6364,17 +6448,20 @@ Statement:
 |	CreateTableStmt
 |	CreateViewStmt
 |	CreateUserStmt
+|   CreateRoleStmt
```
tiancaiamao: ident

parser.y
```diff
@@ -6384,6 +6471,7 @@ Statement:
 |	RenameTableStmt
 |	ReplaceIntoStmt
 |	RevokeStmt
+|   RevokeRoleStmt
```
tiancaiamao: ident

parser.y
```diff
@@ -7374,6 +7467,26 @@ HashString:
 		$$ = $1
 	}
 
+RoleSpec:
+	Rolename
+	{
+		roleSpec := &ast.UserSpec{
+			User: $1.(*auth.UserIdentity),
+		}
+		roleSpec.AuthOpt = nil
```
tiancaiamao: Is there a password for a role? and what is it?

parser.y
```diff
@@ -7374,6 +7467,26 @@ HashString:
 		$$ = $1
 	}
 
+RoleSpec:
+	Rolename
+	{
+		roleSpec := &ast.UserSpec{
```
tiancaiamao: Should we add an `IsRole` field to the `UserSpec`?

parser.y
imtbkcat: Role doesn't have password, so set this field to `nil`.

parser.y
imtbkcat: yes

auth/auth.go
imtbkcat: I plan to reuse `UserIdentity` and `CreateUserStmt` for `CreateRoleStmt`.

parser.y
tiancaiamao: `auth.UserIdentity` is used for authorization, don't make it too heavy.

Role should not be used to login, so it's better to distinguish Role and User.

I think maybe you can define them like this:

```
// UserSpec is used for parsing create user statement.
type UserSpec struct {
	User    *auth.UserIdentity
	AuthOpt *AuthOption
    IsRole  bool
}
type RoleIdentity struct {
	Username     string
	Hostname     string
}
```

And you can still create the `UserSpec` here:

```
role := $1.(*auth.RoleIdentity)
&ast.UserSpec {
    User: &auth.UserIdentity{
	    Username: role.Username,
	    Hostname: role.Hostname,        
    },
    IsRole: true,
}
```

imtbkcat: got it

tiancaiamao:
LGTM

PR #215, opened 2019-02-20
normalize and generate digest at same time

lysu:

in previous PR, we add two method `Normalize` / `DigestHash`, but for real tidb's usage, it's better to general normalized sql and digest at some time, and cache it into queryCtx and only do it one-pass

add a new overwrite method ---- `NormalizeDigest` and also remove `useless lex.reset`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - N/A

Related changes

 - N/A

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/215)


tiancaiamao:
LGTM

PR #221, opened 2019-02-25
parser: fix stored option in generated column restore

amyangfei:
`STORED` option for generated column is missing after `Restore`

- output `STORED` or `VIRTUAL` based on the `Stored` field of column option.
- fix errors lib introduced in commit https://github.com/pingcap/parser/commit/60fd134cfde20424ba588c785debb3be0a086210

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -407,6 +407,11 @@ func (n *ColumnOption) Restore(ctx *RestoreCtx) error {
 			return errors.Annotate(err, "An error occurred while splicing ColumnOption GENERATED ALWAYS Expr")
 		}
 		ctx.WritePlain(")")
+		if n.Stored {
+			ctx.WritePlain(" STORED")
+		} else {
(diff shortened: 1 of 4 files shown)


GregoryIan:
LGTM

GregoryIan:
@kennytm PTAL

ast/ddl.go
```diff
@@ -407,6 +407,11 @@ func (n *ColumnOption) Restore(ctx *RestoreCtx) error {
 			return errors.Annotate(err, "An error occurred while splicing ColumnOption GENERATED ALWAYS Expr")
 		}
 		ctx.WritePlain(")")
+		if n.Stored {
+			ctx.WritePlain(" STORED")
+		} else {
+			ctx.WritePlain(" VIRTUAL")
```
kennytm: These should be `WriteKeyWord`.

amyangfei:
PTAL @kennytm

PR #225, opened 2019-02-27
model: add version field to tableInfo. (#223)

zimulala:
cherry-pick #223


zimulala:
PTAL @crazycs520 @winkyao @kennytm

PR #228, opened 2019-03-01
parser, ast: add `SET ROLE` support

imtbkcat:
`SET ROLE` is an important part of RBAC function. Many other RBAC grammar rely on `SET ROLE`. The detail document for `SET ROLE` is here : https://dev.mysql.com/doc/refman/8.0/en/set-role.html

Adding `RoleIdentity` and `SetRoleStmt` for `ast` package, which contain information about those role is going to be active.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity


go.mod1
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/pingcap/errors v0.11.0
 	github.com/pingcap/tidb v0.0.0-20190218065808-69472bd1a6e9
 	github.com/pingcap/tipb v0.0.0-20190107072121-abbec73437b7
+	github.com/pkg/errors v0.8.1
```
kennytm: Shouldn't need to change go.mod

imtbkcat: fixed

auth/auth.go
```diff
@@ -67,6 +62,26 @@ func (user *UserIdentity) AuthIdentityString() string {
...
+		ctx.WritePlain("@")
+		ctx.WriteName(role.Hostname)
+	}
+	return nil
+}
+
+// String converts UserIdentity to the format user@host.
+func (role *RoleIdentity) String() string {
+	// TODO: Escape username and hostname.
+	return fmt.Sprintf("%s@%s", role.Username, role.Hostname)
```
tiancaiamao: `%s`@`%s` ?

tiancaiamao:
LGTM

imtbkcat:
tidb pr: https://github.com/pingcap/tidb/pull/9581

tiancaiamao:
PTAL @jackysp  @lysu @kennytm

ast/misc.go
```diff
@@ -721,6 +722,62 @@ func (n *SetPwdStmt) Accept(v Visitor) (Node, bool) {
...
+	case SetRoleDefault:
+		ctx.WriteKeyWord(" DEFAULT")
+	case SetRoleNone:
+		ctx.WriteKeyWord(" NONE")
+	case SetRoleAll:
+		ctx.WriteKeyWord(" ALL")
+	case SetRoleAllExcept:
+		ctx.WriteKeyWord(" ALL EXCEPT")
+	}
+	if len(n.RoleList) > 0 {
```
kennytm: This check is unnecessary.

go.mod1
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/pingcap/errors v0.11.0
 	github.com/pingcap/tidb v0.0.0-20190218065808-69472bd1a6e9
 	github.com/pingcap/tipb v0.0.0-20190107072121-abbec73437b7
+	github.com/shirou/gopsutil v2.18.12+incompatible // indirect
```
kennytm: Where is this used? 🤔

imtbkcat:
fixed @kennytm

PR #233, opened 2019-03-06
*: parser subpartition option for 'partition by range columns'

tiancaiamao:

Parse

```
mysql> create table t1 (a varchar(5), b int signed, c varchar(10), d datetime)
    -> partition by range columns(b,c)
    -> subpartition by hash(to_seconds(d))
    -> ( partition p0 values less than (2, 'b'),
    -> partition p1 values less than (4, 'd'),
    -> partition p2 values less than (10, 'za'));
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 3 column 13 near "subpartition by hash(to_seconds(d))
( partition p0 values less than (2, 'b'),
partition p1 values less than (4, 'd'),
partition p2 values less than (10, 'za'))"
```

Parse `subpartition` and ignore it for 'partition by range columns'

### Check List 

Tests 

 - Unit test


tiancaiamao:
PTAL @lysu @jackysp

lysu:
@tiancaiamao please fix ci "unrecognized import path "golang.org/x/oauth2""

PR #234, opened 2019-03-06
parser: parse 'partition by linear hash' and ignore it

tiancaiamao:

```
mysql> CREATE TABLE t1 (a INT, b TIMESTAMP DEFAULT '0000-00-00 00:00:00')
    -> ENGINE=INNODB PARTITION BY LINEAR HASH (a) PARTITIONS 1;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 2 column 34 near "LINEAR HASH (a) PARTITIONS 1"
```

Parse the `linear` keyword and ignore it (treat it as normal table)

### Check List 

Tests 

 - Unit test


tiancaiamao:
PTAL @lysu @jackysp

parser.y
```diff
@@ -1980,16 +1982,19 @@ PartitionOpt:
...
 			// If you do not include a PARTITIONS clause, the number of partitions defaults to 1
 			Num: 1,
 		}
-		if $7 != nil {
-			tmp.Num = getUint64FromNUM($7)
+		if $8 != nil {
+			tmp.Num = getUint64FromNUM($8)
+		}
+		if $3 != "" {
+			tmp = nil
```
lysu: do we need add a warn msg to create table user?

PR #237, opened 2019-03-08
executor: add `DROP ROLE` support

imtbkcat:
add `DROP ROLE` function for parser and executor.

Role is same as user, is store in `mysql.user` table. We do same action as `DROP USER` when we need to drop a role. Also, as we has added some system table for RBAC, like `mysql.role_edges` and 
 `mysql.default_role`. We have to delete information from these tables when execute `DROP ROLE`

tidb pr: https://github.com/pingcap/tidb/pull/9616

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -2429,19 +2429,31 @@ DropViewStmt:
 DropUserStmt:
 	"DROP" "USER" UsernameList
 	{
-		$$ = &ast.DropUserStmt{IfExists: false, UserList: $3.([]*auth.UserIdentity)}
+		$$ = &ast.DropUserStmt{IsDropRole: false, IfExists: false, UserList: $3.([]*auth.UserIdentity)}
 	}
 |	"DROP" "USER" "IF" "EXISTS" UsernameList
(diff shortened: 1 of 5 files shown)


parser.y
```diff
@@ -2429,19 +2429,31 @@ DropViewStmt:
 DropUserStmt:
 	"DROP" "USER" UsernameList
 	{
-		$$ = &ast.DropUserStmt{IfExists: false, UserList: $3.([]*auth.UserIdentity)}
+		$$ = &ast.DropUserStmt{IsDropRole: false, IfExists: false, UserList: $3.([]*auth.UserIdentity)}
```
zz-jason: how about using a new `DropRoleStmt`?

imtbkcat: actually there is no difference between `DROP USER` and `DROP ROLE`. I think `IsDropRole` can be removed.

tiancaiamao:
LGTM

PR #242, opened 2019-03-14
ast: add GrantRoleStmt to support `GRANT ROLE`

imtbkcat:
add `GrantRoleStmt` to support SQL like `GRANT Roles TO User`.

Add `ast.GrantRoleStmt`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


tiancaiamao:
LGTM

PR #245, opened 2019-03-18
ast: add revoke role support

imtbkcat:
Support revoke roles from users.

Add `RevokeRoleStmt` structure which could parser `REVOKE FROM` SQL. Executor in TiDB will read information from `RevokeRoleStmt` to delete relation from `mysql.role_edges`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

### DIFF
--- a/ast/misc.go
+++ b/ast/misc.go
@@ -1597,6 +1597,47 @@ func (n *RevokeStmt) Accept(v Visitor) (Node, bool) {
 	return v.Leave(n)
 }
 
+// RevokeStmt is the struct for REVOKE statement.
+type RevokeRoleStmt struct {
+	stmtNode
+
+	Roles []*auth.RoleIdentity
+	Users []*auth.UserIdentity
+}
+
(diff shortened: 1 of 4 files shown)


ast/misc.go
```diff
@@ -1471,6 +1471,49 @@ func (n *RevokeStmt) Accept(v Visitor) (Node, bool) {
...
+	stmtNode
+
+	Roles []*auth.RoleIdentity
+	Users []*auth.UserIdentity
+}
+
+// Restore implements Node interface.
+func (n *RevokeRoleStmt) Restore(ctx *RestoreCtx) error {
+	ctx.WriteKeyWord("REVOKE ")
+	if len(n.Roles) > 0 {
```
kennytm: This condition is unnecessary.

jackysp:
PTAL @tiancaiamao @kennytm

imtbkcat:
@kennytm conflict has been fixed.

PR #248, opened 2019-03-19
mysql/charset.go: Add quick path to avoid access map

july2993:
for https://github.com/pingcap/tidb/issues/9784, help avoid access map mysql.CharsetIDs

note if write all the case by switch it may be change by compile from linear search to binary search, will be still slow compared to using map.

Add quick path to avoid access map, we can check the only a few supported charsets first.

a benchmark
code:
```go
func BenchmarkSwitchFirst(b *testing.B) {
    for i := 0; i < b.N; i++ {
        var _ = CharsetNameToID("utf8mb4")
    }
}

func BenchmarkSwitchLast(b *testing.B) {
    for i := 0; i < b.N; i++ {
        var _ = CharsetNameToID("latin1")
    }
}

func BenchmarkMap(b *testing.B) {
    for i := 0; i < b.N; i++ {
        var _ = charsetIDs["utf8mb4"]
    }
}
```
result
```
goos: darwin
goarch: amd64
pkg: github.com/pingcap/parser/mysql
BenchmarkSwitchFirst-4          2000000000               1.31 ns/op
BenchmarkSwitchLast-4           2000000000               1.58 ns/op
BenchmarkMap-4                  100000000               14.0 ns/op
PASS
ok      github.com/pingcap/parser/mysql 7.479s
```

### Check List 

Tests 

Code changes

Side effects
- Increased code complexity
make code a little unreadable

Related changes


ngaut:
Good job.

mysql/charset.go
```diff
@@ -15,8 +15,27 @@ package mysql
 
 import "unicode"
 
-// CharsetIDs maps charset name to its default collation ID.
-var CharsetIDs = map[string]uint8{
+// CharsetNameToID maps charset name to its default collation ID.
+func CharsetNameToID(charset string) uint8 {
+	// Use quick path for TiDB to avoid access charsetIDs map
+	// "SHOW CHARACTER SET;" to see all the supported character sets.
+	if charset == "utf8mb4" {
```
kennytm: Why use an if/else chain instead of an actual `switch`? I've benchmarked this locally using Go 1.12 and there's no observable difference (within ±0.05 ns/op).

july2993: oh, the benchmark code i paste should `s/Swithch/If/`.
just want to make sure using linear search(if goto), if there's too many switch case it will be slow  
even if you `SwitchFirst`, because it's doing binary search on a sorted array with all case.

ngaut:
LGTM

ngaut:
Please fix CI

mysql/charset.go
```diff
@@ -15,6 +15,25 @@ package mysql
 
 import "unicode"
 
+// CharsetNameToID maps charset name to its default collation ID.
+func CharsetNameToID(charset string) uint8 {
+	// Use quick path for TiDB to avoid access CharsetIDs map
+	// "SHOW CHARACTER SET;" to see all the supported character sets.
+	if charset == "utf8mb4" {
+		return 45
```
kennytm: Perhaps define consts for these 5 constants so we don't need to repeat the special values in CharsetIDs again.

july2993: add in https://github.com/pingcap/parser/pull/248/commits/9093fbb5c65f9dd6a9ec5968ed23c387cebca438

kennytm: I mean the constants 45, 63, 33, 11, 8 😅

PR #250, opened 2019-03-20
*: skip 'go mod tidy' check in integration test

tiancaiamao:

Check `go mod tidy` in integration test is unnecessary

Use `make gotest` instead of `make test`

### Check List 

Tests 

 - Integration test


tiancaiamao:
PTAL @lysu

PR #252, opened 2019-03-21
model: add MaxShardRowIDBits to table info to record the max ShardRowIDBits be used so far

crazycs520:
Help to fix share-rowid-bit problem. We should  record the max share-rowid-bit has been used.

add MaxShardRowIDBits to table info to record the max ShardRowIDBits be used so far.

### Check List 

Tests


model/model.go
```diff
@@ -212,6 +212,8 @@ type TableInfo struct {
 
 	// ShardRowIDBits specify if the implicit row ID is sharded.
 	ShardRowIDBits uint64
+	// MaxShardRowIDBits uses to record the max ShardRowIDBits be used so far.
+	MaxShardRowIDBits uint64
```
winkyao: Please add json field tags.

crazycs520: done.

PR #243, opened 2019-03-14
Support update pump or drainer status

aliiohs:
### What problem does this PR solve? 
support_update_Pump_or_Drainer_status
proposal：https://github.com/pingcap/tidb/issues/9201

### What is changed and how it works?
it will change parser.y/parser.go/ast.misc.go /misc.go.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #217, opened 2019-02-20
parser: add support for `show pump/drainer status`

caohe:
### What problem does this PR solve? 
Add support for `show pump/drainer status`
[#9201](https://github.com/pingcap/tidb/issues/9201)

### What is changed and how it works?
Add pump and drainer to parser.y and dml.go

### Check List 

Tests 

 - Unit test

Code changes

 - Has interface methods change

Side effects

 - None

Related changes

 - None

PR #259, opened 2019-03-25
support show pump/drainer status && change pump/drainer's status

WangXiangUSTC:
support show pump/drainer status && change pump/drainer's status

git cherry pick 
https://github.com/pingcap/parser/pull/217
https://github.com/pingcap/parser/pull/243

### Check List 

Tests 

 - Unit test


kennytm:
LGTM

parser.y
```diff
@@ -6056,6 +6079,18 @@ ShowTargetFilterable:
 			Tp: ast.ShowProcedureStatus,
 		}
 	}
+|	"PUMP" "STATUS"
+	{
+		$$ = &ast.ShowStmt {
+			Tp: ast.ShowPumpStatus,
+		}
+	}
+|	"DRAINER" "STATUS"
```
zimulala: Do we add a test to parser_test.go ?

WangXiangUSTC: ok，I will add test in master, and then cherry-pick to release branch

zimulala: No need to push a PR only for this, I just give a suggestion

PR #260, opened 2019-03-25
parser, ast: add ColumnOptionCollate for all column types

bb7133:
In MySQL `CREATE TABLE` syntax, column collation is part of the column option, so the following statements are valid(although meaningless):

```
create table t(a decimal collate utf8_bin)
create table t(a decimal collate utf8_bin collate utf8_general_ci)
```

This PR makes those statements acceptable

Column "COLLATE" definitions are moved from 'StringType' to 'ColumnOption'

### Check List

Tests
 - Unit test

Code changes

Side effects
 - Increased code complexity

Related changes
 - None


parser.y
```diff
@@ -1481,7 +1481,18 @@ ColumnDefList:
 ColumnDef:
 	ColumnName Type ColumnOptionListOpt
 	{
-		$$ = &ast.ColumnDef{Name: $1.(*ast.ColumnName), Tp: $2.(*types.FieldType), Options: $3.([]*ast.ColumnOption)}
+		def := &ast.ColumnDef{Name: $1.(*ast.ColumnName), Tp: $2.(*types.FieldType)}
+		opts := make([]*ast.ColumnOption, 0)
+		for _, opt := range $3.([]*ast.ColumnOption) {
```
winkyao: Move this logic to CreateTable?

bb7133: Sure

ast/ddl.go
```diff
@@ -416,6 +418,12 @@ func (n *ColumnOption) Restore(ctx *RestoreCtx) error {
 		if err := n.Refer.Restore(ctx); err != nil {
 			return errors.Annotate(err, "An error occurred while splicing ColumnOption ReferenceDef")
 		}
+	case ColumnOptionCollate:
```
crazycs520: Please add test in `TestDDL` for testing restore.

bb7133: addressed, PTAL!

zimulala:
Please fix the conflict.

bb7133:
Updated, thanks @zimulala

PR #263, opened 2019-03-27
fix key work node_id && node_state

WangXiangUSTC:
fix issue https://github.com/pingcap/tidb/issues/9910
treat `node_id` and `node_stats` as key work

move `pump` `drainer` `node_id` `node_stats` to `NotKeywordToken`

### Check List 

Tests 

 - Manual test (create table, use `pump` `drainer` `node_id` `node_stats` as column name)

Related changes


kennytm:
PTAL @zimulala

parser.y
```diff
@@ -484,15 +482,17 @@ import (
...
 	buckets		"BUCKETS"
 	cancel		"CANCEL"
-	ddl		"DDL"
-	drainer         "DRAINER"
+	ddl		    "DDL"
+	drainer     "DRAINER"
 	jobs		"JOBS"
 	job		    "JOB"
-	pump            "PUMP"
+	node_id		"NODE_ID"
```
lonng:
```suggestion
	nodeId		"NODE_ID"
```
It's better to keep consistent with others.

parser.y
```diff
@@ -267,7 +265,7 @@ import (
 	zerofill		"ZEROFILL"
 	natural			"NATURAL"
 
-	/* The following tokens belong to UnReservedKeyword. */
+	/* The following tokens belong to UnReservedKeyword. Notice: make sure these tokens are contained in UnReservedKeyword. */
```
kennytm: 🤔 So turns out this is already checked in `consistent_test.go`

parser.y
zimulala: I think `nodeID` is better.

parser.y
```diff
@@ -484,15 +482,17 @@ import (
...
 
-	/* The following tokens belong to TiDBKeyword. */
+	/* The following tokens belong to TiDBKeyword. Notice: make sure these tokens are contained in TiDBKeyword. */
 	admin		"ADMIN"
 	buckets		"BUCKETS"
 	cancel		"CANCEL"
-	ddl		"DDL"
-	drainer         "DRAINER"
+	ddl		    "DDL"
+	drainer     "DRAINER"
```
zimulala: Please align.

parser.y
WangXiangUSTC: 🤔but it don't take effect when 'node_id' not in keyword list, I don't know why

tiancaiamao:
LGTM

parser.y
lonng: Only put `node_id` in TiDBKeyword part (or other special parts) can take effect.

PR #264, opened 2019-03-27
fix key word node_id and node_state

WangXiangUSTC:
fix key word node_id and node_state

because have too many conflicts, just copy code from https://github.com/pingcap/parser/pull/263


tiancaiamao:
PTAL @zimulala

tiancaiamao:
LGTM

PR #265, opened 2019-03-28
terror: fix 2.1 CI

tiancaiamao:

Fix CI

Tiny adjust since the error package version is different in 2.1

### Check List 

Tests 

 - Unit test
 - No code


tiancaiamao:
PTAL @WangXiangUSTC @lysu

PR #268, opened 2019-03-31
parser: fix error message for illegal year type definition.

spongedu:

When create table, if year type is not defined correctly, the error message is not consistent with MySQL for now. In MySQL:
```sql
mysql> create table t(f_year year(5))ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
ERROR 1818 (HY000): Supports only YEAR or YEAR(4) column.
```
While in TiDB:
```
tidb> create table t(f_year year(5))ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for th
e right syntax to use line 1 column 30 near ")ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin"Supports only Y
EAR or YEAR(4) column.
```

Return correct error type in parser.

### Check List 

Tests 

 - Unit test
 - No code

Code changes

Side effects

Related changes

### DIFF
--- a/mysql/errname.go
+++ b/mysql/errname.go
@@ -833,7 +833,7 @@ var MySQLErrName = map[uint16]string{
 	ErrInternal:                                              "Internal : %s",
 	ErrInnodbImport:                                          "ALTER TABLE '%-.192s' IMPORT TABLESPACE failed with error %d : '%s'",
 	ErrInnodbIndexCorrupt:                                    "Index corrupt: %s",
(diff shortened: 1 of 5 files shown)


spongedu:
@qw4990  PTAL :)

PR #269, opened 2019-04-01
Parser: Support CREATE USER extensions per MySQL 5.7

spongedu:
### What problem does this PR solve? 
fix #158 

### What is changed and how it works?

Support related syntax in Parser side.  NOTE that this pr just make parser happy, no other things are really done . The related `ast` structures will be supplemented later.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change

Side effects

Related changes

PR #275, opened 2019-04-05
parser: complete Restore for CreateUserStmt

spongedu:
in #269 I've refined the `CREATE USER` syntax in parser. This pr completes the `Restore` part for it.

Define related Asts in `ast/misc.go` and do the related construction work in parser.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


ast/misc.go
```diff
@@ -901,14 +901,114 @@ func (n *UserSpec) EncodedPassword() (string, bool) {
...
+	case Cipher:
+		ctx.WriteKeyWord("CIPHER ")
+		ctx.WriteString(t.Value)
+	case Issuer:
+		ctx.WriteKeyWord("ISSUER ")
+		ctx.WriteString(t.Value)
+	case Subject:
+		ctx.WriteKeyWord("CIPHER")
+		ctx.WriteString(t.Value)
+	}
```
kennytm: Add a default case and return error.

```suggestion
	default:
		return errors.Errorf("Unsupported TslOption.Type %d", t.Type)
	}
```

ast/misc.go
```diff
@@ -901,14 +901,114 @@ func (n *UserSpec) EncodedPassword() (string, bool) {
...
+	switch r.Type {
+	case MaxQueriesPerHour:
+		ctx.WritePlain("MAX_QUERIES_PER_HOUR ")
+	case MaxUpdatesPerHour:
+		ctx.WritePlain("MAX_UPDATES_PER_HOUR ")
+	case MaxConnectionsPerHour:
+		ctx.WritePlain("MAX_CONNECTIONS_PER_HOUR ")
+	case MaxUserConnections:
+		ctx.WritePlain("MAX_USER_CONNECTIONS ")
+	}
```
kennytm: (Ditto, add a default)

ast/misc.go
```diff
@@ -901,14 +901,114 @@ func (n *UserSpec) EncodedPassword() (string, bool) {
...
+
+type ResourceOption struct {
+	Type  int
+	Count int64
+}
+
+func (r *ResourceOption) Restore(ctx *RestoreCtx) error {
+	switch r.Type {
+	case MaxQueriesPerHour:
+		ctx.WritePlain("MAX_QUERIES_PER_HOUR ")
```
kennytm: These should be keywords?

```suggestion
		ctx.WriteKeyWord("MAX_QUERIES_PER_HOUR ")
```

ast/misc.go
```diff
@@ -901,14 +901,114 @@ func (n *UserSpec) EncodedPassword() (string, bool) {
...
+
+type PasswordOrLockOption struct {
+	Type  int
+	Count int64
+}
+
+func (p *PasswordOrLockOption) Restore(ctx *RestoreCtx) error {
+	switch p.Type {
+	case PasswordExpire:
+		ctx.WritePlain("PASSWORD EXPIRE")
```
kennytm: (Ditto, `WriteKeyWord`)

ast/misc.go
```diff
@@ -901,14 +901,114 @@ func (n *UserSpec) EncodedPassword() (string, bool) {
...
+		ctx.WritePlain("PASSWORD EXPIRE NEVER")
+	case PasswordExpireInterval:
+		ctx.WritePlainf("PASSWORD EXPIRE NEVER %d DAY", p.Count)
+	case Lock:
+		ctx.WritePlain("ACCOUNT ")
+		ctx.WriteKeyWord("LOCK")
+	case Unlock:
+		ctx.WritePlain("ACCOUNT ")
+		ctx.WriteKeyWord("UNLOCK")
+	}
```
kennytm: (Ditto, `default:`)

ast/misc.go
```diff
@@ -929,6 +1029,42 @@ func (n *CreateUserStmt) Restore(ctx *RestoreCtx) error {
...
+		if i != tslOptionLen {
+			ctx.WriteKeyWord(" AND ")
+		}
+		if err := n.TslOptions[i-1].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore CreateUserStmt.TslOptions[%d]", i)
+		}
+	}
+
+	if len(n.ResourceOptions) != 0 {
+		ctx.WriteKeyWord(" WITH ")
```
```suggestion
		ctx.WriteKeyWord(" WITH")
```

so that...

ast/misc.go
```diff
@@ -929,6 +1029,42 @@ func (n *CreateUserStmt) Restore(ctx *RestoreCtx) error {
...
+	}
+
+	for i, v := range n.ResourceOptions {
+		if i != 0 {
```
kennytm: ... this doesn't need to be conditional 🙂 

```suggestion
```

ast/misc.go
spongedu: ok

ast/misc.go
spongedu: Should the `UnReservedKeyword`s be restored using `WriteKeyWord` ?

spongedu:
@kennytm PTAL

ast/misc.go
kennytm: Yes, every word that acts like a keyword (being unquoted and case-insensitive) should be restored using `WriteKeyWord`.

PR #270, opened 2019-04-01
types: port 'HasCharset' function from MySQL

bb7133:
### What problem does this PR solve?
This PR aims to solve the some known show charset/collation issues in TiDB, for example, https://github.com/pingcap/tidb/issues/9837

### What is changed and how it works?
A `HasCharset` function is ported from MySQL source, for example:

[Field::has_charset()](https://github.com/mysql/mysql-server/blob/3701bd36bec259fe494c75c9b30c57b01e598297/sql/field.h#L1417)

[Field_string::has_charset()](
 https://github.com/mysql/mysql-server/blob/3701bd36bec259fe494c75c9b30c57b01e598297/sql/field.h#L3541)

This function can be used in the procedure of some show-schema executions, for example, in MySQL it decides where a given column's `Collation` result is `NULL` in `show full columns` statement:

[sql_show.cc::store_column_type()](https://github.com/mysql/mysql-server/blob/3701bd36bec259fe494c75c9b30c57b01e598297/sql/sql_show.cc#L5303)

### Check List

Tests 

 - Unit test
 - Integration test

Code changes

 - Has exported function/method change

Side effects

 - None

Related changes

 - Need to cherry-pick to the release branch

PR #283, opened 2019-04-10
types: port 'HasCharset' function from MySQL (#270)

bb7133:
Cherry-pick https://github.com/pingcap/parser/pull/270


bb7133:
addressed, PTAL @xiekeyi98

Issue #280, opened 2019-04-09
Support ON UPDATE ddl statements


EmmaLouise2018:
## Bug Report

1. What did you do?

Looking at the ddl tests in the `testDDLSuite`, I was under the impression from `TestDDLOnUpdateRestore` that parsing sql statements with "ON UPDATE" were supported, however, in attempting to parse the following statement I encountered an error.  

Code snippet:
```
 p := parser.New()
 testSQL := "create table test(id int key, FOREIGN KEY (id) REFERENCES test2(id) ON UPDATE CASCADE);"
 if _, _, err := p.Parse(testSQL, "", ""); err != nil {
     logger.Error("failed to parse", zap.Error(err))
 } 
```

2. What did you expect to see?

I expected this statement to be able to be parsed without an issue. 


3. What did you see instead?

I encountered the error `"error":"line 1 column 77 near \"UPDATE CASCADE);\"` 


Looking closer at the current tests in `TestDDLOnUpdateRestore`
```
func (ts *testDDLSuite) TestDDLOnUpdateRestore(c *C) {
    testCases := []NodeRestoreTestCase{
        {"ON UPDATE RESTRICT", "ON UPDATE RESTRICT"},
        {"on update CASCADE", "ON UPDATE CASCADE"},
        {"on update SET NULL", "ON UPDATE SET NULL"},
        {"on update no action", "ON UPDATE NO ACTION"},
    }
    extractNodeFunc := func(node Node) Node {
        return node.(*CreateTableStmt).Constraints[1].Refer.OnUpdate
    }
    RunNodeRestoreTest(c, testCases, "CREATE TABLE child ( id INT, parent_id INT, INDEX par_ind (parent_id), FOREIGN KEY (parent_id) REFERENCES parent(id)  ON DELETE CASCADE %s )", extractNodeFunc)
}
```

All `ON UPDATE` test cases are inserted after an `ON DELETE CASCADE` clause. When I removed, the `ON DELETE` and instead used 

`RunNodeRestoreTest(c, testCases, "CREATE TABLE child ( id INT, parent_id INT, INDEX par_ind (parent_id), FOREIGN KEY (parent_id) REFERENCES parent(id) %s )", extractNodeFunc)`

The tests for `TestDDLOnUpdateRestore` start to fail

```
----------------------------------------------------------------------                                                                                                                   
FAIL: ddl_test.go:91: testDDLSuite.TestDDLOnUpdateRestore                                                                                                                                
                                                                                                                                                                                         
[...]

PR #286, opened 2019-04-11
Initial attempt to add ON UPDATE support. 

EmmaLouise2018:
Opening up this PR, because I am not entirely sure how to completely resolve this issue and would be grateful for some input/suggestions. 

This PR was opened in reference to https://github.com/pingcap/parser/issues/280 in an attempt to add support fo ON UPDATE sql statement 

The change was to try to update the grammar to support `ON UPDATE` clauses without `ON DELETE` clauses. The changes are NOT complete. 

Running `make test` generates a `parser.go` file, which when the tests are run they all pass. In addition, all the sql statements I am working with seem to parse correctly. However, the change I am attempting to make also introduces, perhaps unsurprisingly reduce/reduce conflicts that I am not sure how to resolve.

This is my first time working with Yacc, and it's been a while since I've defined a grammar. Would be grateful of any input/suggestions on how to resolve these conflicts or if there is a better approach for how to add this support in the first place. 

Side effects

This is currently introducing reduce/reduce conflicts I am unsure how to resolve.


EmmaLouise2018:
This issue is better resolved in the following pr https://github.com/pingcap/parser/pull/331

PR #287, opened 2019-04-12
*: add table option: pre_split_regions

crazycs520:
This is for pre-split region when create table with shard_row_id_bits.

### Check List 

Tests 

 - Unit test

Code changes

Side effects


misc.go
```diff
@@ -408,6 +408,7 @@ var tokenMap = map[string]int{
 	"QUERIES":                  queries,
 	"QUICK":                    quick,
 	"SHARD_ROW_ID_BITS":        shardRowIDBits,
+	"PRE_SPLIT_REGIONS":        preSplitResions,
```
kennytm:
```suggestion
	"PRE_SPLIT_REGIONS":        preSplitRegions,
```

Fix the PR title too.

parser.y
```diff
@@ -203,6 +203,7 @@ import (
 	primary			"PRIMARY"
 	procedure		"PROCEDURE"
 	shardRowIDBits		"SHARD_ROW_ID_BITS"
+	preSplitResions		"PRE_SPLIT_REGIONS"
```
```suggestion
	preSplitRegions		"PRE_SPLIT_REGIONS"
```

ast/ddl.go
```diff
@@ -1326,6 +1327,10 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
 		ctx.WriteKeyWord("SHARD_ROW_ID_BITS ")
 		ctx.WritePlain("= ")
 		ctx.WritePlainf("%d", n.UintValue)
+	case TableOptionPreSplitRegion:
+		ctx.WriteKeyWord("PRE_SPLIT_REGIONS ")
+		ctx.WritePlain("= ")
+		ctx.WritePlainf("%d", n.UintValue)
```
kennytm: Let's combine this with the statement above.

```go
		ctx.WritePlainf("= %d", n.UintValue)
```

crazycs520: done.

PR #290, opened 2019-04-16
parser, ast: add ColumnOptionCollate for all column types (#260)

bb7133:
Cherry-pick https://github.com/pingcap/parser/pull/260


bb7133:
PTAL @crazycs520 @winkyao

PR #291, opened 2019-04-16
*: Add "show open tables" support

july2993:
*: Add "show open tables" support like in mysql: https://dev.mysql.com/doc/refman/5.7/en/show-open-tables.html

change parser.y support "show open tables" and add tests

### Check List 

Tests 

 - Unit test

Code changes

 

Side effects

 Related changes


parser_test.go
```diff
@@ -2673,6 +2673,21 @@ func (s *testParserSuite) TestSessionManage(c *C) {
...
+	parser := parser.New()
+
+	tests := []string{
+		`SHOW OPEN TABLES`,
+		`SHOW OPEN TABLES IN test`,
+		`SHOW OPEN TABLES FROM test`,
+	}
+
+	for _, test := range tests {
+		_, _, err := parser.Parse(test, "", "")
```
kennytm: Why not write an `s.RunTest` like the `TestSessionManage` above?

july2993: refine in https://github.com/pingcap/parser/pull/291/commits/9350018023bbb7d841df9624ce3797a7a1450ced

tiancaiamao:
LGTM

PR #292, opened 2019-04-16
 parser: support show analyze status

lamxTyler:

Support new syntax for show analyze status.
For https://github.com/pingcap/tidb/pull/10172

Update the parser for show statement.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - None

Related changes


parser_test.go
```diff
@@ -717,6 +717,7 @@ func (s *testParserSuite) TestDBAStmt(c *C) {
 		// for show pump/drainer status.
 		{"show pump status", true, "SHOW PUMP STATUS"},
 		{"show drainer status", true, "SHOW DRAINER STATUS"},
+		{"show analyze status", true, "SHOW ANALYZE STATUS"},
```
kennytm: Please also add test cases involving `ShowLikeOrWhereOpt`

```sql
SHOW ANALYZE STATUS LIKE '%pattern%';
SHOW ANALYZE STATUS WHERE `something` = 'stuff';
```

PR #293, opened 2019-04-17
*: add table option: wait_split_finish

crazycs520:
Add a table option `wait_split_finish`.
This is for PR https://github.com/pingcap/tidb/pull/10138. 
`wait_split_finish` means wait pre-split and scatter finish before return client.

### Check List 

Tests 

 - Unit test

Code changes


misc.go
```diff
@@ -409,6 +409,7 @@ var tokenMap = map[string]int{
 	"QUICK":                    quick,
 	"SHARD_ROW_ID_BITS":        shardRowIDBits,
 	"PRE_SPLIT_REGIONS":        preSplitRegions,
+	"WAIT_SPLIT_FINISH":        waitSplitFinish,
```
winkyao: `WAIT_SPLIT_REGION` is better?

morgo:
I read https://github.com/pingcap/tidb/pull/10138 - I am curious why don't we pre-split regions by default synchronously, is it for backward compatibility or the use case that a user needs to create a lot of tables?

It becomes hard for usability if there are three options the user needs to remember:
- shard-id-bits
- pre-split-regions
- wait-split-finish

I like the names you have chosen though, they make sense to me.

crazycs520:
@morgo , Because scatter region maybe spend a lot of time，in my local test, I have 1 TiDB, 1 PD, 3 TiKV, then pre-split 8 regions will spend 30 ~ 75s,  so we hope client can control this. 

There are 3 options, but you can only use the related options that you need.

model/model.go
```diff
@@ -218,6 +218,9 @@ type TableInfo struct {
 	// The pre-split region num is 2^(PreSplitRegions-1).
 	// And the PreSplitRegions should less than or equal to ShardRowIDBits.
 	PreSplitRegions uint64 `json:"pre_split_regions"`
+	// WaitSplitFinish specify the pre-split region to do with synchronous or asynchronous.
+	// false means pre-split region with asynchronous.
+	WaitSplitFinish bool `json:"wait_split_finish"`
```
winkyao: We can put this option as a session variable, it is weird to be a table option. And we should not add this field in the tableInfo.

crazycs520: em... @morgo, Do you prefer session variable or table option to control synchronously/asynchronism split?

morgo: +1 for session var

PR #295, opened 2019-04-18
charset: add GetSupportedCollations to get only supported collations

bb7133:
Please check corresponding TiDB PR: https://github.com/pingcap/tidb/pull/10186 for the purposes.

- To fix the inconsistency problems, some const table on Charset and Collations are modified.
- A new `GetSupportedCollations` is added.
- Refined some func/vars namings 

### Check List

Tests 

 - Unit test

Code changes

Related changes


mysql/charset.go
```diff
@@ -574,15 +574,15 @@ const (
...
-	DefaultCollationID      = 46
-	Latin1CollationID       = 8
-	ASCIICollationID        = 11
-	UTF8CollationID         = 33
-	UTF8MB4CollationID      = 45
-	BinaryCollationID       = 63
-	UTF8DefaultCollation    = "utf8_bin"
-	UTF8MB4DefaultCollation = "utf8mb4_bin"
-	DefaultCollationName    = UTF8MB4DefaultCollation
+	DefaultCollationID        = 46
```
lonng: Has any risk to change collation id?

charset/charset.go
```diff
@@ -52,6 +53,15 @@ var charsetInfos = []*Charset{
 	{CharsetBin, CollationBin, make(map[string]*Collation), "binary", 1},
 }
 
+// All the names supported collations should be in the following table.
+var supportedCollationNames = map[string]interface{}{
```
winkyao: you can use `map[string]struct{}` instead, struct{} is zero costs.

mysql/charset.go
bb7133: Here only changes the default collation, IDs of collation are not changed, so there's no risk.

charset/charset.go
bb7133: thank you!

mysql/charset.go
bb7133: BTW, the previous naming is misleading, that's why I renamed those variables.

bb7133:
addressed, PTAL @lonng @winkyao

mysql/charset.go
tiancaiamao: The ID is not changed?

I found collation will be translate to proto, which may be persistent, so it's better not  to change the ID

https://github.com/pingcap/parser/pull/295/files#diff-c31b8654bc3f8f52137403cf3dc617faL724

bb7133: Hi @tiancaiamao The change is the default collation for a given charset, for example, the change of `UTF8MB4CollationID`(before rename) from `45` to `46` means changes default collation of `utf8mb4` from `utf8mb4_general_ci` to `utf8mb4_bin`:

```
-	Latin1CollationID       = 8		+Latin1DefaultCollationID  = 47     // Change default collation of `latin1` from `latin1_swedish_ci` to `latin1_bin`
-	ASCIICollationID        = 11		+ASCIIDefaultCollationID   = 65     // Change the default collation of `ascii` from `ascii_general_ci` to `ascii_bin`
-	UTF8CollationID         = 33		+UTF8DefaultCollationID    = 83     // Change the default collation of `utf8` from `utf8_general_ci` to `utf8_bin`
-	UTF8MB4CollationID      = 45		+UTF8MB4DefaultCollationID = 46     // Change the default collation of `utf8mb4` from `utf8mb4_general_ci` to `utf8mb4_bin`
-	BinaryCollationID       = 63		+BinaryDefaultCollationID  = 63     // The default collation of `binary` is not changed.
```

bb7133: > I found collation will be translate to proto, which may be persistent, so it's better not to change the ID
> 
> https://github.com/pingcap/parser/pull/295/files#diff-c31b8654bc3f8f52137403cf3dc617faL724

`mysql.CollationNames` table is not changed, neither the value of `mysql.BinaryCollationID`.

Again, the old namings are really misleading, PTAL

tiancaiamao:
LGTM

PR #297, opened 2019-04-19
*: Add split index region syntax support.

crazycs520:
*: Add split index region syntax support.

This PR support split index region by SQL.  Syntax like below: 

```sql
SPLIT TABLE table_name INDEX index_name BY (val1...), (val2...),(val3...) ...
```

**example**
```sql
split table t1 idx1 (100), (1000), (10000);   
//range is : 0~100, 100 ~1000, 1000~ 10000 ， 10000 ~ maxInt64

split table t1 idx2 ("a", "2000-01-01 00:00:01"),  ("b", "2019-04-17 14:26:19"),  ("c", "");  
// range is: minIndexValue ~ ("a", "2000-01-01 00:00:01"),
("a", "2000-01-01 00:00:01") ~ ("b", "2019-04-17 14:26:19"),
("b", "2019-04-17 14:26:19") ~ ("c", ""),
("c", "") ~ maxIndexValue.
```

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -637,6 +638,7 @@ import (
 	RevokeStmt			"Revoke statement"
 	RevokeRoleStmt      "Revoke role statement"
 	RollbackStmt			"ROLLBACK statement"
+	SplitIndexRegionStmt		"split index region statement"
```
winkyao: `split` --> `Split`

parser.y
```diff
@@ -1443,6 +1445,24 @@ RecoverTableStmt:
         }
     }
 
+/*******************************************************************
+ *
+ *  Split index region Statement
```
winkyao: Statement --> statement

crazycs520: done.

parser.y
crazycs520: done.

ast/dml.go
```diff
@@ -2300,3 +2301,61 @@ func (n *FrameBound) Accept(v Visitor) (Node, bool) {
...
+
+	Table      *TableName
+	IndexName  string
+	ValueLists [][]ExprNode
+}
+
+func (n *SplitIndexRegionStmt) Restore(ctx *RestoreCtx) error {
+	ctx.WriteKeyWord("SPLIT TABLE ")
+	if err := n.Table.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while restore SplitTableIndexRegionStmt.Table")
```
bb7133:
```suggestion
		return errors.Annotate(err, "An error occurred while restore SplitIndexRegionStmt.Table")
```

ast/dml.go
```diff
@@ -2300,3 +2301,61 @@ func (n *FrameBound) Accept(v Visitor) (Node, bool) {
...
+		if i != 0 {
+			ctx.WritePlain(",")
+		}
+		ctx.WritePlain("(")
+		for j, v := range row {
+			if j != 0 {
+				ctx.WritePlain(",")
+			}
+			if err := v.Restore(ctx); err != nil {
+				return errors.Annotatef(err, "An error occurred while restore SplitTableIndexRegionStmt.ValueLists[%d][%d]", i, j)
```
bb7133:
```suggestion
				return errors.Annotatef(err, "An error occurred while restore SplitIndexRegionStmt.ValueLists[%d][%d]", i, j)
```

parser.y
```diff
@@ -1443,6 +1445,24 @@ RecoverTableStmt:
         }
     }
 
+/*******************************************************************
+ *
+ *  Split index region statement
+ *
+ *  Example:
+ *      SPLIT TABLE table_name INDEX index_name SPLIT BY (val1...),(val2...)...
```
bb7133:
```suggestion
 *      SPLIT TABLE table_name INDEX index_name BY (val1...),(val2...)...
```

ast/dml.go
```diff
@@ -2300,3 +2301,61 @@ func (n *FrameBound) Accept(v Visitor) (Node, bool) {
...
+	ctx.WriteKeyWord("SPLIT TABLE ")
+	if err := n.Table.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while restore SplitIndexRegionStmt.Table")
+	}
+	ctx.WriteKeyWord(" INDEX ")
+	ctx.WriteName(n.IndexName)
+	ctx.WriteKeyWord(" BY ")
+	for i, row := range n.ValueLists {
+		if i != 0 {
+			ctx.WritePlain(",")
```
zimulala: It seems that the last row will add the `,` too， is it OK？
Please add tests for this function.

crazycs520: No, `,` will be added before `(value)`, and there is already a test in https://github.com/pingcap/parser/pull/297/files#diff-688912c3f38a8a80d6bdc16c02088d69R641

zimulala: Got it.

PR #300, opened 2019-04-25
Parser: refine `Restore` method for `LIMIT` clause

spongedu:
Ai is described in [MySQL Manual](https://dev.mysql.com/doc/refman/5.7/en/select.html) 
>For compatibility with PostgreSQL, MySQL also supports the LIMIT row_count OFFSET offset syntax.

For now, TiDB Restore SQL like `SELECT * FROM t LIMIT 1 OFFSET 1` as `SELECT * FROM t LIMIT 1, 1`. That's right semantically, but I think it's better to restore the SQL as what it is originally. 

1. Add a boolean field `ExplicitOffsetSyntax` in `Limit` struct to denote if it uses `postgressql compatible` syntax in limit.
2. Act specifically if  `ExplicitOffsetSyntax` is `true` in  `Limit.Restore`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


ast/dml.go
```diff
@@ -1686,21 +1686,32 @@ func (n *UpdateStmt) Accept(v Visitor) (Node, bool) {
 type Limit struct {
 	node
 
-	Count  ExprNode
-	Offset ExprNode
+	Count                ExprNode
+	Offset               ExprNode
+	ExplicitOffsetSyntax bool
```
kennytm: We could just always restore as `LIMIT x OFFSET y`, and remove this field.

spongedu: I think it's better to keep the restored SQL as what it is originally, so we may keep this field for both `LIMIT row_count  OFFSET offset ` syntax and `LIMIT offset, row_count`  syntax

kennytm: `Restore()` by design is allowed to change the output as long as they are equivalent. So I'd prefer not to complicate the structure.

spongedu: OK. I wonder if there's situation when we need the original SQL exactly.

tiancaiamao:
LGTM
We should encourage people to participate in.
As the criteria @shenli always likes:  at least  the result of this change is not worse than before.

PR #294, opened 2019-04-18
suspend unknown collation id 255 error

lysu:
### What problem does this PR solve? 

fixes https://github.com/pingcap/tidb/issues/9426

### What is changed and how it works?

support 255 collation

### Check List 

Tests 

 - Unit test

Code changes

 - N/A

Side effects

 - N/A

Related changes

 - N/A

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/294)

PR #301, opened 2019-04-25
*: fix upper-cased charset and collation name

bb7133:
This PR aims to fix the following 3 potential issues:

* Validate collate name in parser and convert the name to lower case, as how charset is handled:
https://github.com/pingcap/parser/blob/e2cdb851bce2a8fe2a45eb35ffb2527677720366/parser.y#L5887

* Added a new `TableInfoVersion3`

* Added `utf8mb4_0900_ai_ci` in `mysql` package, as a tiny fix for https://github.com/pingcap/parser/pull/294

As stated above

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - (Potentially) Breaking backward compatibility


charset/charset.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
```
lonng: The old version is the correct format.

bb7133: addressed, thanks!

bb7133:
addressed, PTAL @lonng @crazycs520

model/model.go
```diff
@@ -175,9 +175,15 @@ const (
 	// After version V2.1.2 (PR#8738) , TiDB add UTF8 check, then the user upgrade from v2.0.8 insert some UTF8MB4 characters will got error.
 	// This is not compatibility for user. Then we try to fix this in PR #9820, and increase the version number.
 	TableInfoVersion2 = uint16(2)
+	// TableInfoVersion3 means the table info version is 3.
+	// This version aims to deal with upper-cased charset name in TableInfo stored by versions prior to TiDB v2.1.9:
+	// TiDB always suppose all charsets / collations as lower-cased and try to convert them if they're not.
+	// However, the convert is missed in some scenarios before v2.1.9, so for all those tables prior to TableInfoVersion3, their
+	// charsets / collations will be converted to lower-case while loading from the storage.
+	TableInfoVersion3 = uint16(3)
```
crazycs520: Should we update column version too?

charset/charset_test.go
```diff
@@ -165,4 +178,5 @@ func BenchmarkGetCharsetDesc(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		GetCharsetDesc(cs)
 	}
+
```
crazycs520: remove this blank line.

model/model.go
bb7133: IMHO it's not necessary, after all string case convert is not a heavy opeartion, using only `TableInfoVersion` makes the codes simple and clear.

bb7133:
hi @crazycs520 , please check my replys for your comments, thanks!

charset/charset_test.go
bb7133: done

PR #304, opened 2019-04-26
Parser: Support SQL_BIG_RESULT, SQL_SMALL_RESULT, SQL_BUFFER_RESULT

### LINKED ISSUE #141: Support SQL_BIG_RESULT, SQL_SMALL_RESULT, SQL_BUFFER_RESULT
Via https://github.com/pingcap/tidb/issues/4735

These keywords are safe to parse but ignore in TiDB for compatibility.

b41sh:

Support `SQL_BIG_RESULT`, `SQL_SMALL_RESULT`, `SQL_BUFFER_RESULT` Syntax for TiDB Parser.  Fixed #141

Add related syntax in parser.y
Add SQL_BIG_RESULT, SQL_SMALL_RESULT, SQL_BUFFER_RESULT in SelectStmtOpts
Update SelectStmt.Restore

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -3169,9 +3178,7 @@ UnReservedKeyword:
 | "NONE" | "NULLS" | "SUPER" | "EXCLUSIVE" | "STATS_PERSISTENT" | "ROW_COUNT" | "COALESCE" | "MONTH" | "PROCESS" | "PROFILES"
 | "MICROSECOND" | "MINUTE" | "PLUGINS" | "PRECEDING" | "QUERY" | "QUERIES" | "SECOND" | "SEPARATOR" | "SHARE" | "SHARED" | "SLOW" | "MAX_CONNECTIONS_PER_HOUR" | "MAX_QUERIES_PER_HOUR" | "MAX_UPDATES_PER_HOUR"
 | "MAX_USER_CONNECTIONS" | "REPLICATION" | "CLIENT" | "SLAVE" | "RELOAD" | "TEMPORARY" | "ROUTINE" | "EVENT" | "ALGORITHM" | "DEFINER" | "INVOKER" | "MERGE" | "TEMPTABLE" | "UNDEFINED" | "SECURITY" | "CASCADED"
-| "RECOVER" | "CIPHER" | "SUBJECT" | "ISSUER" | "X509" | "NEVER" | "EXPIRE" | "ACCOUNT" | "INCREMENTAL"
-
-
+| "RECOVER" | "CIPHER" | "SUBJECT" | "ISSUER" | "X509" | "NEVER" | "EXPIRE" | "ACCOUNT" | "INCREMENTAL" | "SQL_BUFFER_RESULT"
```
kennytm: Include `SQL_BIG_RESULT` and `SQL_SMALL_RESULT` too

parser_test.go
```diff
@@ -2448,6 +2448,54 @@ func (s *testParserSuite) TestPriority(c *C) {
 	c.Assert(sel.SelectStmtOpts.Priority, Equals, mysql.HighPriority)
 }
 
+func (s *testParserSuite) TestSQLBigResult(c *C) {
```
kennytm: Let's merge all three into the same test function.

Also, I suggest we use the `s.RunTest(c, table)` method for testing instead of the `TestSQLNoCache` approach (see `TestPrepare`), which is used to test `Restore()`.

There is no need to verify the value of `sel.SelectStmtOpts.SQLBigResult` since it is only used for `Restore()` and otherwise ignored. 

BTW please add a test case for 

```sql
select sql_small_result sql_big_result sql_buffer_result 1;
```

parser.y
b41sh: `SQL_BIG_RESULT` and `SQL_SMALL_RESULT` are Reserved Words

parser_test.go
b41sh: ok, I will modify

parser.y
kennytm: Hmm okay.

morgo:
Thank you for working on this! It is good to have this for MySQL compatibility.

morgo:
@b41sh Can you fix the formatting of `parser.y`. From the diff on GitHub, you can see some of the indentation is incorrect.

b41sh:
ok

b41sh:
@morgo PTAL

morgo:
LGTM, but I am a product manager :-)

PTAL @lysu @tiancaiamao

tiancaiamao:
LGTM
Please address conflicts @b41sh

PR #305, opened 2019-04-28
 *: try to support lock/unlock tables function

crazycs520:
1. Help to support lock/unlock tables function in TiDB.

2. There are 4 table lock types: `READ`, `READ LOCAL`, `WRITE`, `WRITE LOCAL`. 
Attention: 
* The `WRITE LOCAL` is add in TiDB to support TiFlash usage. Mysql doesn't have `WRITE LOCAL`.
The `WRITE LOCAL` means the session with this lock has write/read permission, and the other session is still has read permission.
* The `READ LOCAL` is not supported in TiDB, only support the syntax currently. Because of we think is it useless.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


winkyao:
Please explain what is `WRITE LOCAL`

model/model.go
```diff
@@ -230,11 +231,101 @@ type TableInfo struct {
...
+	Lock *TableLockInfo `json:"Lock"`
 
 	// Version means the version of the table info.
 	Version uint16 `json:"version"`
 }
 
+// TableLockInfo provides meta data describing a table lock.
+type TableLockInfo struct {
+	Tp TableLockType
+	// Use array because there maybe multiple session both hold same read lock.
```
bb7133:
```suggestion
	// Use array because there may be multiple sessions holding the same read lock.
```

model/model.go
```diff
@@ -230,11 +231,101 @@ type TableInfo struct {
...
+	Sessions []SessionInfo
+	State    TableLockState
+	// TS use to record the timestamp of this table lock been public locked.
```
bb7133:
```suggestion
	// TS is used to record the timestamp this table lock been locked.
```

model/model.go
```diff
@@ -230,11 +231,101 @@ type TableInfo struct {
...
+}
+
+// SessionInfo contain the session ID and the server ID.
+type SessionInfo struct {
+	ServerID  string
+	SessionID uint64
+}
+
+func (s SessionInfo) String() string {
+	return "server:" + s.ServerID + "_session:" + strconv.FormatUint(s.SessionID, 10)
```
bb7133:
```suggestion
	return "server: " + s.ServerID + "_session: " + strconv.FormatUint(s.SessionID, 10)
```

model/model.go
```diff
@@ -230,11 +231,101 @@ type TableInfo struct {
...
+}
+
+// DBTableID is composed by schema ID and table ID.
+type DBTableID struct {
```
bb7133: Looks that this is not used anymore, right?

model/model.go
```diff
@@ -230,11 +231,101 @@ type TableInfo struct {
...
+		return "none"
+	}
+}
+
+// TableLockType is the type of the table lock.
+type TableLockType byte
+
+const (
+	TableLockNone TableLockType = iota
+	// TableLockRead means the session with this lock has read permission, other session can't read/write until the lock is released.
```
bb7133: IMHO we need to add more detailed explanations. For example, if the locks are exclusive, or can be shared.

Although the behavior of `LOCKS` will be explained in our official document, detailed comments are still useful here for the developers.

bb7133: I've checked the implementations, this comment is wrong(`other session can't read/write ...`), please fix it.

bb7133:
We should also add `ErrTableLocked`(from https://github.com/pingcap/tidb/pull/10343) related error definitions at parser side.

model/model.go
crazycs520: Sorry for my mistakes, PTAL again.

model/model.go
```diff
@@ -230,11 +231,98 @@ type TableInfo struct {
...
+	// TableLockRead means the session with this lock can read the table (but not write it).
+	// Multiple sessions can acquire a READ lock for the table at the same time.
+	// Other sessions can read the table without explicitly acquiring a READ lock.
+	TableLockRead
+	// TableLockReadLocal is not supported.
+	TableLockReadLocal
+	// TableLockWrite means only the session with this lock has write/read permission.
+	// Only the session that holds the lock can access the table. No other session can access it until the lock is released.
+	TableLockWrite
+	// TableLockWriteLocal means the session with this lock has write/read permission, and the other session is still has read permission.
```
bb7133:
```suggestion
	// TableLockWriteLocal means the session with this lock has write/read permission, and the other session still has read permission.
```

bb7133:
PTAL @winkyao @zimulala

PR #307, opened 2019-04-28
Add begin lock syntax for pessimistic transaction.

coocood:

Support pessimistic transaction.

- Add a new syntax `begin lock`
- Add a bool field in `BeginStmt`

### Check List 

Tests 

 - Integration test


tiancaiamao:
LGTM

PR #308, opened 2019-04-29
*: add syntax for split index region in create table statement.

crazycs520:
add syntax for split index region in create table statement.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


tiancaiamao:
What's the tidb pr which is using this one?

PR #311, opened 2019-05-01
Fix handling of FLOAT(p), FLOAT(m,d) and DECIMAL(m,0) types esp. w.r.t. Restore()

### LINKED ISSUE #310: The column type `DECIMAL(20,0)` is restored incorrectly as `DECIMAL`
## Bug Report

1. What did you do?

```go
package main

import (
	"fmt"
	"github.com/pingcap/parser"
	"github.com/pingcap/parser/format"
	"strings"
)

func main() {
	p := parser.New()
	nodes, _ := p.ParseOneStmt(`
		create table precise_types (
			c DECIMAL(20,0) NOT NULL,
			d DOUBLE(20,0) NOT NULL
		);
	`, "", "")

	var s strings.Builder
	ctx := format.NewRestoreCtx(format.DefaultRestoreFlags, &s)
	nodes.Restore(ctx)

	fmt.Println(s.String())
}
```

Prints

```sql
CREATE TABLE `precise_types` (`c` DECIMAL(20) NOT NULL,`d` DOUBLE(20,0) NOT NULL)
```

3. What did you see instead?

The precision is just directly omitted.  

```sql
CREATE TABLE `precise_types` (`c` DECIMAL NOT NULL,`d` DOUBLE NOT NULL)
```

master (366172dc2adfff9e9a4fd77f7f56f4d99ab30a3e)

kennytm:

1. Fix #310 regarding `Restore()` of `FLOAT(m,0)`, `DOUBLE(m,0)` and `DECIMAL(m,0)`.
2. Ensure `FLOAT(m,d)` is always a `FLOAT`, never a `DOUBLE`.
3. Ensure all of `FLOAT(0)` to `FLOAT(24)` map to the same `FLOAT`, and all of `FLOAT(25)` to `FLOAT(53)` map to the same `DOUBLE`.

1. Refactored `(*FieldType).Restore()` such that all types share the same `Tp(m,d)` restore logic (except `TIME(d)` and friends).
2. Changed the parser logic for `FloatingPointType` to distinguish between the `FLOAT(m,d)` and `FLOAT(p)` cases.
3. Added a bunch of test cases about the behavior of `Tp(m,d)` for the rest of types.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

    * `FLOAT(24,1)` produces a `FLOAT` instead of a `DOUBLE`, matching behavior of MySQL 8 (but note that `FLOAT(m,d)` is deprecated since MySQL 8.0.17).

Related changes


kennytm:
PTAL @tiancaiamao

tiancaiamao:
LGTM
PTAL @lysu

PR #314, opened 2019-05-05
*: fix upper-cased charset and collation name (#301)

bb7133:
Cherry-pick https://github.com/pingcap/parser/pull/301


bb7133:
PTAL @crazycs520 @winkyao

PR #235, opened 2019-03-07
type/compatibility: fix show create table output of generated column

pingyu:
<!--
Thank you for contributing to TiDB SQL Parser! Please read [this](https://github.com/pingcap/parser/blob/master/README.md) document **BEFORE** filing this PR.
-->

### What problem does this PR solve? <!--add issue link with summary if exists-->
fix incompatible output of show create table for generated column

### What is changed and how it works?
Change restore format for `BinaryOperationExpr` and `FieldType`
Related issue https://github.com/pingcap/tidb/issues/9457
Related TiDB PR https://github.com/pingcap/tidb/pull/9608

### Check List <!--REMOVE the items that are not applicable-->

Tests <!-- At least one of them must be included. -->

 - Unit test
 - Manual test (add detailed scripts or steps below)
 In TiDB
```sql
mysql> 
mysql> CREATE TABLE t20( first_name varchar(10), last_name varchar(10), full_name varchar(255) AS (CONCAT(first_name,' ' ,last_name) ));
Query OK, 0 rows affected (0.01 sec)

mysql> show create table t20;
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                      |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t20   | CREATE TABLE `t20` (
  `first_name` varchar(10) DEFAULT NULL,
  `last_name` varchar(10) DEFAULT NULL,
  `full_name` varchar(255) GENERATED ALWAYS AS (concat(`first_name`, ' ', `last_name`)) VIRTUAL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> CREATE TABLE t21 ( `a` int(11) DEFAULT NULL, `b` char(20) GENERATED ALWAYS AS (cast(`a` AS CHAR)) VIRTUAL) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
Query OK, 0 rows affected (0.05 sec)

mysql> show create table t21;
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #315, opened 2019-05-05
type/compatibility: fix show create table output of generated column

bb7133:
Cherry-pick https://github.com/pingcap/parser/pull/235 to convert charset name to lower-case.


bb7133:
PTAL @crazycs520 @winkyao @jackysp

PR #316, opened 2019-05-06
parser: support mysql-compatible explain format

spongedu:
Support `MySQL` compatible explain format as is described in [MySQL Manual](https://dev.mysql.com/doc/refman/5.7/en/explain.html). such as:
```sql
explain format=traditional select * from t;
explain format=json select * from t;
```

1. `TRADITIONAL` type would be translate as `row` to be compatible with `TiDB` current implementation. 
2. `JSON` type is implemented just in parser side and will be blocked during preprocessing in `TiDB`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N/A

Side effects

N/A

Related changes

N/A


spongedu:
@kennytm @tiancaiamao @morgo PTAL

spongedu:
@kennytm `'row'` is peculiar to TiDB , and is semantic equivalent to `TRADITIONAL` in MySQL, So I think it's ok to conflate the two.

kennytm:
@spongedu yes but the list of columns of TiDB's "row" and MySQL's TRADITIONAL format are different.

spongedu:
@kennytm I think the `format` section just specify how the results of `EXPLAIN` are displayed.  For example the `TRADITIONAL` display the results as rows, and `JSON` display the results as a json string. so are `'row'` and `'dot'`. 

As for the column count  issues, I think it's a **content** issue ， and is orthogonal to the format.

morgo:
LGTM

PR #318, opened 2019-05-07
parser: add support for `ALTER DATABASE` syntax

bb7133:
Add support for `ALTER DATABASE` syntax

Related issue: https://github.com/pingcap/tidb/issues/4641

Modified parser.y

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


ast/ddl.go
```diff
@@ -118,6 +118,42 @@ func (n *CreateDatabaseStmt) Accept(v Visitor) (Node, bool) {
...
+	ctx.WriteKeyWord("ALTER DATABASE")
+	if n.Name != "" {
+		ctx.WritePlain(" ")
+		ctx.WriteName(n.Name)
+	}
+	for _, option := range n.Options {
+		ctx.WritePlain(" ")
+		err := option.Restore(ctx)
+		if err != nil {
+			return errors.Trace(err)
```
kennytm: We typically use `errors.Annotate()` here

parser_test.go
```diff
@@ -1844,7 +1844,38 @@ func (s *testParserSuite) TestDDL(c *C) {
...
+		{"ALTER DATABASE DEFAULT CHARSET = 'utf8mb4' COLLATE = 'utf8_bin'", true, "ALTER DATABASE CHARACTER SET = utf8mb4 COLLATE = utf8_bin"},
+		{
+			"ALTER DATABASE DEFAULT CHARSET = 'utf8mb4' DEFAULT COLLATE = 'utf8mb4_general_ci' CHARACTER SET = 'utf8' COLLATE = 'utf8mb4_bin'",
+			true,
+			"ALTER DATABASE CHARACTER SET = utf8mb4 COLLATE = utf8mb4_general_ci CHARACTER SET = utf8 COLLATE = utf8mb4_bin",
+		},
+
+		{"ALTER TABLE ADD COLUMN (a SMALLINT UNSIGNED)", false, ""},
+		{"ALTER TABLE t ADD COLUMN (a SMALLINT UNSIGNED, b varchar(255))", true, "ALTER TABLE `t` ADD COLUMN (`a` SMALLINT UNSIGNED, `b` VARCHAR(255))"},
+		{"ALTER TABLE t ADD COLUMN (a SMALLINT UNSIGNED FIRST)", false, ""},
```
kennytm: These three seems irrelevant and duplicating the tests below.

ast/ddl.go
bb7133: addressed, thanks

parser_test.go
bb7133: done

bb7133:
PTAL @kennytm , thank you!

parser.y
```diff
@@ -606,7 +606,8 @@ import (
 
 %type	<statement>
 	AdminStmt			"Check table statement or show ddl statement"
-	AlterTableStmt			"Alter table statement"
+	AlterDatabaseStmt		"Alter table statement"
+	AlterTableStmt			"Alter database statement"
```
kennytm: These two are flipped.

parser.y
```diff
@@ -2036,6 +2040,36 @@ IndexColNameList:
...
+ *
+ *  ALTER {DATABASE | SCHEMA} [db_name]
+ *   alter_specification ...
+ *
+ *  alter_specification:
+ *   [DEFAULT] CHARACTER SET [=] charset_name
+ * | [DEFAULT] COLLATE [=] collation_name
+ *******************************************************************************************/
+ AlterDatabaseStmt:
+	"ALTER" DatabaseSym DBNameOpt DatabaseOptionListOpt
```
kennytm: I think it's better to split into two rules rather than passing an `[]interface{}`:

```
    "ALTER" DatabaseSym DBName DatabaseOptionList
|   "ALTER" DatabaseSym DatabaseOptionList
```

BTW why use a `DatabaseOptionListOpt` and then check `len(opts) == 0` instead of using `DatabaseOptionList`?

parser.y
bb7133: Yep... you are so careful

parser.y
```diff
@@ -3237,7 +3270,7 @@ Identifier:
 identifier | UnReservedKeyword | NotKeywordToken | TiDBKeyword
 
 UnReservedKeyword:
- "ACTION" | "ASCII" | "AUTO_INCREMENT" | "AFTER" | "ALWAYS" | "AVG" | "BEGIN" | "BIT" | "BOOL" | "BOOLEAN" | "BTREE" | "BYTE" | "CLEANUP" | "CHARSET"
+ "ACTION" | "ASCII" | "AUTO_INCREMENT" | "AFTER" | "ALWAYS" | "AVG" | "BEGIN" | "BIT" | "BOOL" | "BOOLEAN" | "BTREE" | "BYTE" | "CLEANUP" | "CHARSET" %prec charsetKwd
```
zimulala: Do we need to care about `"CHARACTER" "SET"`? I am not very clear about this.

winkyao: why we use `%prec` here?

bb7133: > Do we need to care about `"CHARACTER" "SET"`? I am not very clear about this.

@zimulala  The reason is `CHARACTER` is a reserved token but `CHARSET` is unreserved. For example, 

```
mysql> create database charset;
Query OK, 1 row affected

mysql> create database character;
(1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'character' at line 1")
```

bb7133: > why we use `%prec` here?

@zimulala @winkyao the reason is `CHARSET` is an unreserved keyword. There's a shift/reduce conflict when parsing `AlterDatabseStmt CHARSET`, YACC need to decide if charset is treated as `CHARACTER SET` keyword, or the database name.

BTW, this precedence leads to the following parsing result:

```
mysql> alter database charset=utf8;
(1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '=utf8' at line 1")

mysql> alter database character set=utf8;
Query OK, 1 row affected
```
And it is consistent with MySQL parser.

PR #321, opened 2019-05-09
add `split table ... between ... and` syntax

crazycs520:

Add new syntax for `split index region`.

```sql
SPLIT TABLE [table_name] INDEX [index_name] BETWEEN (lower_value) AND (upper_value) REGIONS [region_num]
```

Eg1: int type.

```sql
split table `t` index `idx1` between (0) and (1000) regions 5;
```

Upper sql will split `5` regions of the table `t` index `idx1` in range `0~1000`. Every region range like below:

Region1: -inf ~ 200

Region2: 200 ~ 400

Region3: 400 ~ 600

Region4: 600 ~ 800

Region5: 800 ~ +inf

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


parser.y
```diff
@@ -1478,13 +1481,37 @@ RecoverTableStmt:
...
+		$$ = &ast.SplitRegionStmt{
+			Table: $3.(*ast.TableName),
+			IndexName: model.NewCIStr($5),
+			SplitOpt: $6.(*ast.SplitOption),
+		}
+	}
+
+
+SplitOption:
+	"MIN" RowValue "MAX" RowValue "NUM" NUM
```
bb7133: Since `min` and `max` are not reserved keywords, IMHO it's not a good idea to use those 2 words here. @morgo PTAL

crazycs520: @morgo PTAL

morgo: what do you think about using `BETWEEN` here. i.e.:
`SPLIT TABLE [table_name] INDEX [index_name] BETWEEN min_value AND max_value REGIONS 5;`

This is more sql-like, since using min and max as function-like syntax feels weird to sql-writers, since this would usually be non-[sargable](https://dba.stackexchange.com/questions/162263/what-does-the-word-sargable-really-mean).

crazycs520: done.

ast/dml.go
```diff
@@ -2398,60 +2398,95 @@ func (n *FrameBound) Accept(v Visitor) (Node, bool) {
...
 	if !ok {
 		return n, false
 	}
 	n.Table = node.(*TableName)
-	for i, list := range n.ValueLists {
-		for j, val := range list {
-			node, ok := val.Accept(v)
-			if !ok {
-				return n, false
+	return v.Leave(n)
```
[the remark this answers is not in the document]
crazycs520: done.

ast/dml.go
```diff
@@ -2416,60 +2416,119 @@ func (n *FrameBound) Accept(v Visitor) (Node, bool) {
...
-	IndexName  string
+	Table     *TableName
+	IndexName model.CIStr
+
+	SplitOpt *SplitOption
+}
+
+type SplitOption struct {
+	Min        []ExprNode
+	Max        []ExprNode
```
zz-jason: I think `lower` and `upper` are more suitable to describe ranges

crazycs520: done.

PR #322, opened 2019-05-09
parser: add role to unreserved keyword

imtbkcat:
allowed `ROLE` as role name in TiDB

In MySQL test, there is a test case.
`create role role`
`ROLE` has become unreserved keyword since MySQL 8.0.1

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity


parser.y
```diff
@@ -400,8 +400,8 @@ import (
 	respect		"RESPECT"
 	replication	"REPLICATION"
 	reverse		"REVERSE"
-	role		"ROLE"
 	rollback	"ROLLBACK"
+	role		"ROLE"
```
kennytm: Moving this line shouldn't be needed

parser.y
```diff
@@ -6120,6 +6120,10 @@ RoleNameString:
 	{
 		$$ = $1
 	}
+|	"ROLE"
```
kennytm: It is very odd we special-case `ROLE` here, given that other unreserved keywords could also be used, e.g.

```sql
CREATE ROLE ROLLBACK;
CREATE ROLE ADMIN;
CREATE ROLE COPY;
```

I suggest we just change the `identifier` on Line 6119 to `Identifier`.

(Please also add test cases for these 3 statements.)

tiancaiamao:
Ping @imtbkcat

parser.y

imtbkcat: @kennytm using `Identifier` will cause conflict....

kennytm: If it is a shift/reduce conflict it could be resolved by specifying a precedence.

imtbkcat: it is a reduce/reduce conflict, because `NONE` is an unreserved keyword.

kennytm: 🤔 

So we need to exclude the unreserved keywords `EXECUTE`, `SUPER`, `PROCESS`, `RELOAD`, `EVENT` (for reduce/reduce on `GrantStmt`/`GrantRoleStmt`) and `NONE` (for reduce/reduce on `SetDefaultRoleOpt`). 

This would mean we should separate these six tokens from `UnReservedKeyword` into another list...

PR #328, opened 2019-05-15
charset: add ColumnDef.ResolveCharsetCollation()

bb7133:
This PR tries to resolve charset from the column's collation

A new function is added, but I'm not sure if this is the best way to solve it.

Tests 
 - Unit test

Code changes

Related changes


zz-jason:
@bb7133 any update?

PR #334, opened 2019-05-21
mysql: add write conflict error for latch and change the error code of ErrTxnTooLarge

jackysp:

Add write conflict error for latch and change the error code of ErrTxnTooLarge.

### DIFF
--- a/mysql/errcode.go
+++ b/mysql/errcode.go
@@ -923,6 +923,8 @@ const (
 	ErrMemExceedThreshold         = 8001
 	ErrForUpdateCantRetry         = 8002
 	ErrAdminCheckTable            = 8003
+	ErrTxnTooLarge                = 8004
+	ErrWriteConflictInTiDB        = 8005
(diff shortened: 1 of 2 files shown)


mysql/errcode.go
```diff
@@ -941,6 +943,4 @@ const (
 	ErrRegionUnavailable  = 9005
 	ErrGCTooEarly         = 9006
 	ErrWriteConflict      = 9007
-
-	ErrTxnTooLarge = 9500
```
coocood: Why change this?

jackysp: It is not a related change. But ErrTxnTooLarge is a TiDB error, it should not in the TiKV/PD section.

jackysp: It is better to keep the TiDB errors starts from 8000 and TiKV/PD's starts from 9000.

jackysp: I've updated the title and the description.

coocood:
LGTM

PR #335, opened 2019-05-22
support keyword `OPTIMISTIC` and `PESSIMISTIC`.

coocood:
support `BEGIN OPTIMISTIC` and `BEGIN PESSIMISTIC` syntax.

Add keywords

### Check List 

Tests 

 - Unit test


coocood:
@tiancaiamao @jackysp PTAL

ast/misc.go
```diff
@@ -365,12 +371,21 @@ func (n *ExecuteStmt) Accept(v Visitor) (Node, bool) {
...
+	switch n.Mode {
+	case Pessimistic:
+		ctx.WriteKeyWord("BEGIN ")
+		ctx.WriteKeyWord(Pessimistic)
+	case Optimistic:
+		ctx.WriteKeyWord("BEGIN ")
+		ctx.WriteKeyWord(Optimistic)
+	default:
+		ctx.WriteKeyWord("START TRANSACTION")
+	}
```
kennytm: Maybe combine the optimistic & pessimistic cases

```go 
if n.Mode == "" {
    ctx.WriteKeyWord("START TRANSACTION")
} else {
    ctx.WriteKeyWord("BEGIN ")
    ctx.WriteKeyWord(n.Mode)
}
```

But now we have 3 transaction options now: (default), pessimistic, optimistic. Why do we need to distinguish the default case from the optimistic mode?

coocood: When we specify pessimistic mode as the default mode in the config file, we should be able to run an optimistic transaction.

PR #338, opened 2019-05-23
ast: fix case expression format bug

crazycs520:
Fix `caseExpr` format bug.
```
((case when (c0 = 0) then 0 when (c0 > 0) then (c1 / c0) end))
```

before

```
"((CASE WHEN (`c0` = 0) THEN 0WHEN (`c0` > 0) THEN (`c1` / `c0`) END))"

```

now

```
"((CASE WHEN (`c0` = 0) THEN 0 WHEN (`c0` > 0) THEN (`c1` / `c0`) END))"
```

the differents is `THEN 0WHEN` and `THEN 0 WHEN`

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -289,13 +289,14 @@ func (n *CaseExpr) Restore(ctx *RestoreCtx) error {
 
 // Format the ExprNode into a Writer.
 func (n *CaseExpr) Format(w io.Writer) {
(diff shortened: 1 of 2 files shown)


crazycs520:
@XuHuaiyu

XuHuaiyu:
LGTM

PR #343, opened 2019-05-30
ast: add new field in create view stmt

lamxTyler:

Fix panic when create view using prepare.

For https://github.com/pingcap/tidb/pull/10651

> Previously, when we build the create view statement, the Fields of select statement will be rewritten with only name info, so the Expr will be nil when we try to execute the prepare statement, which will cause panic. This PR adds a new field in CreateViewStmt to record the select schema.
### Check List 

Tests 

 - No code

Code changes

 - None

Side effects

 - None

Related changes

 - None

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -1016,14 +1016,15 @@ func (n *TableToTable) Accept(v Visitor) (Node, bool) {
 type CreateViewStmt struct {
 	ddlNode
 
-	OrReplace   bool
-	ViewName    *TableName
-	Cols        []model.CIStr
-	Select      StmtNode
-	Algorithm   model.ViewAlgorithm
-	Definer     *auth.UserIdentity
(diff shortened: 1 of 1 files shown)


ast/ddl.go
```diff
@@ -1016,14 +1016,15 @@ func (n *TableToTable) Accept(v Visitor) (Node, bool) {
...
-	Select      StmtNode
-	Algorithm   model.ViewAlgorithm
-	Definer     *auth.UserIdentity
-	Security    model.ViewSecurity
-	CheckOption model.ViewCheckOption
+	OrReplace    bool
+	ViewName     *TableName
+	Cols         []model.CIStr
+	Select       StmtNode
+	SelectSchema []model.CIStr
```
kennytm: I suggest calling it `SchemaCols`

PR #344, opened 2019-06-01
*: support parsing all PARTITION BY syntax

kennytm:

Support all PARTITION BY syntax from [MariaDB](https://mariadb.com/kb/en/library/create-table/#partitions), since DM needs to use pingcap/parser to parse these upstream DDLs.

1. Parses all 5 types of partitions supported by MariaDB, i.e.
    * `HASH`
    * `KEY`
    * `RANGE` — and the `VALUES LESS THAN` clauses
    * `LIST` — and the `VALUES IN` and `DEFAULT` clauses
    * `SYSTEM_TIME` — and the `HISTORY` and `CURRENT` clauses

2. The parse result is now saved entirely into the AST. Constructs like `SUBPARTITION` and options like `ENGINE` are now part of the Restore text.

3. As partitions options are a subset of table options, the following 4 new table options can now be parsed:
    * `TABLESPACE` = identifier
    * `DATA DIRECTORY` = string
    * `INDEX DIRECTORY` = string
    * `NODEGROUP` = number

4. Support parsing and restoring `ALTER TABLE ... PARTITION BY` syntax. As a side effect, `ALTER TABLE t1,;` is no longer valid (`ALTER TABLE t1;` is still valid).

5. In `ALTER TABLE ... DROP PARTITION` and `TRUNCATION PARTITION`, support parsing and restoring multiple partition names.

6. Support restoring `ALTER TABLE ... ENABLE KEYS` and `DISABLE KEYS` (for easy testing).

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


kennytm:
Circle-CI is not running. Let me push from a fork instead.

PR #345, opened 2019-06-01
*: support parsing all PARTITION BY syntax

kennytm:

Support all PARTITION BY syntax from [MariaDB](https://mariadb.com/kb/en/library/create-table/#partitions), since DM needs to use pingcap/parser to parse these upstream DDLs.

1. Parses all 5 types of partitions supported by MariaDB, i.e.
    * `HASH`
    * `KEY`
    * `RANGE` — and the `VALUES LESS THAN` clauses
    * `LIST` — and the `VALUES IN` and `DEFAULT` clauses
    * `SYSTEM_TIME` — and the `HISTORY` and `CURRENT` clauses

2. The parse result is now saved entirely into the AST. Constructs like `SUBPARTITION` and options like `ENGINE` are now part of the Restore text.

3. As partitions options are a subset of table options, the following 4 new table options can now be parsed:
    * `TABLESPACE` = identifier
    * `DATA DIRECTORY` = string
    * `INDEX DIRECTORY` = string
    * `NODEGROUP` = number

4. Support parsing and restoring `ALTER TABLE ... PARTITION BY` syntax. As a side effect, `ALTER TABLE t1,;` is no longer valid (`ALTER TABLE t1;` is still valid).

5. In `ALTER TABLE ... DROP PARTITION` and `TRUNCATION PARTITION`, support parsing and restoring multiple partition names.

6. Support restoring `ALTER TABLE ... ENABLE KEYS` and `DISABLE KEYS` (for easy testing).

Corresponding TiDB PR: pingcap/tidb#10672

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


kennytm:
PTAL @tiancaiamao @winkyao cc @csuzhangxc

tiancaiamao:
LGTM

parser_test.go
```diff
@@ -2999,33 +3012,129 @@ func (s *testParserSuite) TestTablePartition(c *C) {
...
+		{"create table t1 (a int, b int) partition by list (a) (partition x values in ((10, 20)))", false, ""},
+		{"create table t1 (a int, b int) partition by list columns (a, b) (partition x values in ((10, 20)))", true, "CREATE TABLE `t1` (`a` INT,`b` INT) PARTITION BY LIST COLUMNS (`a`,`b`) (PARTITION `x` VALUES IN ((10, 20)))"},
+		{"create table t1 (a int, b int) partition by list columns (a, b) (partition x values in (10, 20))", false, ""},
+		{"create table t1 (a int, b int) partition by list columns (a, b) (partition x values in (10, (20, 30)))", false, ""},
+		{"create table t1 (a int, b int) partition by list columns (a, b) (partition x values in ((10, 20), 30))", false, ""},
+		{"create table t1 (a int, b int) partition by list columns (a, b) (partition x values in ((10, 20), (30, 40, 50)))", false, ""},
+
+		// there must be at least one column/partition/value inside (...)
+		{"create table t1 (a int) partition by hash (a) ()", false, ""},
+		{"create table t1 (a int primary key) partition by key ()", true, "CREATE TABLE `t1` (`a` INT PRIMARY KEY) PARTITION BY KEY () PARTITIONS 1"},
```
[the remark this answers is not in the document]
kennytm: @tiancaiamao Yes it is valid, this means partition by the primary key or a non-null unique key.

https://dev.mysql.com/doc/refman/8.0/en/partitioning-key.html

> Where no column name is specified as the partitioning key, the table's primary key is used, if there is one. 
>
>  If there is no primary key but there is a unique key, then the unique key is used for the partitioning key: 
>
> However, if the unique key column were not defined as `NOT NULL`, then the previous statement would fail.

parser.y
```diff
@@ -2312,74 +2411,131 @@ PartitionDefinitionList:
...
 	}
 
 PartDefOption:
 	"COMMENT" EqOpt stringLit
 	{
-		$$ = $3
+		$$ = &ast.TableOption{Tp: ast.TableOptionComment, StrValue: $3}
 	}
-|	"ENGINE" EqOpt Identifier
+|	"ENGINE" EqOpt StringName
```
bb7133: Is there a particular reason why changing `Identifier` to `StringName`?

kennytm: @bb7133 It is valid to use a string here. Also make it possible to share the same rule for both table options and partition options.

```sql
create table t(a int) partition by hash(a) (partition `x` engine = 'innodb');
```

bb7133: Thank you!

PR #346, opened 2019-06-04
add MySQL extension assignment support for VariableAssignment

tangenta:
Support MySQL extension `:=` syntax in variable assignment.

Modified parser.y

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


zz-jason:
@tiancaiamao @kennytm PTAL

PR #350, opened 2019-06-11
*: support syntax for 'admin reload expr_pushdown_blacklist'

XuHuaiyu:
Support syntax for 'admin reload expr_pushdown_blacklist', which is used to reload the data of system table mysql.expr_pushdown_blacklist.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

N/A

Related changes


XuHuaiyu:
related-PR: https://github.com/pingcap/tidb/pull/10688

parser.y
```diff
@@ -6339,6 +6341,12 @@ AdminStmt:
 			ShowSlow: $4.(*ast.ShowSlow),
 		}
 	}
+|	"ADMIN" "RELOAD" "EXPR_PUSHDOWN_BLACKLIST"
```
zz-jason: should we also add a `"ADMIN" "SHOW" "EXPR_PUSHDOWN_BLACKLIST"` syntax support?

XuHuaiyu: I think that `select * from mysql.expr_pushdown_blacklist` may be enough.

zz-jason: OK. Anyway, We can add it later if necessary.

PR #365, opened 2019-06-21
add `hexLit` and `binaryList` support for `LOAD DATA`

imtbkcat:
TiDB cannot support `LOAD DATA` sql  which fields terminated by hex expression.
```
load data local infile '/home/pingcap/jeff-test/merchant.dmp' into table `merchant` fields terminated by X'01'
2018/06/29 13:29:22.709 conn.go:481: [warning] [8448] dispatch error:
```

Add `hexLit` and `binaryLit` for `FieldTerminator`, and convert hex expression to string during parsing.

After: 
```
mysql> load data local infile '~/1.csv' into table t_sp FIELDS TERMINATED BY X'2C2C2C2C' enclosed by '"';
Query OK, 3 rows affected (0.00 sec)
Records: 3  Deleted: 0  Skipped: 0  Warnings: 0
```

### Check List 

Tests 

 - Unit test


parser_test.go
```diff
@@ -487,7 +487,9 @@ func (s *testParserSuite) TestDMLStmt(c *C) {
...
 		{"load data local infile '/tmp/t.csv' into table t fields terminated by 'ab' enclosed by 'b' escaped by '*' ignore 1 lines (a,b)", true, "LOAD DATA LOCAL INFILE '/tmp/t.csv' IGNORE INTO TABLE `t` FIELDS TERMINATED BY 'ab' ENCLOSED BY 'b' ESCAPED BY '*' IGNORE 1 LINES (`a`,`b`)"},
 		{"load data local infile '/tmp/t.csv' into table t fields terminated by 'ab' enclosed by 'b' escaped by ''", true, "LOAD DATA LOCAL INFILE '/tmp/t.csv' IGNORE INTO TABLE `t` FIELDS TERMINATED BY 'ab' ENCLOSED BY 'b' ESCAPED BY ''"},
-
+		{"load data local infile '~/1.csv' into table `t_ascii` fields terminated by X'6B6B';", true, "LOAD DATA LOCAL INFILE '~/1.csv' IGNORE INTO TABLE `t_ascii` FIELDS TERMINATED BY 'kk'"},
+		{"load data local infile '~/1.csv' into table `t_ascii` fields terminated by X'6B6B' enclosed by X'0D';", true, "LOAD DATA LOCAL INFILE '~/1.csv' IGNORE INTO TABLE `t_ascii` FIELDS TERMINATED BY 'kk' ENCLOSED BY '\r'"},
+		{"load data local infile '~/1.csv' into table `t_ascii` fields terminated by X'6B6B' enclosed by X'0D0D';", false, ""},
```
jackysp: Please add some cases for enclosed and escaped.

parser.y
```diff
@@ -8885,6 +8886,20 @@ FieldItem:
...
 	}
 
+FieldTerminator:
+	stringLit
+	{
+		$$ = $1
+	}
+|	hexLit
+	{
+		$$ = ast.NewValueExpr($1).GetString()
```
kennytm: Could we use NewHexLiteral directly here?

https://github.com/pingcap/parser/blob/3b36f86d9b7bba02fef99748e3a98069708a64f3/ast/misc.go#L2035-L2039

imtbkcat: But `NewHexLiteral` will return a n interface. type assert may be very complex

kennytm: 🤔 Could we change the return type from `interface{}` to some real interface?

```go
type BinaryLiteral interface {
    ToString() string
}

var NewHexLiteral func(string) (BinaryLiteral, error) 
var NewBitLiteral func(string) (BinaryLiteral, error) 
```

parser_test.go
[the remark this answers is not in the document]
imtbkcat: added.

parser.y
imtbkcat: It's a good idea.

imtbkcat: @kennytm there are also some problems.

```
# github.com/pingcap/tidb/types/parser_driver [github.com/pingcap/parser/ast.test]
../../../../pkg/mod/github.com/pingcap/tidb@v0.0.0-20190321025159-e8299209340c/types/parser_driver/value_expr.go:50:20: cannot use func literal (type func(string) (interface {}, error)) as type func(string) (ast.BinaryLiteral, error) in assignment
```

imtbkcat: @kennytm I think using `NewValueExpr` is better here, because this expression will be eval and checked in parser. Using `NewHexLiteral` will convert `interface{}` to `types.HexLiteral`.

imtbkcat: @kennytm PTAL

kennytm: @imtbkcat I don't understand what do you mean. Aren't both NewValueExpr and NewHexLiteral defined outside the parser?

imtbkcat: @kennytm `NewValueExpr` works well here. But `NewHexLiteral` returns an `interface{}` , So, using `NewHexLiteral` need to convert `interface{}` to  `HexLiteral` which will import `types.HexLiteral` from tidb.

kennytm: @imtbkcat If we do the `BinaryLiteral` we'll change [these TiDB's code](https://github.com/pingcap/tidb/blob/2e369cd520b8fb2beb18bce0adb2bf526b104ca0/types/parser_driver/value_expr.go#L50-L57) to return `(ast.BinaryLiteral, error)`, not the other way around.

```diff
-	ast.NewHexLiteral = func(str string) (interface{}, error) {
+	ast.NewHexLiteral = func(str string) (ast.BinaryLiteral, error) {
		h, err := types.NewHexLiteral(str)
 		return h, err
 	}
-	ast.NewBitLiteral = func(str string) (interface{}, error) {
+	ast.NewBitLiteral = func(str string) (ast.BinaryLiteral, error) {
		b, err := types.NewBitLiteral(str)
 		return b, err
 	}
```

While `NewValueExpr` do work I don't feel this is a correct approach, since this goes through an unnecessary abstraction layer 😕.

Anyway, I realized that the lexer already invoked `types.NewHexLiteral`, so we should be able to simply use:

```suggestion
		$$ = lit.(ast.BinaryLiteral).ToString()
```

imtbkcat: @kennytm ok, i will fix it.

imtbkcat: @kennytm I meet some difficult in changing [this TiDB code](https://github.com/pingcap/tidb/blob/2e369cd520b8fb2beb18bce0adb2bf526b104ca0/types/parser_driver/value_expr.go#L50-L57). I add `ast.BinaryLiteral` interface and return type of `ast.NewHexLiteral`. But both side can not pass test. Is there any way to change both side?

kennytm: @imtbkcat The standard sequence of introducing breaking changes into parser is applicable here (see https://github.com/pingcap/parser/#how-to-update-parser-for-tidb).

1. Change `ast.NewHexLiteral` in parser to return `ast.BinaryLiteral` in parser.
2. Update the PR here (integration test will fail but it is expected and allowed).
3. Add a `replace` directive to tidb's `go.mod` pointing to your branch.
4. Submit a PR to TiDB (check-dev will fail but it is fine).
5. After the Parser PR is merged, remove the `replace` from tidb's `go.mod` and point the parser dependency to latest master instead.

However, as I mentioned above, we don't need to modify TiDB now:

1. Keep the signature `ast.NewHexLiteral` returning `interface{}`
2. The type of `$1` is already `types.HexLiteral`. That means `$1.(string)` is going to fail.
3. However, `types.HexLiteral` already implemented `ast.BinaryLiteral` by providing a `ToString()` method, so `$1.(ast.BinaryLiteral)` is going to succeed.

imtbkcat: @kennytm I have added `ToString` for `HexLiteral` and `BitLiteral`. PTAL: https://github.com/pingcap/tidb/pull/11029

imtbkcat: @kennytm how to update parser's go mod tidb to latest version ?

kennytm: @imtbkcat Parser doesn't depend on TiDB.

The reason the unit test is failing is because you are still using `$1.(string)` but `$1` isn't a string.

imtbkcat: @kennytm There is also an error `Panic: interface conversion: types.HexLiteral is not ast.BinaryLiteral: missing method ToString`

imtbkcat: but it works will in TiDB😂

kennytm: Maybe try updating `go.mod1` and `go.sum1`? 😂

imtbkcat: @kennytm how to upgrade `go.mod` in parser😂

kennytm: @imtbkcat Try to rename `go.mod1` and `go.sum1` to `go.mod` and `go.sum`, then do the regular update, then rename them back to `go.mod1` and `go.sum1` 😂

imtbkcat: @kennytm fixed, PTAL

PR #367, opened 2019-06-30
parser: parse SQL statement to the last character

tangenta:

Fix TiDB [issue #10906](https://github.com/pingcap/tidb/issues/10906).

When a SQL statement ends with an invalid part, i.e ` drop table if exists t'xyz`, Scanner's scanString() treat `'xzy` as an incomplete string, and returns `unicode.PlacementChar`.

However, the Lex() incorrectly converts it into a `0`, which means EOF, making it impossible to distinguish from `drop table if exists t`.

So, as long as an invalid token is spotted, the code indicating invalid should be returned.

### Check List 

Tests 

 - Integration test

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -211,9 +211,11 @@ func (s *Scanner) Lex(v *yySymType) int {
 	case quotedIdentifier:
 		tok = identifier
 	}
-	if tok == unicode.ReplacementChar && s.r.eof() {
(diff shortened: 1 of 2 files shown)


tangenta:
Since this PR can fail a malformed SQL test in TiDB, we should correct that first.

PR #370, opened 2019-07-01
ast: add BinaryLiteral interface

imtbkcat:
#365 need to change the `NewHexLiteral` driver's return type to an new interface type `BinaryLiteral`.
There are also some change in TiDB side.

Change return  type of `NewHexLiteral` and `NewBitLiteral`. Adding `BinaryLiteral` interface.

### Check List 

Tests 

 - Unit test

Code changes


imtbkcat:
@kennytm PTAL

PR #375, opened 2019-07-08
parser: add opt_rule_blacklist and `admin reload` it.

lzmhhh123:
Refer to: https://github.com/pingcap/tidb/pull/11096

Just add an `admin reload` in the parser and a new word in the ast.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

 - None


misc.go
```diff
@@ -589,6 +589,7 @@ var tokenMap = map[string]int{
 	"BINDING":                  binding,
 	"BINDINGS":                 bindings,
 	"EXPR_PUSHDOWN_BLACKLIST":  exprPushdownBlacklist,
+	"DISABLED_OPTIMIZE_LIST":   disabledOptimizeList,
```
zz-jason: Better to keep the same naming pattern. How about `OPT_RULE_BLACKLIST`?

lzmhhh123: It contains expression push down and other things like cascade rules in the future besides logical rules. Shall we also change the name?

zz-jason: I prefer to let this table to only contain the disabled rules. This the PR can be cherry-picked to release-3.0 and don't need to worry about the compatibility problems.

lzmhhh123: OK, if that's so. I should change it.

PR #377, opened 2019-07-08
test: count coverage for all packages when unit testing

kennytm:

By default, when testing package A with coverage, only lines in package A will be counted towards the final report, even if the tests called functions from package B.

In this project, however, almost all of our tests are written to `parser_test.go`. Thus even if `parser_test.go` invoked many functions in `ast/*` (package B), those coverages will be lost since `parser_test.go` is outside this directory (in package A).

We changed the default to tell Go to count coverage for all packages regardless of the source. The coverage report should be closer to reality.

It seems like covering all packages in `atomic` mode uses too much RAM for CircleCI when compiling (see attempts 2300, 2302 and 2305 in https://circleci.com/gh/pingcap/parser/tree/pull%2F377), so the test is split into two runs: once with race detector only and once with coverage only.

(An alternative is to split `parser_test.go` but I don't see much benefits doing so.)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,7 +1,8 @@
 {
 	  mv go.mod1 go.mod
 	  mv go.sum1 go.sum
-	  GO111MODULE=on go test -race -covermode=atomic -coverprofile=coverage.txt ./...
+	  GO111MODULE=on go test -race ./... &&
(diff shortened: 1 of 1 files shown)


kennytm:
PTAL @lamxTyler @jackysp

PR #378, opened 2019-07-09
add definition of json_objectagg 

hg2990656:
add the definition of `JSON_OBJECTAGG` in parser

### Check List 

Tests 
 - No code

Code changes


imtbkcat:
@hg2990656  Thanks for your contribution, please fix test and resolve conflict.

parser.y
```diff
@@ -4580,6 +4582,10 @@ SumExpr:
 	{
 		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
 	}
+|   builtinJsonObjectAgg '(' Expression ',' Expression ')'
```
bb7133: 1. Please use single tab rather than several spaces, just like the other builtin aggregate functions:

https://github.com/pingcap/parser/blob/169e7384f4cf04d92675bf65784974de77f69f95/parser.y#L4581

2. Please notice that the `BuggyDefaultFalseDistinctOpt` before `Expression`, which allows optional `ALL` token before the expression, for example:

```
SELECT JSON_OBJECTAGG(all c, all i) FROM t;
```
is valid and is equivalent to:
```
SELECT JSON_OBJECTAGG(c, i) FROM t;
```
3. It's better to add `OptWindowingClause` as well since window function is now supported in TiDB

parser.y
```diff
@@ -5359,6 +5362,10 @@ SumExpr:
 	{
 		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
 	}
+|   builtinJsonObjectAgg '(' BuggyDefaultFalseDistinctOpt Expression ',' BuggyDefaultFalseDistinctOpt Expression ')'  OptWindowingClause
```
zz-jason: please use tab instead of spaces to indent the code.

zz-jason:
@hg2990656 friendly ping, any update?

hg2990656:
When I make test, there are some error in test_driver/test_driver_mydecimal.go like expected ';', found _000_000. Is there anything else I need to modify?

tangenta:
> When I make test, there are some error in test_driver/test_driver_mydecimal.go like expected ';', found _000_000. Is there anything else I need to modify?

It is a new feature in go 1.13. Hold on, I will change it to the old representation.

parser.y
```diff
@@ -5712,6 +5714,10 @@ SumExpr:
 	{
 		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
 	}
+|	builtinJsonObjectAgg '(' BuggyDefaultFalseDistinctOpt Expression ',' BuggyDefaultFalseDistinctOpt Expression ')'  OptWindowingClause
+	{
+		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$3, $4}}
```
tangenta:
```suggestion
		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4, $7}}
```

misc.go
```diff
@@ -734,42 +734,44 @@ var tokenMap = map[string]int{
...
+	"SUBSTR":         builtinSubstring,
+	"SUBSTRING":      builtinSubstring,
+	"SUM":            builtinSum,
+	"SYSDATE":        builtinSysDate,
+	"SYSTEM_USER":    builtinUser,
+	"TRIM":           builtinTrim,
+	"VARIANCE":       builtinVarPop,
+	"VAR_POP":        builtinVarPop,
+	"VAR_SAMP":       builtinVarSamp,
+	"JSON_OBJECTAGG": builtinJsonObjectAgg,
```
tangenta: I don't think `JSON_OBJECTAGG` should be a builtin function here.

parser.y
```diff
@@ -678,6 +679,7 @@ import (
 	builtinUser
 	builtinVarPop
 	builtinVarSamp
+	builtinJsonObjectAgg
```
tangenta: Ditto.

hg2990656:
@zz-jason Sorry to bother you.Is there any problem now to add definition of json_objectagg in parser?There seems to be some error when Jenkins check my tidb code and replace my parser rep, so if this can be merged, it may be ok on Jenkins.

parser.y
```diff
@@ -6006,6 +6008,14 @@ SumExpr:
 	{
 		$$ = &ast.AggregateFuncExpr{F: $1, Args: []ast.ExprNode{$4}, Distinct: $3.(bool)}
 	}
+|	"JSON_OBJECTAGG" '(' BuggyDefaultFalseDistinctOpt Expression ',' BuggyDefaultFalseDistinctOpt Expression ')' OptWindowingClause
```
```suggestion
|	"JSON_OBJECTAGG" '(' Expression ',' Expression ')' OptWindowingClause
```

MySQL's [`json_objectagg`](https://dev.mysql.com/doc/refman/8.0/en/group-by-functions.html#function_json-objectagg) does not support DISTINCT. And indeed, adding DISTINCT to the key is useless, since JSON objects would remove duplicated keys anyway.

(Perhaps add a test case to ensure `JSON_OBJECTAGG(DISTINCT a, b)` and `JSON_OBJECTAGG(a, DISTINCT b)` are parse failure)

hg2990656: Thanks for your suggestions.I have tried it and json_objectagg does's support distinct but support all, so I add ALL to parser.y and add some test in parser_test.go as your suggestion.

PR #381, opened 2019-07-10
utils: `IsReadOnly` returns wrong result when checking SQLs like `explain analyze insert ...`

qw4990:
SQLs like `explain analyze insert into t values (1)` are not read-only, but the `IsReadOnly` function regards them as read-only queries.

Let `IsReadOnly `checks the `Analyze` flag and the internal `Stmt` of `ExplainStmt` to decide if it is `ReadOnly`.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/ast/util.go
+++ b/ast/util.go
@@ -28,10 +28,7 @@ func IsReadOnly(node Node) bool {
 		node.Accept(&checker)
 		return checker.readOnly
 	case *ExplainStmt:
-		if !st.Analyze {
-			return true
-		}
-		return IsReadOnly(st.Stmt)
+		return !st.Analyze || IsReadOnly(st.Stmt)
 	case *DoStmt:
 		return true
 	default:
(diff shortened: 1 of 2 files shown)


ast/util.go
```diff
@@ -27,7 +27,12 @@ func IsReadOnly(node Node) bool {
 
 		node.Accept(&checker)
 		return checker.readOnly
-	case *ExplainStmt, *DoStmt:
+	case *ExplainStmt:
+		if !st.Analyze {
+			return true
+		}
+		return IsReadOnly(st.Stmt)
```
kennytm: These 4 lines could simply be `return !st.Analyze || IsReadOnly(st.Stmt)`

qw4990: Nice.
It has been updated.

kennytm:
Unit test failures are addressed in #382.

qw4990:
CI problems have been resolved, PTAL @kennytm @XuHuaiyu

PR #382, opened 2019-07-10
Do not compile unit test in parallel

kennytm:

Do not compile tests in parallel, to avoid the compiler OOM inside CircleCI (which has 2 cores + 4 GiB RAM).

Add `-p 1` to limit the number of parallel compiler jobs. 

Merged the two tests back together since there is more space for memory.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,8 +1,7 @@
 {
 	  mv go.mod1 go.mod
 	  mv go.sum1 go.sum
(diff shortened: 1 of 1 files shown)


kennytm:
PTAL @lamxTyler @jackysp @qw4990

qw4990:
LGTM

PR #383, opened 2019-07-10
parser: support DELETE FROM ... AS ... syntax

kennytm:

Fix issue pingcap/tidb#11173. The `DELETE FROM ... AS ...` syntax was introduced in [MySQL 8.0.16](https://dev.mysql.com/doc/refman/8.0/en/delete.html).

Recognize the new syntax in the parser.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

    * May need to update https://pingcap.com/docs/v3.0/reference/sql/statements/delete/ if it is not auto-generated.


kennytm:
(Integration test requires pingcap/tidb#11184 to pass)

kennytm:
PTAL @tiancaiamao @shenli

imtbkcat:
Hi kennytm, i try to run this SQL in MySQL 8.0, but failed.

```
mysql> DELETE FROM x t WHERE t.a = 4;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 't WHERE t.a = 4' at line 1
```

kennytm:
@imtbkcat 

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-16.html

>  For consistency with the SQL standard and other RDBMS, table aliases are now supported in single-table as well as multi-table `DELETE` statements. (Bug #27455809)

bb7133:
hi @kennytm , please update TiDB test case:

(write_test.go:1747: `testSuite4.TestQualifiedDelete`)
```
1767     _, err := tk.Exec("delete from t1 as a where a.c1 = 1")
1768     c.Assert(err, NotNil)
```

PR #386, opened 2019-07-12
parser: parse and ignore 'PARTITION BY LIST ..."

tiancaiamao:

On the release-2.1 branch, this SQL fails to execute:

```
create table t (a int) /*!50100 partition by list (a) (
partition p0 values in (1) ENGINE = InnoDB,
partition p1 values in (29) ENGINE = InnoDB,
partition p2 values in (2) ENGINE = InnoDB,
partition p3 values in (3) ENGINE = InnoDB) */
```

Parse `partition by list` and ignore the result.

### Check List 

Tests 

 - Unit test
 - Integration test


parser.y
```diff
@@ -2038,6 +2039,9 @@ PartitionOpt:
 			Definitions:	defs,
 		}
 	}
+|	"PARTITION" "BY" "LIST" '(' Expression ')' PartitionNumOpt SubPartitionOpt PartitionDefinitionListOpt
+	{
+	}
```
kennytm: I think you should still assign `$$ = nil` like line 2003.

parser.y
```diff
@@ -2150,6 +2154,9 @@ PartDefValuesOpt:
 	{
 		$$ = $5
 	}
+|	"VALUES" "IN" '(' ExpressionList ')'
+	{
+	}
```
kennytm: Ditto (`$$ = nil`)

tiancaiamao:
Done @kennytm

PR #388, opened 2019-07-16
 parser: support more analyze options

lamxTyler:

Support more analyze options.

Add new syntax like `analyze table t with 4 buckets, 4 topn, 4 cmsketch width, 4 cmsketch depth`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - None

Related changes

For https://github.com/pingcap/tidb/pull/11278


ast/stats.go
```diff
@@ -32,13 +32,38 @@ type AnalyzeTableStmt struct {
...
+// Analyze option types.
+const (
+	AnalyzeOptNumBuckets = iota
+	AnalyzeOptNumTopN
+	AnalyzeOptCMSketchDepth
+	AnalyzeOptCMSketchWidth
+)
+
+// AnalyzeOptionString stores the string form of analyze options.
+var AnalyzeOptionString = map[AnalyzeOptionType]string{
```
winoros: We can use a slice directly?

lamxTyler:
@lzmhhh123 It does not matter if the internal implementation changed as long as the syntax is the same.

ast/stats.go
lamxTyler: Yes, but `map` could reduce our mental burden by maintaining their relationships explicitly.

eurekaka:
@lamxTyler CI failed, PTAL

lamxTyler:
@eurekaka It is expected since it removed some asts, that is it requires the tidb version of https://github.com/pingcap/tidb/pull/11278.

PR #390, opened 2019-07-17
prser: add alter user default role support

imtbkcat:
support gramma like `ALTER USER 'root'@'localhost' DEFAULT ROLE r1`
reference: https://dev.mysql.com/doc/refman/8.0/en/alter-user.html#alter-user-role

`ALTER USER ... DEFAULT ROLE` is much like `SET DEFAULT ROLE`.
So they use the same structure `SetDefaultRoleStmt`.
Adding `IsAlter` and `IfExist` to support `Restore` and `IF EXISTS` option.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


parser.y
```diff
@@ -6098,6 +6098,19 @@ SetDefaultRoleStmt:
 			UserList: $6.([]*auth.UserIdentity),
 		}
 	}
+| 	"ALTER" "USER" IfExists Username "DEFAULT" "ROLE" SetDefaultRoleOpt
+	{
+		user := make([]*auth.UserIdentity, 0, 1)
+		user = append(user, $4.(*auth.UserIdentity))
```
```go
user := []*auth.UserIdentity{$4.(*auth.UserIdentity)}
```

parser.y
```diff
@@ -6098,6 +6098,19 @@ SetDefaultRoleStmt:
...
+		tmp := $7.(*ast.SetRoleStmt)
+		$$ = &ast.SetDefaultRoleStmt{
+        	SetRoleOpt: tmp.SetRoleOpt,
+        	RoleList: tmp.RoleList,
+        	IsAlter: true,
+        	IfExist: $3.(bool),
+        	UserList: user,
+        }
```
kennytm: Fix indentation of lines 6107–6112 (reindent with tabs).

zz-jason:
@imtbkcat any update?

imtbkcat:
@zz-jason PTAL

PR #392, opened 2019-07-17
parser: support some new syntax

leoppro:
support followed syntax:

`ALTER TABLE t LOCK [=] lock_type`

`ALTER TABLE t ALGORITHM [=] algorithm_type`

`ALTER TABLE t ALTER COLUMN a SET DEFAULT (expr)`

`ALTER TABLE t ADD CONSTRAINT aaa CHECK (expr) [NOT [ENFORCED]]`
`CREATE TABLE t (a INT CHECK(expr) [NOT [ENFORCED]])`
`CREATE TABLE t (a INT CHECK(expr) NOT NULL)`
*(CHECK only parsed but not implement function)*
Here I refer to the implementation of MySQL:
https://github.com/mysql/mysql-server/blob/c000dcf1a9b2ee4a4f986dde08c1f6e1b6461a18/sql/sql_yacc.yy#L6733

BTW： Thanks for the guidance from @kennytm 

Improve the compatibility of tidb parser

### Check List 

Tests 

 - [x] Unit test
 - [ ] Integration test
 - [x] Manual test (add detailed scripts or steps below)
i tested `SET DEFAULT (expr)` in tidb which use the parser of this commi.
we support const value like : (1+1),(1*1). but we don't support functionexpr like: (now())
 - No code

Code changes

none
add a new lex token named 'enforced'
none

Side effects

 - Increased code complexity

Related changes


leoppro:
@kennytm @tiancaiamao PTAL

parser.y
```diff
@@ -1961,6 +2007,13 @@ ConstraintElem:
 			Refer:		$8.(*ast.ReferenceDef),
 		}
 	}
+|	"CHECK" '(' Expression ')' EnforcedOrNotOpt
+	{
+		/* Nothing to do now */
+		$$ = nil
```
kennytm: I think we should still parse this into a valid `ast.Constraint` instance (like the FOREIGN KEY above), so we don't need to add all those `nil` checking.

leoppro: if we return `*ast.Constraint{}` here, We need special handling of `Constraint.Restore`.
and I don't know what will tidb do when it finds an `ast.Constraint` with `type=0`

kennytm: I mean we do support CHECK in constraint in the AST and add an `ast.ConstraintCheck` type constant (like FOREIGN KEY and FULLTEXT), but we ignore or warn about this on TiDB level.

leoppro: fine

leoppro: PTAL again

leoppro: tidb will ignore `ConstraintCheck`,
see:https://github.com/pingcap/tidb/blob/source-code/ddl/ddl_api.go#L901

so, I think parser should warn this

parser.y
```diff
@@ -1961,6 +1995,15 @@ ConstraintElem:
 			Refer:		$8.(*ast.ReferenceDef),
 		}
 	}
+|	"CHECK" '(' Expression ')' EnforcedOrNotOpt
+	{
+		$$ = &ast.Constraint{
+			Tp:		ast.ConstraintCheck,
+			Expr:		$3.(ast.ExprNode),
+                }
```
kennytm: Reindent this line using tabs.

Also, the `ENFORCED`/`NOT ENFORCED` should be included (the default is `ENFORCED`) in the Restore() text. On MySQL it is restored as `/*!80016 NOT ENFORCED */`.

parser.y
```diff
@@ -1824,6 +1868,8 @@ ColumnOption:
 		// See https://dev.mysql.com/doc/refman/5.7/en/create-table.html
```
leoppro: we should implement `EnforcedOrNotOpt` for `"CHECK" '(' Expression ')'` item of `ColumnOption`,too.
but if I append `EnforcedOrNotOpt` to `"CHECK" '(' Expression ')'`, `goyacc` will find conflict around `NOT`,I have no idea how to fix it, can you help me? @kennytm

parser.y
leoppro: done

parser.y
kennytm: @leoppro What does it conflict with and is this a shift/reduce or reduce/reduce conflict? Can you assign a precedence an arm?

leoppro:
@imtbkcat PTAL

tiancaiamao:
integration-test fail

parser.y
```diff
@@ -1780,6 +1806,42 @@ PrimaryOpt:
...
+//	 This branch is needed to workaround the need of a lookahead of 2 for the grammar:
+//
+//	  { [NOT] NULL | CHECK(...) [NOT] ENFORCED } ...
+	"NOT" "NULL"
+	{
+		$$ = 0
+	}
+|	EnforcedOrNotOpt
+	{
+		if($1.(bool)){
```
```suggestion
		if ($1.(bool)) {
```

parser.y
```diff
@@ -1780,6 +1806,42 @@ PrimaryOpt:
...
+			$$ = 1
+		}else{
```
```suggestion
		} else {
```

parser.y
```diff
@@ -1820,11 +1882,30 @@ ColumnOption:
...
 		// The CHECK clause is parsed but ignored by all storage engines.
-		$$ = &ast.ColumnOption{}
+		// See the branch named `EnforcedOrNotOrNotNullOpt`.
+
+		optionCheck := &ast.ColumnOption{
+			Tp: ast.ColumnOptionCheck,
+			Expr: $3,
+			Enforced: true,
+		}
+		switch $5.(int){
```
```suggestion
		switch $5.(int) {
```

parser.y
```diff
@@ -1869,11 +1950,19 @@ VirtualOrStored:
 ColumnOptionList:
 	ColumnOption
 	{
-		$$ = []*ast.ColumnOption{$1.(*ast.ColumnOption)}
+		if columnOption,ok := $1.(*ast.ColumnOption); ok{
```
```suggestion
		if columnOption, ok := $1.(*ast.ColumnOption); ok {
```

parser.y
```diff
@@ -1869,11 +1950,19 @@ VirtualOrStored:
...
+			$$ = []*ast.ColumnOption{columnOption}
+		}else{
```
```suggestion
		} else {
```

parser.y
```diff
@@ -1869,11 +1950,19 @@ VirtualOrStored:
...
+			$$ = $1
+		}
 	}
 |	ColumnOptionList ColumnOption
 	{
-		$$ = append($1.([]*ast.ColumnOption), $2.(*ast.ColumnOption))
+		if columnOption,ok := $2.(*ast.ColumnOption); ok{
```
```suggestion
		if columnOption, ok := $2.(*ast.ColumnOption); ok {
```

parser.y
```diff
@@ -1869,11 +1950,19 @@ VirtualOrStored:
...
+			$$ = append($1.([]*ast.ColumnOption), columnOption)
+		}else{
```
```suggestion
		} else {
```

parser.y
```diff
@@ -1962,6 +2051,16 @@ ConstraintElem:
 			Refer:		$8.(*ast.ReferenceDef),
 		}
 	}
+|	"CHECK" '(' Expression ')' EnforcedOrNotOpt
+	{
+		$$ = &ast.Constraint{
+			Tp:		ast.ConstraintCheck,
+			Expr:		$3.(ast.ExprNode),
+			Enforced:	$5.(bool),
+                }
```
```suggestion
		}
```

leoppro:
@kennytm @tiancaiamao PTAL again

tiancaiamao:
LGTM

PR #396, opened 2019-07-24
parser: fix remove partitioning syntax

tangenta:
#### Fix compatibility problem about keyword `REMOVE PARTITIONING`

Link: pingcap/tidb#xx

#### MySQL Syntax:

```
alter_specification:
...
  | REMOVE PARTITIONING
```

#### Bad SQL Case:

```
alter table t remove partitioning;
alter table t lock = default remove partitioning;
```

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -1133,6 +1135,17 @@ AlterTableStmt:
...
+|	"ALTER" IgnoreOptional "TABLE" TableName AlterTableSpecListOpt "REMOVE" "PARTITIONING"
+	{
+		specs := $5.([]*ast.AlterTableSpec)
+		specs = append(specs, &ast.AlterTableSpec{
+				Tp: ast.AlterTableRemovePartitioning,
+			})
+		$$ = &ast.AlterTableStmt{
+			Table: $4.(*ast.TableName),
+			Specs: specs,
+		}
```
leoppro: throw a warn

parser_test.go
```diff
@@ -2158,6 +2158,10 @@ func (s *testParserSuite) TestDDL(c *C) {
 		{"recover table ", false, ""},
 		{"recover table t1 100", true, "RECOVER TABLE `t1` 100"},
 		{"recover table t1 abc", false, ""},
+
+		// for remove partitioning
+		{"alter table t remove partitioning", true, "ALTER TABLE `t` REMOVE PARTITIONING"},
+		{"alter table db.ident remove partitioning", true, "ALTER TABLE `db`.`ident` REMOVE PARTITIONING"},
```
leoppro: add more case like`alter table t lock = xxx, remove partitioning`

parser.y
```diff
@@ -1133,6 +1135,17 @@ AlterTableStmt:
 			Specs: specs,
 		}
 	}
+|	"ALTER" IgnoreOptional "TABLE" TableName AlterTableSpecListOpt "REMOVE" "PARTITIONING"
```
kennytm: Could you make this an `AlterTableSpec` rule on lines 1164–1392 instead?

parser.y
```diff
@@ -204,6 +204,7 @@ type OnDeleteUpdateDef struct {
 	over			"OVER"
 	packKeys		"PACK_KEYS"
 	partition		"PARTITION"
+	partitioning		"PARTITIONING"
```
leoppro: is `PARTITIONING` Reserved Keyword?

parser.y
```diff
@@ -216,6 +217,7 @@ type OnDeleteUpdateDef struct {
 	realType		"REAL"
 	references		"REFERENCES"
 	regexpKwd		"REGEXP"
+	remove			"REMOVE"
```
leoppro: is `REMOVE` Reserved Keyword?

parser.y
tangenta: @kennytm `REMOVE PARTITIONING` can't repeat in MySQL.

tangenta: I think `remove partitioning` should not be a kind of `AlterTableSpec`, for the reason that `Restore()` will insert a comma  before it.

kennytm: 🤔

leoppro: see:https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy#L7900

we can add a rule named `alter_table_partition_options`
```
alter_table_partition_options:
          partition_clause
          {
          }
        | REMOVE_SYM PARTITIONING_SYM
          {
          }
        ;
```

tangenta: @leoppro Sounds good. Let me try it.

leoppro:
LGTM 
@kennytm  PTAL

ast/ddl.go
```diff
@@ -2510,12 +2510,16 @@ func (n *PartitionMethod) acceptInPlace(v Visitor) bool {
 type PartitionOptions struct {
 	node
 	PartitionMethod
-	Sub         *PartitionMethod
-	Definitions []*PartitionDefinition
+	Sub                  *PartitionMethod
+	Definitions          []*PartitionDefinition
+	IsRemovePartitioning bool
```
kennytm: Don't modify `PartitionOptions`, this structure is used in `CREATE TABLE … PARTITION` as well.

This also makes no sense that placing `REMOVE PARTITIONING` inside `PartitionOptions` to fix the `Restore`, as it is _still_ being wrapped inside an `AlterTableSpec`. You could avoid the comma by changing these lines.

https://github.com/pingcap/parser/blob/e6a8bfdc5d3fe591e32138fd89c2c5b941e6639b/ast/ddl.go#L2069-L2073

leoppro: in mysql8 doc:
`REMOVE PARTITIONING` is a rule of `alter_specification`
see: https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

but in implement of mysql8 yacc:
`REMOVE PARTITIONING` is a kind of `partition_options`
See: https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy#L7900

so if we wanna full imitation from the implement of mysql8, it makes sense that modifies `PartitionOptions`.

however, I think it will be better that we follow the mysql8 doc, the syntaxes in mysql8 doc compatible with the implement of mysql8.

kennytm: @leoppro In the YACC implementation `REMOVE PARTITIONING` produces a [`PT_alter_table_remove_partitioning` instance](https://github.com/mysql/mysql-server/blob/4869291f7ee258e136ef03f5a50135fe7329ffb9/sql/parse_tree_nodes.h#L4051-L4057) and the `PARTITION BY` produces a [`PT_alter_table_partition_by` instance](https://github.com/mysql/mysql-server/blob/4869291f7ee258e136ef03f5a50135fe7329ffb9/sql/parse_tree_nodes.h#L4034-L4049), and both classes are subclasses of `PT_alter_table_action` shared by all `AlterTableSpec`-like things like [`PT_alter_table_add_column`](https://github.com/mysql/mysql-server/blob/4869291f7ee258e136ef03f5a50135fe7329ffb9/sql/parse_tree_nodes.h#L3729-L3747). So making `REMOVE PARTITION` an `AlterTableSpec` at AST level *is* implementing like MySQL 8.

The names of the rules inside `parser.y` is irrelevant here.

tangenta: Updated. How about this? PTAL @kennytm @leoppro

parser.y
```diff
@@ -1147,6 +1147,28 @@ AlterTableStmt:
 		}
 	}
 
+AlterTablePartitionOpt:
+	PartitionOpt
+        {
+        	if $1 != nil {
+	        	$$ = &ast.AlterTableSpec{
+               			Tp:        ast.AlterTablePartition,
```
leoppro: Reindent this line using tabs.

parser.y
```diff
@@ -1147,6 +1147,28 @@ AlterTableStmt:
...
+                        }
+        	} else {
+        		$$ = nil
+        	}
+
+        }
+| 	"REMOVE" "PARTITIONING"
+        {
+        	$$ = &ast.AlterTableSpec{
+        		Tp:        ast.AlterTableRemovePartitioning,
```
```suggestion
        		Tp: ast.AlterTableRemovePartitioning,
```

parser.y
```diff
@@ -1147,6 +1147,28 @@ AlterTableStmt:
...
+        	}
+		yylex.AppendError(yylex.Errorf("The REMOVE PARTITIONING clause is parsed but ignored by all storage engines."))
+                parser.lastErrorAsWarn()
```
leoppro: Reindent this line using tabs.

parser.y
```diff
@@ -1147,6 +1147,28 @@ AlterTableStmt:
...
+        }
```
leoppro: Reindent this line using tabs.

parser.y
```diff
@@ -1147,6 +1147,28 @@ AlterTableStmt:
...
+	        	$$ = &ast.AlterTableSpec{
+               			Tp:        ast.AlterTablePartition,
+                		Partition: $1.(*ast.PartitionOptions),
+                        }
+        	} else {
+        		$$ = nil
+        	}
+
+        }
+| 	"REMOVE" "PARTITIONING"
```
```suggestion
|	"REMOVE" "PARTITIONING"
```

PR #398, opened 2019-07-25
*: add where for show table region syntax

crazycs520:
* Add `where` for `show table region` syntax

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


parser_test.go
```diff
@@ -709,8 +709,10 @@ AAAAAAAAAAAA5gm5Mg==
 
 		// for show table regions.
 		{"show table t1 regions", true, "SHOW TABLE `t1` REGIONS"},
+		{"show table t1 regions where a=1", true, "SHOW TABLE `t1` REGIONS WHERE `a`=1"},
 		{"show table t1", false, ""},
 		{"show table t1 index idx1 regions", true, "SHOW TABLE `t1` INDEX `idx1` REGIONS"},
+		{"show table t1 index idx1 regions where a=2", true, "SHOW TABLE `t1` INDEX `idx1` REGIONS WHERE `a`=2"},
```
kennytm: Also add two test cases for `show table t1 regions like 'something%'`?

winkyao:
What is this PR used for？

crazycs520:
@winkyao This  use for support SQL like `show table t regions where scattering=1;`

parser_test.go
crazycs520: done.

kennytm: @crazycs520 I mean literal ``SHOW TABLE `t1` INDEX `idx1` REGIONS LIKE 'some%'``, not ``WHERE `a` LIKE 'some%'``. Don't use `ShowLikeOrWhereOpt` if you can't support that...

crazycs520: Great, thanks. PTAL

parser.y
```diff
@@ -6886,20 +6886,33 @@ ShowStmt:
...
 			Tp:	ast.ShowRegions,
 			Table:	$3.(*ast.TableName),
 		}
+		if $5 != nil {
+			stmt.Where = $5.(ast.ExprNode)
+		}
+		$$ = stmt
 	}
-|	"SHOW" "TABLE" TableName "INDEX" Identifier "REGIONS"
+|	"SHOW" "TABLE" TableName "INDEX" Identifier "REGIONS" ShowLikeOrWhereOpt
```
kennytm: Should this still be `ShowLikeOrWhereOpt`?

crazycs520: Done. Thanks.

parser.y
```diff
@@ -7612,20 +7612,33 @@ ShowStmt:
...
 	{
-		$$ = &ast.ShowStmt{
+		stmt := &ast.ShowStmt{
 			Tp:	ast.ShowRegions,
 			Table:	$3.(*ast.TableName),
 			IndexName: model.NewCIStr($5),
 		}
+		if $7 != nil {
+			if x, ok := $7.(*ast.PatternLikeExpr); ok && x.Expr == nil {
+				stmt.Pattern = x
```
kennytm: Remove this branch.

crazycs520: Done. Thanks.

crazycs520:
@kennytm PTAL

PR #399, opened 2019-07-25
parser: support cast as double

wshwsh12:
Related TiDB PR: https://github.com/pingcap/tidb/pull/11443
Support `cast(xxx as double)`.


wshwsh12:
> Please include a test case showing that `CAST(x AS DOUBLE)` is now a valid expression.
> 
> Rest LGTM.
> 
> `CAST(x AS FLOAT)`, `CAST(x AS FLOAT(p))` and `CAST(x AS REAL)` are also added in 8.0.17, would you mind at least supporting `CAST(x AS FLOAT)` in this or a future PR? Thanks.

UPDATE. PTAL @kennytm  
And I will try to support `CAST(x AS FLOAT)` in next PR.

PR #401, opened 2019-07-25
parser : add support for octet_length function 

gaoxingliang:
Add support for octet_length and Fix https://github.com/pingcap/tidb/issues/11195

add a ast node 

### Check List 

Tests 

 - Unit test added.

The PR in tidb is : https://github.com/pingcap/tidb/pull/11451


ast/functions.go
```diff
@@ -192,6 +192,7 @@ const (
 	Lcase           = "lcase"
 	Left            = "left"
 	Length          = "length"
+	OctetLength     = "octet_length"
```
SunRunAway: Please keep the alphabet order. 💯

gaoxingliang: fixed

Issue #404, opened 2019-07-29
Fix STATS_SAMPLE_PAGES syntax


leoppro:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```
ALTER TABLE t STATS_SAMPLE_PAGES 10
ALTER TABLE t STATS_SAMPLE_PAGES 0x0aaa , LOCK DEFAULT
ALTER TABLE t STATS_SAMPLE_PAGES = 123
ALTER TABLE t STATS_SAMPLE_PAGES = 11.11 AVG_ROW_LENGTH 11
ALTER TABLE t STATS_SAMPLE_PAGES = 10 MIN_ROWS = 12
```

Claim this subtask based on the comments below this issue.

leiysky:
Let me fix it

leoppro:
> Let me fix it

ok thanks

leiysky:
https://github.com/pingcap/parser/blob/2414ee76eeb7bd0b6659fda650c8b98df4945e4c/ast/ddl.go#L1589-L1594
Should I handle it in this way or append a warning in `parser.y`?

leoppro:
no, if we can't restore sql text from ast, we throw a warn in `Restore`
but in this case, the ast include intact sql stmt data, so we don't need to throw a warn.

PR #414, opened 2019-07-30
add support for STATS_SAMPLE_PAGES

leiysky:
MySQL syntax:
```
create_table_option:
...
        | STATS_SAMPLE_PAGES_SYM opt_equal ulong_num
        | STATS_SAMPLE_PAGES_SYM opt_equal DEFAULT_SYM
```
Issue: #404 

### Check List 

Tests 

 - Unit test


ast/ddl.go
```diff
@@ -1622,6 +1624,13 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
 	case TableOptionStorageMedia:
 		ctx.WriteKeyWord("STORAGE ")
 		ctx.WriteKeyWord(n.StrValue)
+	case TableOptionStatsSamplePages:
+		ctx.WriteKeyWord("STATS_SAMPLE_PAGES ")
+		ctx.WritePlainf("= %d", n.UintValue)
```
kennytm: If `n.UintValue == 0` the restored output should be `STATS_SAMPLE_PAGES = DEFAULT`

parser.y
```diff
@@ -7643,6 +7662,22 @@ TableOption:
 		yylex.AppendError(yylex.Errorf("The STORAGE clause is parsed but ignored by all storage engines."))
 		parser.lastErrorAsWarn()
 	}
+|	"SECONDARY_ENGINE" EqOpt "NULL"
+	{
+		// Parse it but will ignore it
+		// See https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy#L5977-L5984
+		$$ = &ast.TableOption{Tp: ast.TableOptionSecondaryEngine, StrValue: ""}
```
kennytm: `SECONDARY_ENGINE = NULL` and `SECONDARY_ENGINE = ''` have two different meanings and shouldn't be mixed.

leiysky: It should be in another pr, I'll revert it

PR #416, opened 2019-07-30
parser: support cast as float

wshwsh12:

Related TiDB PR: pingcap/tidb#11519
Support `cast(xxx as float)`.


parser.y
```diff
@@ -5083,6 +5083,22 @@ CastType:
...
+|	"FLOAT" FloatOpt
+	{
+		x := types.NewFieldType(mysql.TypeFloat)
+		fopt := $2.(*ast.FloatOpt)
+		if fopt.Flen >= 54 {
+			yylex.AppendError(ErrTooBigPrecision.GenWithStackByArgs(fopt.Flen,"CAST",53))
+		}
+		if fopt.Flen >= 25 && fopt.Flen <= 53 && !parser.lexer.GetSQLMode().HasRealAsFloatMode() {
+		    x = types.NewFieldType(mysql.TypeDouble)
+		    }
```
zz-jason: please indent code with tab.

parser.y
```diff
@@ -5083,6 +5083,22 @@ CastType:
...
 		$$ = x
 	}
+|	"FLOAT" FloatOpt
+	{
+		x := types.NewFieldType(mysql.TypeFloat)
+		fopt := $2.(*ast.FloatOpt)
+		if fopt.Flen >= 54 {
+			yylex.AppendError(ErrTooBigPrecision.GenWithStackByArgs(fopt.Flen,"CAST",53))
+		}
+		if fopt.Flen >= 25 && fopt.Flen <= 53 && !parser.lexer.GetSQLMode().HasRealAsFloatMode() {
```
kennytm: I don't think `REAL_AS_FLOAT` as any effect on the result type of `CAST(… AS FLOAT)`. It is only relevant for `CAST(… AS REAL)` which isn't implemented in this PR. 

Also, the `Flen <= 53` condition is redundant.

```suggestion
		if fopt.Flen >= 25 {
```

----

Demo. Note that only the type of `CAST(… AS REAL)` is affected by `REAL_AS_FLOAT` mode.

<table><tr><th>SQL Mode</th><th>Resulting schema</th></tr><tr><td>

```sql
set sql_mode = '';
create table a as select 
  cast(5 as float(24)) f24, 
  cast(5 as float(53)) f53, 
  cast(5 as real) r, 
  cast(5 as float) f, 
  cast(5 as double) d;
show create table a;
```

</td><td>

```sql
CREATE TABLE `a` (
  `f24` float NOT NULL DEFAULT '0',
  `f53` double NOT NULL DEFAULT '0',
  `r` double NOT NULL DEFAULT '0',
  `f` float NOT NULL DEFAULT '0',
  `d` double NOT NULL DEFAULT '0'
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
```

</td></tr><tr><td>

```sql
set sql_mode = 'real_as_float';
create table a as select 
  cast(5 as float(24)) f24, 
  cast(5 as float(53)) f53, 
  cast(5 as real) r, 
  cast(5 as float) f, 
  cast(5 as double) d;
show create table a;
```

</td><td>

```sql
CREATE TABLE `a` (
  `f24` float NOT NULL DEFAULT '0',
  `f53` double NOT NULL DEFAULT '0',
  `r` float NOT NULL DEFAULT '0',
  `f` float NOT NULL DEFAULT '0',
  `d` double NOT NULL DEFAULT '0'
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci 
```

</td></tr></table>

wshwsh12: You are right. And I update.
PTAL again. @kennytm

parser.y
```diff
@@ -5083,6 +5083,22 @@ CastType:
...
 	}
+|	"FLOAT" FloatOpt
+	{
+		x := types.NewFieldType(mysql.TypeFloat)
+		fopt := $2.(*ast.FloatOpt)
+		if fopt.Flen >= 54 {
+			yylex.AppendError(ErrTooBigPrecision.GenWithStackByArgs(fopt.Flen,"CAST",53))
+		}
+		if fopt.Flen >= 25 {
+		    x = types.NewFieldType(mysql.TypeDouble)
```
```suggestion
			x = types.NewFieldType(mysql.TypeDouble)
```

parser.y
```diff
@@ -5083,6 +5083,22 @@ CastType:
...
 		$$ = x
 	}
+|	"FLOAT" FloatOpt
+	{
+		x := types.NewFieldType(mysql.TypeFloat)
+		fopt := $2.(*ast.FloatOpt)
+		if fopt.Flen >= 54 {
+			yylex.AppendError(ErrTooBigPrecision.GenWithStackByArgs(fopt.Flen,"CAST",53))
+		}
+		if fopt.Flen >= 25 {
```
kennytm: Maybe use `else if` here, though it doesn't really matter 🙃

parser_test.go
```diff
@@ -351,6 +351,54 @@ func (s *testParserSuite) RunRestoreTest(c *C, sourceSQLs, expectSQLs string) {
...
+		}
+		c.Assert(err, IsNil, comment)
+		// restore correctness test
+		if t.ok {
+			s.RunRestoreTestInRealAsFloatMode(c, t.src, t.restore)
+		}
+	}
+}
+
+func (s *testParserSuite) RunRestoreTestInRealAsFloatMode(c *C, sourceSQLs, expectSQLs string) {
```
kennytm: Is this still needed?

wshwsh12: I think that it is needed.  
It can judge whether `REAL_AS_FLOAT` has any effect on `CAST(… AS FLOAT)`
And, if it will support `CAST(… AS REAL)` in future , it also can use the function to test.

parser.y
wshwsh12: update.

parser.y
wshwsh12: update.

PR #424, opened 2019-08-01
parser: add several optimizer hints

foreyes:

Add optimizer hints.

Change `TableOptimizerHint` struct, change `parser.y`, add tests. 

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported struct change

Related changes

 - Related TiDB PR: https://github.com/pingcap/tidb/pull/11673


foreyes:
Now only `TIDB_HJ`, `TIDB_INLJ`, `TIDB_SMJ`, `HASH_JOIN`, `INL_JOIN`, `SM_JOIN`, `HASH_AGG`, `STREAM_AGG` are in use. We can ignore other hints at this moment.

And integration test fails because it rely on https://github.com/pingcap/tidb/pull/11673.
We should merge this first, and then merge TiDB PR.

parser.y
```diff
@@ -6034,32 +6038,118 @@ TableOptimizerHintList:
...
+HintMemoryQuota:
+	NUM Identifier
+	{
+		// May change into MB/MiB or GB/GiB
+		switch model.NewCIStr($2).L {
+		case "m":
+			$$ = getUint64FromNUM($1)
+		case "g":
+			$$ = getUint64FromNUM($1) * 1024
+		default:
```
foreyes: Do we have better way to do this?

foreyes:
PTAL. @lzmhhh123 @lamxTyler

misc.go
```diff
@@ -327,6 +331,7 @@ var tokenMap = map[string]int{
 	"IS":                       is,
 	"ISSUER":                   issuer,
 	"ISOLATION":                isolation,
+	"USE_TOJA":                 hintUseToja,
```
lamxTyler: What is `toja`?

parser.y
```diff
@@ -6034,32 +6038,118 @@ TableOptimizerHintList:
...
+	}
+|	hintHASHAGG
+	{
+		$$ = &ast.TableOptimizerHint{HintName: model.NewCIStr($1)}
+	}
+|	hintSTREAMAGG
+	{
+		$$ = &ast.TableOptimizerHint{HintName: model.NewCIStr($1)}
+	}
+|	hintNoIndexMerge
```
lamxTyler: Why it cannot specify table names?

ast/misc.go
```diff
@@ -2026,25 +2026,56 @@ type TableOptimizerHint struct {
...
 		ctx.WritePlainf("%d", n.MaxExecutionTime)
-	case "tidb_hj", "tidb_smj", "tidb_inlj":
+	case "tidb_hj", "tidb_smj", "tidb_inlj", "hash_join", "sm_join", "inl_join":
 		for i, table := range n.Tables {
 			if i != 0 {
 				ctx.WritePlain(", ")
 			}
 			ctx.WriteName(table.String())
 		}
+	case "use_index_merge":
```
lamxTyler: These lines are not covered.

parser.y
```diff
@@ -552,11 +552,21 @@ import (
 	statsBuckets    "STATS_BUCKETS"
 	statsHealthy    "STATS_HEALTHY"
 	tidb		"TIDB"
-	tidbHJ		"TIDB_HJ"
```
zz-jason: old hints should not be removed.

ast/misc.go
foreyes: Oh my god...

parser.y
foreyes: Old hints will be treat as alias, the token will be same as `HASH_JOIN`.
You can check this out at the alias test.

misc.go
foreyes: Convert `in subquery` into inner join and aggregation. Seems like this one.
The name is advised by product team.

parser.y
foreyes: It's just a switch to disable this optimizer rule.
Also advised by product team...

misc.go
lzmhhh123: How about adding a comment for this abbreviation.

parser.y
```diff
@@ -6034,32 +6038,118 @@ TableOptimizerHintList:
...
+	}
+|	hintSTREAMAGG
+	{
+		$$ = &ast.TableOptimizerHint{HintName: model.NewCIStr($1)}
+	}
+|	hintNoIndexMerge
+	{
+		$$ = &ast.TableOptimizerHint{HintName: model.NewCIStr($1)}
+	}
+|	hintReadConsistentReplica
```
lzmhhh123: Is this one enforce table read from follower or learner?

lzmhhh123: Does it also need a table list?

parser.y
```diff
@@ -6034,32 +6038,118 @@ TableOptimizerHintList:
...
+	"TRUE"
+	{
+		$$ = true
+	}
+|	"FALSE"
+	{
+		$$ = false
+	}
+
+HintQueryType:
```
lzmhhh123: how about change:
```suggestion
HintSQLType:
```

parser.y
foreyes: This is depend on TiKV. It  will not be used in 4.0GA, so it's just a draft.

parser.y
foreyes: You means change hint name from "query_type" into "sql_type" ?

misc.go
foreyes: Okay, I will contact PM team also.

foreyes: Okay, and I will contact PM team also.

parser.y
foreyes: It's about read from follower, but not decided to enforce or advise.

foreyes:
PTAL. @lzmhhh123 @lamxTyler .
We can ignore problems about hint name or hint arguments now.
Those are decided by PM team, and many won't be used in 4.0GA. Some of them are just draft.

parser.y
lamxTyler: Why not add a syntax for memory unit? Or you cannot raise the right warnings since the quota maybe `0M`.

foreyes: Since syntax start with number and contains a-zA-Z is difficult to handle, we may not be able to use syntax like `233M`. And in my opinion, `M` and `G` should not become a keyword, so...

foreyes: `0M` is a good case

foreyes: now we can use `MEMORY_QUOTA(233 M)`

lzmhhh123:
@foreyes pls resolve the conflicts, then LGTM.

PR #425, opened 2019-08-01
parser: support cast as real

wshwsh12:
Releated PR: https://github.com/pingcap/tidb/pull/11564
Support cast(xxx as real).


parser.y
```diff
@@ -5098,6 +5098,18 @@ CastType:
 		x.Collate = charset.CollationBin
 		$$ = x
 	}
+|	"REAL"
+	{
+		x := types.NewFieldType(mysql.TypeDouble)
+		if parser.lexer.GetSQLMode().HasRealAsFloatMode() {
+			x = types.NewFieldType(mysql.TypeFloat)
+		}
```
kennytm: Let's make this an `if/else` statement.

```go
var x *types.FieldType
if ... {
    x = ... TypeFloat ...
} else {
    x = ... TypeDouble ...
}
```

wshwsh12: Update. PTAL again. @kennytm

zz-jason:
@wshwsh12 please update with the master branch and resolve conflicts.

Issue #408, opened 2019-07-30
Fix TEMPORARY syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
create temporary table t (a int);
create temporary table if not exists t (a int);
create temporary table t1 (like test.t);
```

Claim this subtask based on the comments below this issue.

leiysky:
I'll fix it.

PR #438, opened 2019-08-05
add support for CREATE TEMPORARY TABLE

leiysky:
Add support for CREATE TEMPORARY TABLE #408 

MySQL syntax:
```
create_table_stmt:
          CREATE opt_temporary TABLE_SYM opt_if_not_exists table_ident
          '(' table_element_list ')' opt_create_table_options_etc
        | CREATE opt_temporary TABLE_SYM opt_if_not_exists table_ident
          opt_create_table_options_etc
```

### Check List 

Tests 

 - Unit test


tangenta:
@leiysky TiDB does not support `TEMPORARY` tables. Would you please add a warning in the parser for both `CreateTableStmt` and `DropTableStmt`?

leiysky:
@tangenta PTAL

tangenta:
@leoppro PTAL

parser.y
```diff
@@ -2457,25 +2457,35 @@ DatabaseOptionList:
...
-		stmt.Table = $4.(*ast.TableName)
-		stmt.IfNotExists = $3.(bool)
-		stmt.Options = $6.([]*ast.TableOption)
-		if $7 != nil {
-			stmt.Partition = $7.(*ast.PartitionOptions)
+		stmt := $6.(*ast.CreateTableStmt)
+		stmt.Table = $5.(*ast.TableName)
+		stmt.IfNotExists = $4.(bool)
+		stmt.IsTemporary = $2.(bool)
+		if $2.(bool) {
```
```suggestion
		if stmt.IsTemporary {
```

```suggestion
		if stmt.IsTemporary {
```

parser.y
```diff
@@ -2457,25 +2457,35 @@ DatabaseOptionList:
...
+			yylex.AppendError(yylex.Errorf("TiDB doesn't support CREATE TEMPORARY TABLE, TEMPORARY will be parsed but ignored."))
```
leoppro: could you throw a warn at line 3081?

parser.y
```diff
@@ -2457,25 +2457,35 @@ DatabaseOptionList:
...
+			parser.lastErrorAsWarn()
+		}
+		stmt.Options = $7.([]*ast.TableOption)
+		if $8 != nil {
```
leoppro: why we need to check `$8`?

parser_test.go
```diff
@@ -1766,6 +1766,8 @@ func (s *testParserSuite) TestDDL(c *C) {
 		{"CREATE TABLE foo (a.b, b);", false, ""},
 		{"CREATE TABLE foo (a, b.c);", false, ""},
 		{"CREATE TABLE (name CHAR(50) BINARY)", false, ""},
+		// for create temporary table
+		{"CREATE TEMPORARY TABLE t (a varchar(50), b int);", true, "CREATE TEMPORARY TABLE `t` (`a` VARCHAR(50),`b` INT)"},
```
leoppro: could you add more test cases, cover most of the branches

parser.y
leiysky: It would be `nil interface{}` if there is no `PartitionOpt`, and the type assertion would panic.

parser_test.go
leiysky: added

Issue #428, opened 2019-08-01
Should we need to support `.` in System Variables?


wjhuang2016:
According to https://dev.mysql.com/doc/refman/8.0/en/using-system-variables.html

> System variables implemented by a server component are exposed when the component is installed and have names that begin with a component-specific prefix. For example, the log_filter_dragnet error log filter component implements a system variable named log_error_filter_rules, the full name of which is dragnet.log_error_filter_rules. To refer to this variable, use the full name.

The problem I met is in [#9812](https://github.com/pingcap/tidb/pull/9812) . Presently, TiDB doesn't support `.` in a identifier. After MySQL 8.0, validate_password became a component and its variables uses a `.` instead of `_` to sepearate  component name.
So I wonder should we need to support `.` in system variables or just using `_`?

tiancaiamao:
Could you do this in MySQL8  `set @@SESSION.validate_password.policy = xxx`? @wjhuang2016

wjhuang2016:
> Could you do this in MySQL8 `set @@SESSION.validate_password.policy = xxx`? @wjhuang2016

No, I couldn't.
```
mysql> set @@SESSION.validate_password.policy = low;
ERROR 1229 (HY000): Variable 'validate_password.policy' is a GLOBAL variable and should be set with SET GLOBAL
mysql> set @@GLOBAL.validate_password.policy = low;
Query OK, 0 rows affected (0.00 sec)
```

tiancaiamao:
Just now I try it and doubt that "TiDB doesn't support . in a identifier."

git diff:
```
diff --git a/sessionctx/variable/sysvar.go b/sessionctx/variable/sysvar.go
index f42c04bf3..fbd5aaba8 100644
--- a/sessionctx/variable/sysvar.go
+++ b/sessionctx/variable/sysvar.go
@@ -227,6 +227,7 @@ var defaultSysVars = []*SysVar{
        {ScopeNone, "ssl_key", ""},
        {ScopeNone, "ssl_cipher", ""},
        {ScopeNone, "tls_version", "TLSv1,TLSv1.1,TLSv1.2"},
+       {ScopeGlobal | ScopeSession, "xx.xx", "666"},
        {ScopeNone, "system_time_zone", "CST"},
        {ScopeGlobal, InnodbPrintAllDeadlocks, "0"},
        {ScopeNone, "innodb_autoinc_lock_mode", "1"},
```

and use the modified tidb:

```
mysql> select @@session.xx.xx;
[...]

PR #441, opened 2019-08-05
parser.y: fix variable in set, 'set xx.xx = 42' is a valid statement

tiancaiamao:

'xx.xx' is a valid variable name in the set statement

Fix https://github.com/pingcap/parser/issues/428

Before:

```
mysql> set SESSION xx.xx = 123;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 15 near ".xx = 123" 
```

After:

```
mysql> set global xx.xx=666;
ERROR 1193 (HY000): Unknown system variable 'xx.xx'
```

```
VariableName:
	Identifier
|	Identifier '.' VariableName
	{
		$$ = $1 + "." + $3
	}
```

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -6436,20 +6437,27 @@ SetExpr:
 EqOrAssignmentEq:
     eq | assignmentEq
 
+VariableName:
+	Identifier
```
kennytm: Don't this branch needs `$$ = $1`?

parser.y
```diff
@@ -6436,20 +6437,27 @@ SetExpr:
...
+|	Identifier '.' VariableName
```
wjhuang2016: Should it be `Identifier '.' Identifier`?
```
mysql> set xx.xx.xx=666;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '.xx=666' at line 1
```

parser.y
tiancaiamao: No, the default action for the rule is `$$ = $1`

parser.y

tiancaiamao:
PTAL @kennytm @jackysp

leoppro:
LGTM

PR #446, opened 2019-08-07
add a table hint to access table from a specific type of storage.

lzmhhh123:
At now, tidb need a hint to make table reading from tiflash.

Add a hint to enforce table reading form tiflash.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


ast/misc.go
```diff
@@ -2001,7 +2001,7 @@ func (n *TableOptimizerHint) Restore(ctx *RestoreCtx) error {
 	switch n.HintName.L {
 	case "max_execution_time":
 		ctx.WritePlainf("%d", n.MaxExecutionTime)
-	case "tidb_hj", "tidb_smj", "tidb_inlj":
+	case "tidb_hj", "tidb_smj", "tidb_inlj", "access_from_flash":
```
zz-jason: I think this hint should be a table hint?

lzmhhh123: What do you mean? It exactly a table hint.

misc.go
```diff
@@ -132,6 +132,7 @@ func init() {
 }
 
 var tokenMap = map[string]int{
+	"ACCESS_FROM_FLASH":        accessFromFlash,
```
eurekaka: The name is inconsistent with the name advised by product team?

foreyes:
Product team advised hint `QUERY_TYPE(OLAP/OLTP)`, is this what you want?

ast/misc.go
```diff
@@ -2074,6 +2075,19 @@ func (n *TableOptimizerHint) Restore(ctx *RestoreCtx) error {
 		ctx.WriteKeyWord(n.QueryType.String())
 	case "memory_quota":
 		ctx.WritePlainf("%d M", n.MemoryQuota)
+	case "read_consistent_storage":
+		ctx.WritePlain(n.StoreType.O)
```
foreyes:
```suggestion
		ctx.WritePlain(n.StoreType.String())
```

parser_test.go
```diff
@@ -3849,8 +3866,8 @@ func (s *testParserSuite) TestTablePartitionNameList(c *C) {
 		tableName, ok := source.Source.(*ast.TableName)
 		c.Assert(ok, IsTrue)
 		c.Assert(len(tableName.PartitionNames), Equals, 2)
-		c.Assert(tableName.PartitionNames[0], Equals, model.CIStr{"p0", "p0"})
-		c.Assert(tableName.PartitionNames[1], Equals, model.CIStr{"p1", "p1"})
+		c.Assert(tableName.PartitionNames[0], Equals, model.CIStr{O: "p0", L: "p0"})
```
foreyes: How about `c.Assert(tableName.PartitionNames[0].L, Equals, "p0"})`?

parser_test.go
```diff
@@ -2723,6 +2723,23 @@ func (s *testParserSuite) TestOptimizerHints(c *C) {
 	c.Assert(hints[1].Indexes, HasLen, 1)
 	c.Assert(hints[1].Indexes[0].L, Equals, "c1")
 
+	// Test READ_CONSISTENT_STORAGE
+	stmt, _, err = parser.Parse("select /*+ READ_CONSISTENT_STORAGE(tiflash[t1, t2], tikv[t3]) */ c1, c2 from t1, t2, t1 t3 where t1.c1 = t2.c1 and t2.c1 = t3.c1", "", "")
```
lamxTyler: Why choose this syntax? Why not `READ_CONSISTENT_STORAGE(tiflash, t1, t2), READ_CONSISTENT_STORAGE(tikv,t3)` to be consistent with others？

misc.go
```diff
@@ -89,6 +89,8 @@ func init() {
 	initTokenByte('<', int('<'))
 	initTokenByte('(', int('('))
 	initTokenByte(')', int(')'))
+	initTokenByte('[', int('['))
+	initTokenByte(']', int(']'))
```
lamxTyler: Will it make `[]` reserved?

ast/misc.go
lzmhhh123: If we use `String()`, it may causes `TIFLASH` ---> <code>\`tiflash\`</code>. Then the SQL hint will like <code>READ_CONSISTENT_STORAGE(\`tiflash\`[\`t1\`, \`t2\`])</code>, that's strange.

parser_test.go
lzmhhh123: This line is not a part of this PR. The go-lint in IDE told me it should be changed, so I fixed the lint.

parser_test.go
lzmhhh123: Because we can regard a `tiflash[t1, t2]` as a whole. If the "query block" also exists in this hint, the hint will like `READ_CONSISTENT_STORAGE(@sel1 tiflash t1, t2)`. That's incomprehensible.

ast/misc.go
kennytm: Only `ctx.WriteName()` will add those backquotes. BTW you should use `ctx.WriteKeyWord()` for case-insensitive keyword-like output.

parser_test.go
lzmhhh123: Also seems incomprehensible  in this case: `READ_CONSISTENT_STORAGE(tiflash[t1, t2], tikv[t3])` ----> `READ_CONSISTENT_STORAGE(tiflash t1, t2 tikv t3)`

ast/misc.go
lzmhhh123: OKay. Thanks for your advisement.

foreyes: ??? I'm confused.
``` golang
func (cis CIStr) String() string {
	return cis.O
}
```

parser_test.go
lamxTyler: - I don't have strong opinion on this specific syntax, but if you decided to group similiar things, it is better to change all others that has similiar syntax, like `index`.
- I think it is quite clear what the comment want to say in `READ_CONSISTENT_STORAGE(@sel1 tiflash t1, t2)`.

ast/misc.go
lzmhhh123: @foreyes My fault. I thought you suggested using the `WriteName`.

parser_test.go
lamxTyler: No, just seperate them into two.

lzmhhh123: Hmmm...So may I keep this format? The product manager says that we can only focus on functions. The form is under consideration now.

Issue #412, opened 2019-07-30
Fix EXCHANGE PARTITION syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:
```sql
alter table pt exchange partition p with table nt;
```

Claim this subtask based on the comments below this issue.

leiysky:
I'll fix it

PR #449, opened 2019-08-09
add support for EXCHANGE PARTITION

leiysky:
Add support for EXCHANGE PARTITION. #412 

MySQL syntax:
```
standalone_alter_commands:
    ......
    | EXCHANGE_SYM PARTITION_SYM ident
          WITH TABLE_SYM table_ident opt_with_validation
          {
            $$= NEW_PTN PT_alter_table_exchange_partition($3, $6, $7);
          }
```

### Check List 

Tests 

 - Unit test


leiysky:
`with_validation` is actually a part of [`alter_commands_modifier`](https://github.com/mysql/mysql-server/blob/77306a41670543e103f94fa0db77a71d2b9f88b2/sql/sql_yacc.yy#L8200-L8216). 

And for `EXCHANGE PARTITION`, there is a specific `opt_with_validation` rule.

The current solution is to treat "WITH/WITHOUT VALIDATION" as a part of `AlterTableSpec`, which   is maybe not appropriate.

I think we can add a type `ValidationType` to identify validation type of both `AlterTableStmt` and `AlterTableSpec`(for `EXCHANGE PARTITION`).

parser.y
```diff
@@ -1268,6 +1271,17 @@ AlterTableSpec:
 			PartitionNames: $4.([]model.CIStr),
 		}
 	}
+|	"EXCHANGE" "PARTITION" Identifier "WITH" "TABLE" TableName WithValidationOpt
```
leoppro: i think it's better that putting `PARTITON` rules into `PartitionOpt`

ast/ddl.go
```diff
@@ -1821,6 +1833,7 @@ type AlterTableSpec struct {
 	Partition       *PartitionOptions
 	PartitionNames  []model.CIStr
 	PartDefinitions []*PartitionDefinition
+	ValidationType  ValidationType
```
leoppro:
```suggestion
	WithValidation  bool // only for `ExchangePartition`
```

parser.y
```diff
@@ -1426,6 +1440,25 @@ AlterTableSpec:
 		parser.lastErrorAsWarn()
 	}
 
+WithValidationOpt:
```
leoppro: return `true` or `false`?

leiysky: validation could have default value, in which case restore should omit it

leoppro: fine

parser.y
leiysky: i think it's no necessary to do that. `EXCHANGE PARTITION` is a unique rule so that you cannot reuse it, and the `PARTITION` is not optional.

leiysky:
I'm just worried about the validation rules hard coded in `AlterTableSpec` may cause ambiguity. we  might need to modify them one day we need `alter_commands_modifier`.

parser.y
leoppro: @tangenta PTAL what do you think about that?

leoppro: @kennytm PTAL

tangenta:
@leiysky I think the parser does already support `alter_commands_modifier`.
```
alter_commands_modifier: alter_algorithm_option
| alter_lock_option
| with_validation
```
All of its branches are implemented.

parser.y
```diff
@@ -957,6 +958,8 @@ import (
 	WhenClauseList		"When clause list"
 	WithReadLockOpt		"With Read Lock opt"
 	WithGrantOptionOpt	"With Grant Option opt"
+	WithValidation		"with validation"
+	WithValidationOpt		"optional with validation"
```
tangenta: The indention here is weird. Would you please fix it?

leiysky: fine, but is there any format tool for parser.y?

leoppro: unfortunatly, not

ast/ddl.go
```diff
@@ -1807,6 +1808,17 @@ func (a AlterAlgorithm) String() string {
 	}
 }
 
+// ValidationType represents validation types of alter command
+type ValidationType byte
```
kennytm: https://dev.mysql.com/doc/refman/8.0/en/partitioning-management-exchange.html

> Optionally, you can append `WITH VALIDATION` or `WITHOUT VALIDATION`. … `WITH VALIDATION` is the default. 

I think a boolean indicating whether `WITHOUT VALIDATION` is specified is enough. Specifying `WITH VALIDATION` should be equivalent to not specifying it.

leiysky: that's ok. 

we would treat default as `WITH VALIDATION` in `AlterTableExchangePartition` spec. and when restoring, we just write `WITHOUT VALIDATION`, by doing this we would make no break change.

leiysky:
@kennytm @leoppro PTAL again.

leoppro:
LGTM

Issue #417, opened 2019-07-30
Fix CREATE DATABASE … ENCRYPTION syntax


kennytm:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

(Ref: https://dev.mysql.com/doc/refman/8.0/en/create-database.html)

```sql
CREATE DATABASE t ENCRYPTION = 'N';
CREATE DATABASE t ENCRYPTION 'N';
CREATE DATABASE t DEFAULT ENCRYPTION = 'N';
CREATE DATABASE t DEFAULT ENCRYPTION 'N';

-- the following should warn
CREATE DATABASE t ENCRYPTION = 'Y';
CREATE DATABASE t ENCRYPTION 'Y';
CREATE DATABASE t DEFAULT ENCRYPTION = 'Y';
CREATE DATABASE t DEFAULT ENCRYPTION 'Y';
```

Claim this subtask based on the comments below this issue.

zhenghaoz:
Let me fix it

PR #450, opened 2019-08-09
Fix CREATE DATABASE … ENCRYPTION syntax

zhenghaoz:
#### Fix CREATE DATABASE … ENCRYPTION syntax

Issue: pingcap/parser#417

#### MySQL Syntax:

```
create_specification:
...
  | DEFAULT ENCRYPTION [=] {'Y' | 'N'}
```

### Bad SQL Case:

```sql
CREATE DATABASE t ENCRYPTION = 'N';
CREATE DATABASE t ENCRYPTION 'N';
CREATE DATABASE t DEFAULT ENCRYPTION = 'N';
CREATE DATABASE t DEFAULT ENCRYPTION 'N';

CREATE DATABASE t ENCRYPTION = 'Y';
CREATE DATABASE t ENCRYPTION 'Y';
CREATE DATABASE t DEFAULT ENCRYPTION = 'Y';
CREATE DATABASE t DEFAULT ENCRYPTION 'Y';
```

### Check List

Tests
- Unit test


zhenghaoz:
It seems that an error should be raised when the value for `ENCRYPTION` is neither `Y` nor `N`.
```sql
CREATE DATABASE t ENCRYPTION = 'xxx';
```
In MySQL 8.0, the raised error is:
```
ERROR 1525 (HY000): Incorrect argument (should be Y or N) value: 'xxxxx'
```
The error handling seems necessary but I don't know where we could declare an error type so that we can check the value in `parser.y`.

kennytm:
Let's ignore the error when the value is neither N or Y. This could be checked after parsing if needed.

parser.y
```diff
@@ -2429,6 +2434,10 @@ DatabaseOption:
 	{
 		$$ = &ast.DatabaseOption{Tp: ast.DatabaseOptionCollate, Value: $4.(string)}
 	}
+|	DefaultKwdOpt "ENCRYPTION" EqOpt EncryptionOption
```
kennytm: Let's just use `stringLit` here (and remove `EncryptionOption`). Like `TableOption` we normally don't create a separate rule for every option unless it is necessary.

```suggestion
|	DefaultKwdOpt "ENCRYPTION" EqOpt stringLit
```

leoppro:
@zhenghaoz resolve the conflicts please

zhenghaoz:
@leoppro done

PR #337, opened 2019-05-23
*: add `IF EXISTS` and `IF NOT EXISTS` supported for MariaDB

csuzhangxc:
### What problem does this PR solve? 

for MaraiDB 10.0.2+, `ALTER TABLE` support `IF EXISTS` and `IF NOT EXISTS` clauses for some statements, and `CREATE INDEX` also support `IF NOT EXISTS`.

see <https://mariadb.com/kb/en/library/alter-table/> and <https://mariadb.com/kb/en/library/create-index/>

### What is changed and how it works?

1. add `IF EXISTS` for
	1. `DROP COLUMN`
	2. `CHANGE COLUMN`
	3. `MODIFY COLUMN`
	4. `DROP INDEX`
	5. `DROP FOREIGN KEY`
	6. `DROP PARTITION`
2. add `IF NOT EXISTS` for
	1. `ADD COLUMN`
	2. `ADD INDEX`
	3. `ADD FOREIGN KEY`
	4. `CREATE INDEX`
	5. `ADD PARTITION`

### Check List 

Tests 

 - Unit test

Corresponding TiDB PR: https://github.com/pingcap/tidb/pull/10723

PR #455, opened 2019-08-10
parser: fix `IF NOT EXISTS` support for MariaDB `ADD PARTITION` syntax

spongedu:

#337 support `IF NOT EXISTS` option for`ADD PARTITION` syntax in MaraiDB 10.0.2+, but not complete.  the same support for `ADD PARTITION PARTITIONS NUM` syntax is missing.
For example, we have a table defined as follows:
```sql
 CREATE TABLE `t2` (
  `name` varchar(30) DEFAULT NULL,
  `started` date DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
PARTITION BY HASH( year(`started`) )
PARTITIONS 6 
```
in MaraiDB 10.0.2+, the following statement is ok :
```sql
MariaDB [test]> alter table t2 add partition if not exists partitions 2;
Query OK, 0 rows affected (0.11 sec)
```
while for now, this sql failed in parser:
```sql
tidb> alter table t2 add partition if not exists partitions 2;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use li
ne 1 column 53 near "partitions 2" 

```

Add `IfNotExists` option for the corresponding branch for `ADD PARTITION` syntax in parser.y. 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N/A

Side effects

N/A


spongedu:
@csuzhangxc PTAL

PR #458, opened 2019-08-12
remove executable on parser_test.go

TennyZhuang:

It seems that the parser_test.go should not be executable.

chmod -x parser_test.go

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


leoppro:
LGTM

leoppro:
there are some files in `ast` dir that should not be executable.
could you fix them by the way?

TennyZhuang:
Ok, I will trim them later today.

kennytm <notifications@github.com>于2019年8月14日 周三06:27写道：

Issue #467, opened 2019-08-13
Fix SERIAL syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
create table t (a serial)
create table t (b int, a serial)
create table t (a serial, b decimal)
create table t (a int serial default value)
create table t (a bigint serial default value)
create table t (a smallint serial default value, b int serial default value)
```

Claim this subtask based on the comments below this issue.

spongedu:
Let me fix it :)

PR #481, opened 2019-08-14
Support keyword `SERIAL`

spongedu:

for #467 , support keyword `SERIAL`.

1. Support `SERIAL` as a `Numeric Type` in `ColumnDef`
2. Support `SERIAL DEFAULT VALUE` as an `ColumnOption`
3. Add related tests

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N / A

Side effects

N / A


spongedu:
it's interesting that the `Verify parser.go is up-to-date` stage in tests is not consistent with the `fmt` stage. The tests will fail anyway...

spongedu:
I think we should add a `fmt` command in `Verify parser.go is up-to-date` stage in  the `ci` test base

spongedu:
@leoppro  PTAL :)

parser_test.go
```diff
@@ -2426,6 +2426,15 @@ func (s *testParserSuite) TestDDL(c *C) {
 		{"create table nchar (a int, b nchar);", true, "CREATE TABLE `nchar` (`a` INT,`b` CHAR)"},
 		{"create table nchar (a int, b nchar(50));", true, "CREATE TABLE `nchar` (`a` INT,`b` CHAR(50))"},
 
+		// Test keyword `SERIAL`
+		{"create table t (a serial);", true, "CREATE TABLE `t` (`a` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE KEY)"},
+		{"create table t (a serial null);", true, "CREATE TABLE `t` (`a` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE KEY NULL)"},
```
leoppro:
```suggestion
		{"create table t (a serial null);", true, "CREATE TABLE `t` (`a` BIGINT UNSIGNED AUTO_INCREMENT UNIQUE KEY)"},
```
this should be right

parser_test.go
```diff
@@ -2426,6 +2426,15 @@ func (s *testParserSuite) TestDDL(c *C) {
...
+		{"create table t (b int, a serial);", true, "CREATE TABLE `t` (`b` INT,`a` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE KEY)"},
+		{"create table t (a int serial default value);", true, "CREATE TABLE `t` (`a` INT NOT NULL AUTO_INCREMENT UNIQUE KEY)"},
+		{"create table t (a int serial default value null);", true, "CREATE TABLE `t` (`a` INT NOT NULL AUTO_INCREMENT UNIQUE KEY NULL)"},
```
leoppro:
```suggestion
		{"create table t (a int serial default value null);", true, "CREATE TABLE `t` (`a` INT AUTO_INCREMENT UNIQUE KEY)"},
```

parser_test.go
spongedu: The `Restore` method only require the restored sql is semantic equivalence to the origin, so  
I just make sure the semantic equivalence here,  leave the duplicate removal job behind.

The current codebase fail on this kind of sql as well (`create table t (a serial not null null not null null not null);` for example), Shall I fix this in this pr?

parser_test.go
spongedu: ditto

parser_test.go
tangenta: I think it's TiDB's responsibility to remove duplication. @leoppro 
If we do this completely in the parser, tons of code in TiDB needs to be move.

leoppro: @kennytm what do you think about it

kennytm: Let's leave it as-is.

leoppro:
LGTM please resolve conflicts
@spongedu

Issue #476, opened 2019-08-14
Fix COLUMN_FORMAT syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
create table t (a int column_format fixed)
create table t (a int column_format default)
create table t (a int column_format dynamic)
alter table t modify a bigint column_format default
```

Claim this subtask based on the comments below this issue.

doggeral:
let me fix it.

PR #482, opened 2019-08-14
Support COLUMN_FORMAT syntax.

doggeral:
To support the COLUMN_FORMAT syntex. 
https://github.com/pingcap/parser/issues/476

Add the COLUMN_FORMAT support.

### Check List 

 - Unit test


parser.y
```diff
@@ -2100,6 +2101,20 @@ ColumnOption:
...
+|
+    "COLUMN_FORMAT" "FIXED"
+	{
+		$$ = &ast.ColumnOption{Tp: ast.ColumnOptionColumnFormat, StrValue: "FIXED"}
+	}
+|
+    "COLUMN_FORMAT" "DYNAMIC"
+	{
+		$$ = &ast.ColumnOption{Tp: ast.ColumnOptionColumnFormat, StrValue: "DYNAMIC"}
+	}
```
tangenta: It would be better to abstract `DEFAULT`, `FIXED`, `DYNAMIC` as a new production named `columnFormat`.

parser.y
```diff
@@ -299,6 +299,7 @@ import (
 	client		"CLIENT"
 	coalesce	"COALESCE"
 	collation	"COLLATION"
+	columnFormat "COLUMN_FORMAT"
```
tangenta: Please indent the line with tab.

doggeral: Will do.

parser.y
doggeral: What's the benefit to create new columnFormat name? Can it be reused by anywhere else?

tangenta: No, it can't. Just in line with the existing style.

doggeral: got you, will do.

doggeral:
Resolved the conflict.

Issue #472, opened 2019-08-14
Fix STATS_AUTO_RECALC syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
create table t (a int) stats_auto_recalc = 0
create table t (a int) stats_persistent = 1, stats_auto_recalc = 1
create table t (a int) stats_auto_recalc = 1, stats_sample_pages = 25
alter table t modify a bigint, ENGINE=InnoDB, stats_auto_recalc = 0
```

documentation: https://dev.mysql.com/doc/refman/8.0/en/innodb-persistent-stats.html
Claim this subtask based on the comments below this issue.

lauhg:
Let me fix it.

lauhg:
```bash
mysql> select version();
+-----------+
| version() |
+-----------+
| 8.0.14    |
+-----------+
1 row in set (0.00 sec)

mysql> create table t (a int) stats_auto_recalc = 1, algorithm default;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'algorithm default' at line 1
mysql>
```

`create table t (a int) stats_auto_recalc = 1, algorithm default` seems not a valid sql case.

tangenta:
@lauhg Oh, it a mistake. I will update the issue.

PR #483, opened 2019-08-15
parser: fix STATS_AUTO_RECALC syntax

lauhg:
#### Fix compatibility problem about `STATS_AUTO_RECALC` syntax

Issue: pingcap/parser#472

MySQL Syntax:
```sql
type:
...
  | STATS_AUTO_RECALC_SYM opt_equal ternary_option
```

Bad SQL Case:
```sql
create table t (a int) stats_auto_recalc = 0
create table t (a int) stats_persistent = 1, stats_auto_recalc = 1
create table t (a int) stats_auto_recalc = 1, stats_sample_pages = 25
alter table t modify a bigint, ENGINE=InnoDB, stats_auto_recalc = 0
```

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -7913,9 +7914,21 @@ TableOption:
 	{
 		$$ = &ast.TableOption{Tp: ast.TableOptionStatsPersistent}
 	}
+|	"STATS_AUTO_RECALC" EqOpt LengthNum
+	{
+		$$ = &ast.TableOption{Tp: ast.TableOptionStatsAutoRecalc, UintValue: $3.(uint64)}
```
leoppro: STATS_AUTO_RECALC = [1|0|Default]
we should check if `LengthNum` is invalid

parser.y
```diff
@@ -7913,9 +7914,21 @@ TableOption:
...
+		yylex.AppendError(yylex.Errorf("The STATS_AUTO_RECALC is parsed but ignored by all storage engines."))
+		parser.lastErrorAsWarn()
+	}
+|	"STATS_AUTO_RECALC" EqOpt "DEFAULT"
+	{
+		$$ = &ast.TableOption{Tp: ast.TableOptionStatsAutoRecalc, StrValue: "DEFAULT"}
```
```suggestion
		$$ = &ast.TableOption{Tp: ast.TableOptionStatsAutoRecalc, UintValue: 0}
```

lauhg:
@leoppro PTAL

ast/ddl.go
```diff
@@ -1678,6 +1679,14 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
 		ctx.WritePlain("= ")
 		ctx.WriteKeyWord("DEFAULT")
 		ctx.WritePlain(" /* TableOptionStatsPersistent is not supported */ ")
+	case TableOptionStatsAutoRecalc:
+		ctx.WriteKeyWord("STATS_AUTO_RECALC ")
+		ctx.WritePlain("= ")
+		if n.UintValue == 0 {
+			ctx.WriteKeyWord("DEFAULT")
```
leoppro: i think it's not necessary to hold "DEFAULT"

lauhg: Do you mean just show `0` or `1`?

leoppro: yes

lauhg:
@leoppro PTAL

leoppro:
@tangenta PTAL

parser.y
```diff
@@ -7994,9 +7995,26 @@ TableOption:
...
+			yylex.AppendError(yylex.Errorf("The value of STATS_AUTO_RECALC must be one of [0|1|DEFAULT]."))
+			return 1
+		}
+		$$ = &ast.TableOption{Tp: ast.TableOptionStatsAutoRecalc, UintValue: n}
+		yylex.AppendError(yylex.Errorf("The STATS_AUTO_RECALC is parsed but ignored by all storage engines."))
+		parser.lastErrorAsWarn()
+	}
+|	"STATS_AUTO_RECALC" EqOpt "DEFAULT"
+	{
+		$$ = &ast.TableOption{Tp: ast.TableOptionStatsAutoRecalc, UintValue: 0}
```
tangenta: @leoppro @lauhg I don't think `DEFAULT` is the same as 0. 

Here is the MySQL 8.0 [documentation](https://dev.mysql.com/doc/refman/8.0/en/create-table.html):
> - STATS_AUTO_RECALC
Specifies whether to automatically recalculate persistent statistics for an InnoDB table. The value DEFAULT causes the persistent statistics setting for the table to be determined by the innodb_stats_auto_recalc configuration option. The value 1 causes statistics to be recalculated when 10% of the data in the table has changed. The value 0 prevents automatic recalculation for this table;

TiDB needs the information to decide whether using `innodb_stats_auto_recalc`(or something similar) or not.

ast/ddl.go
```diff
@@ -1678,6 +1679,10 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
 		ctx.WritePlain("= ")
 		ctx.WriteKeyWord("DEFAULT")
 		ctx.WritePlain(" /* TableOptionStatsPersistent is not supported */ ")
+	case TableOptionStatsAutoRecalc:
+		ctx.WriteKeyWord("STATS_AUTO_RECALC ")
+		ctx.WritePlain("= ")
+		ctx.WritePlainf("%d", n.UintValue)
```
tangenta: Would you please merge these two lines?

parser.y
lauhg: @tangenta I agree with you.
@leoppro What do you think?

ast/ddl.go
lauhg: ok.

parser.y
leoppro: fine, I think we should add a feild named `default` to TableOption

leoppro: could fix `"STATS_SAMPLE_PAGES" EqOpt "DEFAULT"` by the way?

kennytm: Could this be parsed into a `nil` option?

lauhg: @kennytm what do you mean?

lauhg: > fine, I think we should add a feild named `default` to TableOption

I think this is a good way to solve this problem.

lauhg:
I add a feild named `Default` to TableOption.
@leoppro @tangenta PTAL

tangenta:
Adding a field this way could lead to every TableOption increases 4 bytes memory usage.
Here are two options:
1. Add a `TableOption` type named `TableOptionStatsAutoRecalcDefault`
2. Assign a special value to `UIntValue`(i.e. `math.MaxUint64`) in `TableOption`, which represents `default`.

How do you think? @leoppro @lauhg

lauhg:
Or just assign `DEFAULT` to StrValue?

tangenta:
@lauhg It's unnecessary to compare a whole string when checking the `default` property.

lauhg:
@tangenta  ok, I'd like to choose option 2.
@leoppro  How do you think?

parser.y
kennytm: @lauhg Since `STATS_SAMPLE_PAGES = DEFAULT` is the default the option could just be entirely removed from the AST, so changing this branch to `$$ = nil` should be possible.

lauhg: @kennytm As far as I know, this parser is not only part of TiDB, it will also be used by other projects. In order to be more compatible with mysql, I think it is necessary to know if the default value is explicitly set.

kennytm: @lauhg What project would benefit from knowing that the DEFAULT is explicitly set? In MySQL (8.0.17) even if you explicit write `STATS_AUTO_RECALC = DEFAULT` it will disappear in `SHOW CREATE TABLE`, showing these are semantically equivalent.

```mysql
mysql> create table t (a int) stats_auto_recalc = 0;
Query OK, 0 rows affected (0.06 sec)

mysql> show create table t;
+-------+--------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                         |
+-------+--------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci STATS_AUTO_RECALC=0 |
+-------+--------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> drop table t;
Query OK, 0 rows affected (0.02 sec)
```

```mysql
mysql> create table t (a int) stats_auto_recalc = default;

mysql> show create table t;
+-------+------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                     |
+-------+------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci |
+-------+------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> drop table t;
```

lauhg: @kennytm 
In `create table` case, these are semantically equivalent.
But in `alter table` case, these are not equivalent.
For example:
```sql
create table t (a int) stats_auto_recalc = 1;
alter table t stats_auto_recalc = default;
```
If parser parses `stats_auto_recalc = default` to `nil`, it would be equivalent to
`alter table t`.
Apparently, it changes semantics.

leoppro:
> Adding a field this way could lead to every TableOption increases 4 bytes memory usage.
> Here are two options:
> 
> 1. Add a `TableOption` type named `TableOptionStatsAutoRecalcDefault`
> 2. Assign a special value to `UIntValue`(i.e. `math.MaxUint64`) in `TableOption`, which represents `default`.
> 
> How do you think? @leoppro @lauhg

I don't think we should reduce the readability of code for save 4 bytes(or 4*n bytes)

lauhg:
What's the conclusion?

tangenta:
Fine, keep the current changes is ok. Please resolve the conflicts.

lauhg:
Resolved.
PTAL @tangenta

PR #486, opened 2019-08-15
Support foreign key choose parent schema as default in CreateTable stmt 

3pointer:
Aligned with MySQL behavior
```
drop database if exists fk_test;
create database if not exists fk_test;
create table fk_test.t1 (c1 int not null primary key);
create table fk_test.t2 (id int not null primary key, c1 int not null, constraint foreign key (c1) references t1(c1))
```
MySQL can deal with above sql well, but TiDB will get `ERROR 1046 (3D000) at line 4: No database selected`

if foreign key has no schema, use parent table schema

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes


zz-jason:
@3pointer Is this PR ready to review?

Issue #474, opened 2019-08-14
Fix FLOAT4/FLOAT8 syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
create table t (a float4)
create table t (a float4(4, 2))
create table t (a float4, b float)
alter table t modify a float4
create table t (a float, b float4 as (a + 1))

create table t (a float8)
create table t (a float8(4, 2))
create table t (a float8, b float)
alter table t modify a float8
create table t (a float, b float8 as (a + 1))
```

documentation: https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html
Claim this subtask based on the comments below this issue.

PR #490, opened 2019-08-15
parser: support float4 and float8

spongedu:

1. For #474 , support type `FLOAT4` and `FLOAT8`
2. Bug fix, support `CAST AS DOULBE PRECISION` syntax.

1. modify `misc.go`, extend `tokenMap` and `aliasMap`
2. Fix gramma in `parser.y`
3. Add related tests

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N / A

Side effects

N  / A

Related changes


leoppro:
@tangenta PTAL

parser_test.go
```diff
@@ -1784,8 +1788,6 @@ func (s *testParserSuite) TestIdentifier(c *C) {
 func (s *testParserSuite) TestDDL(c *C) {
 	table := []testCase{
 		{"CREATE", false, ""},
-		{"CREATE TABLE", false, ""},
-		{"CREATE TABLE foo (", false, ""},
```
tangenta: Why do we remove these lines?

spongedu: > Why do we remove these lines?

By mistake. I'll revert this change

parser.y
```diff
@@ -112,7 +112,7 @@ import (
 	distinct		"DISTINCT"
 	distinctRow		"DISTINCTROW"
 	div 			"DIV"
-	doubleType		"DOUBLE"
+	doubleType		"DOUBLE / FLOAT8"
```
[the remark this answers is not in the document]
spongedu: > According to [documentation](https://www.gnu.org/software/bison/manual/html_node/Token-Decl.html#Token-Decl) in terms of `%token`,
> 
> > Once you equate the literal string and the token name, you can use them interchangeably in further declarations or the grammar rules. The yylex function can use the token name or the literal string to obtain the token type code number (see Calling Convention). Syntax error messages passed to yyerror from the parser will reference the literal string instead of the token name.
> 
> it's better not to make this change.

MySQL alias `FLOAT4` to `FLOAT`, `FLOAT8` to `DOUBLE` in its symbol map. See: [mysql-server/sql/lex.h](https://github.com/mysql/mysql-server/blob/817fcb883a8457fcf507156bb1a2052e887acc4c/sql/lex.h#L243).
So, there's no independent symbol for `FLOAT4` and `FLOAT8`, they are done in the lexer. I think it's better to follow `MySQL`'s behavior here. What do you think? @tangenta

tangenta:
Because of the possibility of breaking the forward compatibility, this PR will not be merged.
We are so sorry :( @spongedu

spongedu:
> Because of the possibility of breaking the forward compatibility, this PR will not be merged.
> We are so sorry :( @spongedu

What `possibility of breaking the forward compatibility` mean ?

tangenta:
@spongedu 
Suppose a user created a table with the name `float4` or `float8`. After this PR merged, selecting this table will encounter syntax error.

spongedu:
> @spongedu
> Suppose a user created a table with the name `float4` or `float8`. After this PR merged, selecting this table will encounter syntax error.
> 
> Correction: it is **backward** compatibility.

Opps.... That's a problem indeed....  so we have to hold all this changes on `ReservedKeyword`  and left them to a **`backward compatibility  session`** ?

tangenta:
After careful consideration and discussion, we decide not to support `float4/float8`, so the issue will be closed.

spongedu:
> After careful consideration and discussion, we decide not to support `float4/float8`, so the issue will be closed.

OK. So I'll close this pr.

tangenta:
We will collect similar issues. Thank you for your contribution. @spongedu

PR #491, opened 2019-08-16
 parser: support query block in hint

lamxTyler:

Support query block syntax in hint.

Add optional query block as arugments for each hint.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - None

Related changes

 - None


parser.y
```diff
@@ -6324,84 +6327,105 @@ TableOptimizerHintList:
 	}
 
 TableOptimizerHintOpt:
-	index '(' HintTableAndIndexList ')'
+	index '(' QueryBlockOpt HintTable IndexNameList ')'
```
foreyes: Do we have other way to do this? `INDEX(@sel_1 t idx1, idx2, idx3)` looks weird. Because `IndexNameList` here can not be empty, we can split by comma.

ast/misc.go
```diff
@@ -2035,16 +2037,40 @@ type TableOptimizerHint struct {
 	HintFlag         bool
 }
 
+// HintTable is table in the hint. It may have query block info.
+type HintTable struct {
+	TableName model.CIStr
+	QBName    model.CIStr
```
lzmhhh123: Why need we save the `QBName` in both `HintTable` and `TableOptimizerHint`?

parser.y
```diff
@@ -6324,84 +6327,105 @@ TableOptimizerHintList:
...
 
-HintTableAndIndexList:
-	Identifier
+QueryBlockOpt:
 	{
-		$$ = []model.CIStr{model.NewCIStr($1)}
+		$$ = model.NewCIStr("")
 	}
-|	HintTableAndIndexList ',' Identifier
+|	singleAtIdentifier
```
foreyes: I'm wondering whether we can use comma to split arguments in SQL hints too, it looks better.
```suggestion
|	singleAtIdentifier
	{
		$$ = model.NewCIStr($1)
	}
|	singleAtIdentifier ‘,’
```

ast/misc.go
foreyes: Because we can specify the QB within which the hint will take effect, and also specify a table in some QB?

foreyes: For example: `/*+ HASH_JOIN(@sel1 t1, t2) */` and `/*+ HASH_JOIN(t1@sel1, t2@sel2)`.

lamxTyler: Yes, you can take a look at the test.

parser.y
lamxTyler: No, even if you want `,`, it should not be put here.

parser.y
lamxTyler: It can be empty now.

parser.y
```diff
@@ -574,6 +574,7 @@ import (
 	hintEnablePlanCache	"ENABLE_PLAN_CACHE"
 	hintUsePlanCache	"USE_PLAN_CACHE"
 	hintReadConsistentReplica	"READ_CONSISTENT_REPLICA"
+	hintQBName	"QB_NAME"
```
lzmhhh123: I'm not against using `QBName` in `ast.go`. But could we use the full name in `parser.y` to make its mean more clear?

lamxTyler: But full name would make using it harder beacause `queryblock` has 10 chars, it is reasonable trade off I think.

lzmhhh123: I mean change `hintQBName "QB_NAME"` to `hintQBName "QUERY_BLOCK_NAME"`.

lamxTyler: It is also what I means, It contains 8 + 1 more chars that `QB_NAME`.

lzmhhh123: All right...

lamxTyler:
@foreyes We can change it latter needed. Please approve it.

PR #513, opened 2019-08-21
change memory quota hint argument

foreyes:

Argument should be `x MB` or `x GB`.
Executor need memory quota as `int64`.
Makes it possible to use MEMORY_QUOTA(0 MB)

Change `TableOptimizerHint` struct and parser

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -6621,15 +6621,14 @@ HintQueryType:
...
 	NUM Identifier
 	{
-		// May change into MB/MiB or GB/GiB
 		switch model.NewCIStr($2).L {
-		case "m":
-			$$ = getUint64FromNUM($1)
-		case "g":
-			$$ = getUint64FromNUM($1) * 1024
+		case "mb":
+			$$ = $1.(int64) * 1024 * 1024
```
lzmhhh123: Shall we regard `MB` as the base unit? So that we can avoid multiplying `1024*1024`.

foreyes: Executor handle Byte as the base unit, I think we should sync here.

lzmhhh123: All right.

Issue #478, opened 2019-08-14
Fix CREATE INDEX .. VISIBLE/INVISIBLE syntax


leoppro:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
CREATE INDEX ident ON t_n ( ident ) VISIBLE 
CREATE INDEX ident ON t_n ( ident ) INVISIBLE 
```

Claim this subtask based on the comments below this issue.

zhenghaoz:
Let me fix it ^_^

PR #518, opened 2019-08-23
Fix CREATE INDEX .. VISIBLE/INVISIBLE syntax

zhenghaoz:
#### Fix CREATE INDEX .. VISIBLE/INVISIBLE syntax

Issue: pingcap/parser#478

#### MySQL Syntax:

```
CREATE [UNIQUE | FULLTEXT | SPATIAL] INDEX index_name
...
index_option:
....
   | {VISIBLE | INVISIBLE}
```

### Bad SQL Case:

```sql
CREATE INDEX idx ON t ( a ) VISIBLE
CREATE INDEX idx ON t ( a ) INVISIBLE
CREATE INDEX idx ON t ( a ) USING HASH VISIBLE
CREATE INDEX idx ON t ( a ) USING HASH INVISIBLE
CREATE INDEX idx ON t ( a ) INVISIBLE VISIBLE
CREATE INDEX idx ON t ( a ) VISIBLE INVISIBLE
```

### Check List

Tests
- Unit test

PS:

Since the default option is visible, `VISIBLE` won't be restored.

> - VISIBLE, INVISIBLE
> 
> Specify index visibility. Indexes are visible by default. An invisible index is not used by the optimizer. Specification of index visibility applies to indexes other than primary keys (either explicit or implicit). For more information, see Section 8.3.12, “Invisible Indexes”.


parser.y
```diff
@@ -4136,7 +4140,9 @@ IndexOptionList:
 			} else if opt2.Tp != 0 {
 				opt1.Tp = opt2.Tp
 			} else if opt2.KeyBlockSize > 0 {
-			    opt1.KeyBlockSize = opt2.KeyBlockSize
+			    	opt1.KeyBlockSize = opt2.KeyBlockSize
+			} else if opt2.Invisible {
+				opt1.Invisible = opt2.Invisible
```
kennytm: 🤔 

In MySQL the following should create a _visible_ index (the last option takes effect):

```sql
create table a (a int, key(a) invisible visible);
```

leoppro:
resolve conflicts please

PR #189, opened 2019-01-23
implement Restore for DeleteStmt

arnkore:
1. implement Restore for DeleteStmt;
2. fix [issue#188](https://github.com/pingcap/parser/issues/188) support parse tbl_name.* in delete multiple-table syntax;
pingcap/tidb#8532

Issue #506, opened 2019-08-19
Keep track of compatibility issues for various of MySQL version


spongedu:
Recently I'm working on improving TiDB parser's compatibility( see [Improve parser compatibility](https://github.com/pingcap/tidb/issues/11486)). As we all know, parser's behavior is mostly base on MySQL5.7,  but as time goes by there are some MariaDB (https://github.com/pingcap/parser/pull/337 , e.g) and MySQL8.0 (https://github.com/pingcap/parser/pull/496 e.g) feature join in. Maybe we could add some control here so that the parser could stick to one version at a time.

I think it's a good start to keep track of all the issues that is for `MySQL 8.0`, `MariaDB` or anything not `MySQL 5.7`. Maybe we should add tags for these issue, as well.

The `MySQL 5.8` compatibility issues:

- #444, solved in #496 
- #497, solved in #505 

The `MariaDB` compatibility issues:

- #337 and improved in #455

winkyao:
Agree, Could you write a proposal for this?

PR #521, opened 2019-08-25
parser: fix  tableAlias in `DeleteStmt`

### LINKED ISSUE #357: `TableName` rule should not derive to `Identifier '.' '*'`
In fix of #188, the following new rule was introduced:

https://github.com/pingcap/parser/blob/690470b074e1ed6d7ce074ff0ea156787ecee523/parser.y#L4886-L4898

but this affects all uses of TableName and leads to some non-sense syntax being accepted, e.g.

```sql
ANALYZE TABLE a.*;
-- should throw syntax error, but accepted and treated as equivalent to ANALYZE TABLE a;
```

We should create a separate rule for DELETE statement instead, and revert the TableName change.

spongedu:

 #189 tried to implement `Restore` for `DeleteStmt`, but there's a little mistake: the `.*` syntax should not be defined as part of `TableName`. It should be defined as an independent term and be used in `DeleteStmt` only, as is defined in [MySQL Parser](https://github.com/mysql/mysql-server/blob/817fcb883a8457fcf507156bb1a2052e887acc4c/sql/sql_yacc.yy#L13078)

This pr seems fix #357 as well.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N / A

Side effects

N / A

Related changes


kennytm:
```sql
CREATE table .t (a bigint)
```

Since this is invalid in MySQL 8.0 I don't think we want to support this. WDYT @leoppro @tangenta

spongedu:
> ```sql
> CREATE table .t (a bigint)
> ```
> 
> Since this is invalid in MySQL 8.0 I don't think we want to support this. WDYT @leoppro @tangenta

But currently `TiDB` and `the Parser` is mainly MySQL 5.7 compatible, and it seems long way for `TiDB` and the users to migrate to MySQL 8.0's behavior. I'm not sure if we should ignore the MySQL 5.7 compatible issues now....
 BTW, What's the compatibility strategy do we hold for `the Parser` now?  As I've mentioned in #506 , currently `the Parser` seems a mixture of MySQL 5.7, MySQL 8.0 and MariaDB.

kennytm:
@spongedu We intend to be eventually compatible with 8.0 (pingcap/tidb#7968), so if a missing 5.7 feature is deprecated or removed on 8.0 we won't support it (unless it is trivial). Example of such rejected feature requests: pingcap/tidb#3685, pingcap/tidb#4025, pingcap/tidb#10092.

spongedu:
> @spongedu We intend to be eventually compatible with 8.0 ([pingcap/tidb#7968](https://github.com/pingcap/tidb/issues/7968)), so if a missing 5.7 feature is deprecated or removed on 8.0 we won't support it (unless it is trivial). Example of such rejected feature requests: [pingcap/tidb#3685](https://github.com/pingcap/tidb/issues/3685), [pingcap/tidb#4025](https://github.com/pingcap/tidb/issues/4025), [pingcap/tidb#10092](https://github.com/pingcap/tidb/issues/10092).

OK.

spongedu:
@kennytm @leoppro @tangenta  PTAL

parser.y
```diff
@@ -5789,6 +5788,37 @@ TableNameList:
 		$$ = append($1.([]*ast.TableName), $3.(*ast.TableName))
 	}
 
+TableNameOptWild:
```
kennytm: Could this be replaced as `TableName OptWild`? An extra rule doesn't seem necessary.

parser.y
```diff
@@ -5789,6 +5788,37 @@ TableNameList:
...
+
+OptWild:
+	%prec empty
+	{
+		$$ = nil
+	}
+|	'.' '*'
+	{
+		$$ = nil
+	}
```
kennytm: No need to write `$$ = nil` for these two branches.

spongedu: ok

parser.y
spongedu: I think we need this to avoid shift/reduce conflicts.

spongedu: > Could this be replaced as `TableName OptWild`? An extra rule doesn't seem necessary.

I think we need this to avoid shift/reduce conflicts.

kennytm: Okay.

spongedu:
@crazycs520  PTAL :)

SunRunAway:
@spongedu Please help resolve the confliting files.

spongedu:
@SunRunAway  PTAL :)

PR #522, opened 2019-08-26
ast: change the order of visiting select stmt

lamxTyler:

Adjust the visit order in select stmt so we can tranverse from left to right in the origin text.

For https://github.com/pingcap/tidb/pull/11861

Adjust the vist order and add a offset variable in select stmt.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

 - None


ast/dml.go
```diff
@@ -764,6 +764,8 @@ type SelectStmt struct {
 	IsAfterUnionDistinct bool
 	// IsInBraces indicates whether it's a stmt in brace.
 	IsInBraces bool
+	// QueryBlockOffset indicates the order of this SelectStmt if counted from left to right in the sql text.
+	QueryBlockOffset int
```
eurekaka: Where would this field be set?

winoros: It's set in TiDB?

lamxTyler: In TiDB, see https://github.com/pingcap/tidb/pull/11861/files#diff-59f10b93675b42905c972e190d2f059dR45

lamxTyler: Yes, see https://github.com/pingcap/tidb/pull/11861/files#diff-59f10b93675b42905c972e190d2f059dR45

eurekaka: It looks pretty weird that the values is set from another repo, can we set it during the induction of parser?

lamxTyler: It seems that there is no way to set it correctly here, since the subpart of select stmt will get the offset first. I don't find good ways to set it correctly here.

Issue #507, opened 2019-08-21
Fix CREATE FULLTEXT...WITH PARSER syntax


tangenta:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
create fulltext index idx on t (a, b) with parser ident
create fulltext index idx on t (a, b) with parser ident comment 'string'
create fulltext index idx on t (a, b) comment 'string' with parser ident
create fulltext index idx on t (a, b) with parser ident lock default
```

Claim this subtask based on the comments below this issue.

yunnian:
Let me fix it

PR #512, opened 2019-08-21
parser: Fix DISCARD/IMPORT TABLESPACE syntax

fzhedu:
### What problem does this PR solve? 
Fix compatibility problem about keyword import/discard tablespace
issue #501 

### What is changed and how it works?
I added a rule to identify the statments in the parser, and it introduced just a few changes in related files (ast/ddl.go, parser.y), as well as added test in parser_test.go.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #524, opened 2019-08-26
Parser: support with parser ident

yunnian:
#### Fix compatibility problem about keyword REMOVE PARTITIONING

Issue: pingcap/parser#507

#### MySQL Syntax:

index_option:
...
  | WITH PARSER parser_name

### Bad SQL Case:

create fulltext index idx on t (a, b) with parser ident
create fulltext index idx on t (a, b) with parser ident comment 'string'
create fulltext index idx on t (a, b) comment 'string' with parser ident
create fulltext index idx on t (a, b) with parser ident lock default

### Check List

Tests
- Unit test


ast/ddl.go
```diff
@@ -1335,7 +1346,7 @@ func (n *CreateIndexStmt) Restore(ctx *RestoreCtx) error {
 	}
 	ctx.WritePlain(")")
 
-	if n.IndexOption.Tp != model.IndexTypeInvalid || n.IndexOption.KeyBlockSize > 0 || n.IndexOption.Comment != "" || n.IndexOption.Visibility != IndexVisibilityDefault {
+	if n.IndexOption.Tp != model.IndexTypeInvalid || n.IndexOption.KeyBlockSize > 0 || n.IndexOption.Comment != "" || len(n.IndexOption.ParserName.O) > 0 || n.IndexOption.Visibility != IndexVisibilityDefault {
```
kennytm: This condition is getting too long. Could you refactor this into a method of  `IndexOption`?

yunnian: let me have a try

yunnian: I tried a lot of ways, but they didn't work well because there were so many influences, such as creating table.

parser.y
```diff
@@ -4186,6 +4190,12 @@ IndexOption:
 			Tp: $1.(model.IndexType),
 		}
 	}
+|	"WITH" "PARSER" Identifier
+	{
+		$$ = &ast.IndexOption {
+			ParserName: model.NewCIStr($3),
+		}
```
leoppro: throw a warning here
see: #512

ast/ddl.go
kennytm: Okay.

PR #528, opened 2019-08-28
parser: implement alter view parser

AndrewDi:
Add support alter view grammer.
[alter-view](https://dev.mysql.com/doc/refman/5.7/en/alter-view.html)
```mysql
ALTER
    [ALGORITHM = {UNDEFINED | MERGE | TEMPTABLE}]
    [DEFINER = user]
    [SQL SECURITY { DEFINER | INVOKER }]
    VIEW view_name [(column_list)]
    AS select_statement
    [WITH [CASCADED | LOCAL] CHECK OPTION]
```
This statement changes the definition of a view, which must exist. The syntax is similar to that for CREATE VIEW see Section 13.1.21, “CREATE VIEW Syntax”). This statement requires the CREATE VIEW and DROP privileges for the view, and some privilege for each column referred to in the SELECT statement. ALTER VIEW is permitted only to the definer or users with the SUPER privilege.

Add AlterViewStmt ast and related yacc.

### Check List 

Tests 
 - Unit test

Related changes


AndrewDi:
@XuHuaiyu PTAL

parser.y
```diff
@@ -3451,6 +3452,48 @@ ViewCheckOption:
...
+			ViewName:      $6.(*ast.TableName),
+			Select:        selStmt,
+			Algorithm:     -1,
+			Security:      -1,
+		}
+		if $2 != nil {
+			x.Algorithm = $2.(model.ViewAlgorithm)
+		}
+		if $3 != nil {
+                	x.Definer = $3.(*auth.UserIdentity)
```
zz-jason: please use tab instead of spaces to indent the code.

AndrewDi:
will implement this later

PR #530, opened 2019-08-29
parser: fix obscure error message in `bit(65)` column type

tangenta:

Fix pingcap/tidb#11923.

Remove the check against field-display width of `BIT` .

The template error message is 
```
ErrTooBigDisplaywidth: "Display width out of range for column '%-.192s' (max = %d)",
```
It requires a column name argument, which is impossible to be accessed in the semantic action of `NumericType`. 

Although we can move the validation code to the two-level-higher `ColumnDef`, it can break the encapsulation. So let TiDB handle it.

Related test: pingcap/tidb#11985

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -2137,7 +2137,7 @@ ColumnDef:
 			yylex.AppendError(yylex.Errorf("Invalid column definition"))
 			return 1
 		}
-        $$ = colDef
+        	$$ = colDef
```
zz-jason: could you replace all the spaces to tab?

tangenta: Done.

kennytm:
> So let TiDB handle it.

Is there a corresponding TiDB PR to reject `BIT(65)`?

tangenta:
@kennytm I have found the related code in TiDB. And I am trying to test it locally.

leoppro:
LGTM

PR #541, opened 2019-09-04
executor: cache point get plan for prepared point 

cfzjywxk:

add field for prepared statement to cache point get plan despite the prepared plan cache switch
the related tidb pull request [https://github.com/pingcap/tidb/pull/11970](link)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/ast/misc.go
+++ b/ast/misc.go
@@ -321,7 +321,7 @@ type Prepared struct {
 	Params        []ParamMarkerExpr
 	SchemaVersion int64
 	UseCache      bool
-	PointPlan     interface{}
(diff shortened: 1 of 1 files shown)


coocood:
LGTM

ast/misc.go
```diff
@@ -321,6 +321,7 @@ type Prepared struct {
 	Params        []ParamMarkerExpr
 	SchemaVersion int64
 	UseCache      bool
+	PointPlan     interface{}
```
lysu: maybe change name to `cachedPlan`

PR #548, opened 2019-09-04
fix CONSTRAINT syntax

### LINKED ISSUE #413: Fix CONSTRAINT syntax
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
ALTER TABLE t ADD ( CHECK ( expr ) )
ALTER TABLE t ADD ( CONSTRAINT CHECK ( expr ) )
ALTER TABLE t ADD COLUMN ( CONSTRAINT CHECK ( expr ) )
ALTER TABLE t ADD COLUMN ( CONSTRAINT CHECK ( expr ) ENFORCED , CHECK ( expr ) )
ALTER TABLE t ADD COLUMN RESOURCE DATE CONSTRAINT CHECK ( expr ) FIRST
ALTER TABLE ident ADD ( CONSTRAINT FOREIGN KEY ident ( EXECUTE ( 123 ) ) REFERENCES t ( a ) MATCH SIMPLE ON DELETE CASCADE ON UPDATE SET NULL )
ALTER TABLE t ADD COLUMN ( CONSTRAINT ident CHECK ( 1>2 ) NOT ENFORCED )
ALTER TABLE t ADD a TINYBLOB CONSTRAINT ident CHECK ( 1>2 ) REFERENCES b ON DELETE CASCADE ON UPDATE CASCADE
```

Claim this subtask based on the comments below this issue.

ichn-hu:

Fix #413 

Tests proposed in #413 reflect 2 different aspects of the incompatibility between our parser and MySQL's.

1. MySQL supports constraints in altering columns, therefore in addtition to modify `parser.y` to support the syntax, I also added a new attribute `NewConstraints` in parallel with already exsited `NewColumns` to `AlterTableSpec`, and also added the corresponding restoration and visit code. The modification is based on observation of how `CreateTableStmt` is implemented, because the syntax is similar in MySQL's `sql_yacc.yy` and our `parser.y`, and it also uses two different array `Cols` and `Constraints` to reflect the 2 different kinds of column definition.
2. The later 2 test cases in #413 is a different incompatibility case, it can be resolved easily as done in [`sql_yacc.yy`](https://github.com/mysql/mysql-server/blob/77306a41670543e103f94fa0db77a71d2b9f88b2/sql/sql_yacc.yy#L6236)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


parser.y
```diff
@@ -1284,12 +1284,24 @@ AlterTableSpec:
 			Position:	$5.(*ast.ColumnPosition),
 		}
 	}
-|	"ADD" ColumnKeywordOpt IfNotExists '(' ColumnDefList ')'
+|       "ADD" ColumnKeywordOpt IfNotExists '(' TableElementList ')'
```
tangenta: Please indent this line with tab.

ichn-hu: Indented, thanks for your review!

ichn-hu:
@tangenta Hi, I've tried hard to increase the coverage, but it still drops, will this block the merge of this PR?

BTW, I'll be very appreciated if you could let me know how to increase the coverage. I figued out that it is because I added code in the visit of `AlterTableSpec`, however it won't be covered by tests in `parser_test.go`, since no visitor will be dispatched for parsing.

leoppro:
> @tangenta Hi, I've tried hard to increase the coverage, but it still drops, will this block the merge of this PR?
> 
> BTW, I'll be very appreciated if you could let me know how to increase the coverage. I figued out that it is because I added code in the visit of `AlterTableSpec`, however it won't be covered by tests in `parser_test.go`, since no visitor will be dispatched for parsing.

This won't block the merge.

To test `Accept` function, you can reference https://github.com/pingcap/parser/blob/master/ast/ddl_test.go#L26

ichn-hu:
@tangenta @leoppro I've added two more test cases in https://github.com/pingcap/parser/blob/master/ast/ddl_test.go#L26, not sure if it will have effect on improving coverage.

But if it won't block the merge, please help review the code and merge it, I guess I won't continue to work on improving coverage.

Issue #525, opened 2019-08-27
column with CHARACTER SET got `Unknown character set` error


martianzhang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

```sql
CREATE TABLE `tb` (    
  `id` int(10) NOT NULL AUTO_INCREMENT,    
  `col` varchar(10) CHARACTER SET gbk DEFAULT NULL,                                                                                                                                      
  PRIMARY KEY (`id`)    
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```
[parser:1115]Unknown character set: 'gbk'

2. What version of TiDB SQL Parser are you using?

41d48df058643bdb2e24c64b1685c4e9ff6608f8

PR #558, opened 2019-09-12
Adds GBK charset

jonahgeorge:

https://github.com/pingcap/parser/issues/525

Added GBK as a default, supported charset and collation as per:
https://mariadb.com/kb/en/library/supported-character-sets-and-collations/

#### Tests
 - Unit test

#### Code changes

#### Side effects
None expected.

#### Related changes
Unsure

### DIFF
--- a/charset/charset.go
+++ b/charset/charset.go
@@ -63,6 +63,7 @@ var charsetInfos = []*Charset{
 	{CharsetASCII, CollationASCII, make(map[string]*Collation), "US ASCII", 1},
 	{CharsetLatin1, CollationLatin1, make(map[string]*Collation), "Latin1", 1},
 	{CharsetBin, CollationBin, make(map[string]*Collation), "binary", 1},
+	{CharsetGBK, CollationGBK, make(map[string]*Collation), "GBK Simplified Chinese", 2},
 }
 
 // All the names supported collations should be in the following table.
@@ -72,6 +73,7 @@ var supportedCollationNames = map[string]struct{}{
 	CollationASCII:   {},
 	CollationLatin1:  {},
 	CollationBin:     {},
+	CollationGBK:     {},
 }
 
 // Desc is a charset description.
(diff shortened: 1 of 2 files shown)


jonahgeorge:
There appear to be failures in TIDB's test suite:
```
FAIL: tables_test.go:181: testTableSuite.TestCharacterSetCollations

tables_test.go:185:
    // The description column is not important
    tk.MustQuery("SELECT default_collate_name, maxlen FROM information_schema.character_sets ORDER BY character_set_name").Check(
        testkit.Rows("ascii_bin 1", "binary 1", "latin1_bin 1", "utf8_bin 3", "utf8mb4_bin 4"))
/go/src/github.com/pingcap/tidb/util/testkit/testkit.go:62:
    res.c.Assert(resBuff.String(), check.Equals, needBuff.String(), res.comment)
... obtained string = "" +
...     "[ascii_bin 1]\n" +
...     "[binary 1]\n" +
...     "[gbk_chinese_ci 2]\n" +
...     "[latin1_bin 1]\n" +
...     "[utf8_bin 3]\n" +
...     "[utf8mb4_bin 4]\n"
... expected string = "" +
...     "[ascii_bin 1]\n" +
...     "[binary 1]\n" +
...     "[latin1_bin 1]\n" +
...     "[utf8_bin 3]\n" +
...     "[utf8mb4_bin 4]\n"
... sql:SELECT default_collate_name, maxlen FROM information_schema.character_sets ORDER BY character_set_name, args:[]
```

Would someone be able to shed some light on the interaction between the parser and TIDB with regards to supported charsets? It seems that the parser should be able to parse statements that indicate a `gbk` charset even if TIDB doesn't currently support it.

shenli:
@jonahgeorge Thanks for your contribution! Currently, TiDB doesn't support GBK charset. So we decide not to show GBK related info. Because this may lead to misunderstanding.

jonahgeorge:
@shenli Prior to https://github.com/pingcap/parser/pull/295, the parser was able to parse statements like below without error.
```sql
CREATE TABLE `tb` (    
  `id` int(10) NOT NULL AUTO_INCREMENT,
  `col` varchar(10) CHARACTER SET gbk DEFAULT NULL,
  PRIMARY KEY (`id`)    
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

Is the intention to continue not supporting this as long as TIDB doesn't support `gbk`?

shenli:
@jonahgeorge Yes, this is by design.

PR #561, opened 2019-09-16
*: refine split region syntax add split partition table region syntax.

crazycs520:
* *: refine split region syntax add split partition table region syntax.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


parser.y
```diff
@@ -2015,6 +2021,31 @@ SplitOption:
...
+	{
+		$$ = &ast.SplitSyntaxOption{}
+	}
+|	"REGION" "FOR"
+	{
+		$$ = &ast.SplitSyntaxOption{
+			HasRegionFor: true,
+		}
+	}
+|	"PARTITION"
```
zimulala: What's the different between `"PARTITION"` and `"REGION" "FOR"`? Could we remove one?

crazycs520: `PARTITION` used to indicate splitting partition table.
`region for` will make this syntax more readable.  `split region for table ` is meaningful than `split table`.

PR #562, opened 2019-09-20
parser: optimize performance

lauhg:
Optimizing parser performance

Here is the benchmark, 8%~12% faster than before:

goos: linux
goarch: amd64

Before optimization:
```sh
BenchmarkSysbenchSelect       200000          6121 ns/op        1824 B/op         20 allocs/op
BenchmarkParseComplex          10000        223586 ns/op        61772 B/op        642 allocs/op
BenchmarkParseSimple          100000         19047 ns/op        5144 B/op         71 allocs/op
```
After optimization:
```sh
BenchmarkSysbenchSelect       200000          5586 ns/op        1824 B/op         20 allocs/op
BenchmarkParseComplex          10000        206988 ns/op        61772 B/op        642 allocs/op
BenchmarkParseSimple          100000         16873 ns/op        5144 B/op         71 allocs/op
```

### Check List
Tests

Unit test

### DIFF
--- a/goyacc/main.go
+++ b/goyacc/main.go
@@ -552,7 +552,6 @@ func %[1]sParse(yylex %[1]sLexer, parser *Parser) int {
 	yyEx, _ := yylex.(%[1]sLexerEx)
 	var yyn int
 	parser.yylval = %[1]sSymType{}
-	parser.yyVAL = %[1]sSymType{}
 	yyS := parser.cache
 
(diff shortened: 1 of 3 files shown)


tiancaiamao:
LGTM
I've tested this branch with TiDB and the CI passed.

tiancaiamao:
PTAL @winkyao @zz-jason

PR #564, opened 2019-09-27
*: suport alter table set flash replica syntax

crazycs520:
This is for TiDB PR:  https://github.com/pingcap/tidb/pull/12453

* Add DDL syntax to support add table flash replica. 
```sql
ALTER TABLE [table_name] SET FLASH REPLICA [count] LOCATION LABELS [location_labels]
eg:
ALTER TABLE t SET FLASH REPLICA 3 LOCATION LABELS "rack", "host", "abc";
```

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


model/model.go
```diff
@@ -323,6 +326,13 @@ func (t TableLockType) String() string {
 	return ""
 }
 
+// FlashReplicaInfo means the flash replica info.
+type FlashReplicaInfo struct {
+	Count          uint64
+	LocationLabels []string
```
lzmhhh123: Should we move the location labels out of the struct and put it in `TableInfo` as a sperate one. Because this one is also an essential part of isolate read.

lzmhhh123: Should we move the location labels out of the struct and put it in `TableInfo` as a sperate one. Because this one is also an essential part of isolate read.

PR #214, opened 2019-02-20
fix now frac in column option

Ryan-Git:
### What problem does this PR solve? 
https://github.com/pingcap/parser/issues/186 is not completely fixed by #211 , this PR parses fraction of CURRENT_TIMESTAMP in ON UPDATE clause.

### What is changed and how it works?
ref parsed NowSymOptionFraction in ColumnOption rather than manually construct it.

### Check List 

Tests 

adjusted related parser_test.go

PR #565, opened 2019-09-28
parser: parse the fraction of CURRENT_TIMESTAMP default value (#211)

tangenta:
Cherry-pick #211 and #214.


zz-jason:
@tangenta It's better to cherry pick the bug-fixes one by one. So that we can tell which PR is not cherry-picked to the release branch through the commit history in git logs.

tangenta:
@zz-jason These two PRs are filed to solve the same problem. Should I append `(#214)` to the end of the title?

Issue #498, opened 2019-08-19
Fix CHARSET DEFAULT syntax


leoppro:
Index Issue: pingcap/tidb#11486

These are bad (incompatible) SQL cases:

```sql
ALTER TABLE d_n.t_n CONVERT TO CHARSET DEFAULT
ALTER TABLE d_n.t_n CONVERT TO CHAR SET DEFAULT
ALTER TABLE d_n.t_n CONVERT TO CHARACTER SET DEFAULT
```

Claim this subtask based on the comments below this issue.

sim41:
Let me fix it.

tangenta:
@sim41  Hi, what is the current progress?

sim41:
@tangenta I’m so sorry for taking such long time. I have finished the testcase, parser.y, but have some problem with the AST. I’m trying to find a nice way to make less difference in AST.

sim41:
sry，I may need some help. sorry for bother you @tangenta 

## what the question is

These are also some bad (incompatible) SQL while set default char set

## testcase

```sql
set character set default;
set names default;
set char set default;
set charset default;
```

## what I have changed in parser.y

```go
CharsetKw CharsetNameOrDefault
        {
                $$ = &ast.VariableAssignment{Name: ast.SetNames, Value: $2}
        }

CharsetNameOrDefault:
        CharsetName
        {
                $$ = ast.NewValueExpr($1.(string))
        }
|       "DEFAULT"
        {
                // it can not work,
                // this node has no accept method, 
                // but it seems like the DefaultExpr in SetExpr node
                $$ = &ast.DefaultExpr{}    
        }
```

## what confused me is that

```go
VariableAssignment:
        VariableName EqOrAssignmentEq SetExpr
        {
                $$ = &ast.VariableAssignment{Name: $1, Value: $3, IsSystem: true}
        }
|       "GLOBAL" VariableName EqOrAssignmentEq SetExpr
        {
                $$ = &ast.VariableAssignment{Name: $2, Value: $4, IsGlobal: true, IsSystem: true}

SetExpr:
        "ON"
        {    
                // this is a node of VariableAssignment type,
                // I want to use this kind of node to 
                // set default value for charset.
                $$ = ast.NewValueExpr("ON")
        }
|       ExprOrDefault

ExprOrDefault:
        Expression
|       "DEFAULT"
        {
                $$ = &ast.DefaultExpr{} 
        }
```

## related definitions

```go
// VariableAssignment is a variable assignment struct.
[...]

PR #567, opened 2019-09-29
fix SET CHARSET | NAMES DEFAULT

sim41:
https://dev.mysql.com/doc/refman/8.0/en/set-character-set.html
It is similar to the issue: https://github.com/pingcap/parser/issues/498. set charset default syntax, but cause of different syntax.

### MySQL Syntax:
```
SET {CHARACTER SET | CHARSET}
    {'charset_name' | DEFAULT}
```

### Bad SQL Case:
```
 set names default;
 set character set default;
 set charset default;
 set char set default;
```

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -686,6 +686,7 @@ import (
 	SignedLiteral			"Literal or NumLiteral with sign"
 	DefaultValueExpr		"DefaultValueExpr(Now or Signed Literal)"
 	NowSymOptionFraction		"NowSym with optional fraction part"
+        CharsetNameOrDefault            "Character set name or default"
```
tangenta: Please reindent this line with tab.

parser.y
```diff
@@ -7326,13 +7327,25 @@ VariableAssignment:
...
+
+CharsetNameOrDefault:
+        CharsetName
+        {
+                $$ = ast.NewValueExpr($1.(string))
+        }
+|       "DEFAULT"
+        {
+                $$ = &ast.DefaultExpr{}
+        }
```
tangenta: Ditto

parser.y
sim41: sry, I used to set expandtab in vim, and gofmt does not work in *.y file. I'll reindent the tab.

leoppro:
LGTM

leoppro:
@tiancaiamao  PTAL

tiancaiamao:
LGTM
Please resolve conflicts @sim41

PR #571, opened 2019-10-08
parser: support syntax for AUTO_RANDOM

tangenta:
This PR adds a keyword, auto_random. The usage is as follows:

```sql
create table t(a bigint auto_random(3) primary key, b varchar(10))
create table t(a int auto_random(3), b char, primary key (a))
```

For a more detail description, please refer to pingcap/tidb#13127.

- Add a field `AutoRandomBitLength uint64` in `ColumnOption`.\
- Add version special comment syntax `/*T!00001 auto_random */`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


ast/ddl.go
```diff
@@ -529,6 +531,8 @@ func (n *ColumnOption) Restore(ctx *RestoreCtx) error {
 	case ColumnOptionStorage:
 		ctx.WriteKeyWord("STORAGE ")
 		ctx.WriteKeyWord(n.StrValue)
+	case ColumnOptionAutoShard:
+		ctx.WritePlainf("AUTO_SHARD(SHARD_BITS = %d)", n.AutoShardBitLength)
```
```suggestion
		ctx.WriteKeyword("AUTO_SHARD")
		ctx.WritePlain("(")
		ctx.WriteKeyword("SHARD_BITS")
		ctx.WritePlainf("= %d)", n.AutoShardBitLength)
```

go.mod1
```diff
@@ -14,3 +14,5 @@ require (
 	github.com/sirupsen/logrus v1.3.0
 	golang.org/x/text v0.3.1-0.20180807135948-17ff2d5776d2
 )
+
+go 1.13
```
kennytm:
```suggestion
```

parser.y
```diff
@@ -2410,6 +2412,10 @@ ColumnOption:
 		yylex.AppendError(yylex.Errorf("The STORAGE clause is parsed but ignored by all storage engines."))
 		parser.lastErrorAsWarn()
 	}
+|	"AUTO_SHARD" '(' "SHARD_BITS" eq LengthNum ')'
```
kennytm: AFAIK we haven't agreed on the exact syntax yet.

tiancaiamao:
PTAL @winkyao @crazycs520

mysql/const.go
```diff
@@ -488,6 +488,12 @@ var DefaultLengthOfMysqlTypes = map[byte]int{
 	TypeSet:    8,
 }
 
+// BitLengthPerByte is the bit number of a byte.
+const BitLengthPerByte = 8
+
+// RowIDBitLength is the bit number of a row id in TiDB.
+const RowIDBitLength = 64
```
crazycs520: Why not define this in `TiDB`?

tangenta: You are right. These are not MySQL-specific. Which place should I put them to? Can I hardcode them?

crazycs520: Put it in the pkg you used in TiDB?

lexer.go
```diff
@@ -55,6 +55,9 @@ type Scanner struct {
 	// lastScanOffset indicates last offset returned by scan().
 	// It's used to substring sql in syntax error message.
 	lastScanOffset int
+
+	// commentCodeVersion means the biggest number can be parsed in the comment with pattern /*!v00001 xxx */
```
kennytm: It mentions `/*!v` here...

lexer.go
```diff
@@ -445,6 +448,21 @@ func startWithSlash(s *Scanner) (tok int, pos Pos, lit string) {
 			}
 		}
 
+		if strings.HasPrefix(comment, "/*v") {
```
kennytm: but here it definitely lacks a `!`.

misc.go
```diff
@@ -14,11 +14,47 @@
...
+func (ccv CommentCodeVersion) String() string {
+	return fmt.Sprintf("%05d", ccv)
+}
+
+func extractVersionCodeInComment(comment string) CommentCodeVersion {
+	if SpecVersionCodePattern.MatchString(comment) {
+		verDigitNum := 5
+		if isDigit(rune(comment[len("/*T!xxxxx")])) {
+			verDigitNum += 1
+		}
```
kennytm: Please use regex captures ([`(*Regexp).FindStringSubmatch`](https://godoc.org/regexp#Regexp.FindStringSubmatch)) to extract the digits instead of this...

tangenta: Updated. PTAL

model/model.go
```diff
@@ -235,6 +235,8 @@ type TableInfo struct {
 	ShardRowIDBits uint64
 	// MaxShardRowIDBits uses to record the max ShardRowIDBits be used so far.
 	MaxShardRowIDBits uint64 `json:"max_shard_row_id_bits"`
+	// AutoRandomBits is used to set the bit number to shard automatically when PKIsHandle.
+	AutoRandomBits uint64 `json:"auto_shard_bits"`
```
kennytm: @tangenta Missed this when reviewing. Is it intended that the JSON field be called `auto_shard_bits` instead of `auto_random_bits`?

tangenta: It is a mistake. I fixed it in #702.

PR #575, opened 2019-10-11
ast: add `CachedNames` for Prepared

winoros:

to compatible with https://github.com/pingcap/tidb/pull/12573

Add a new field

### Check List 

Tests 

 - Unit test


kennytm:
It's still missing the CLA from @winoros

PR #582, opened 2019-10-17
parser: add nowait keyword, add for update no wait grammar, and related error…

cfzjywxk:
… code/msg

support select for update nowait
```
# Session 1:

mysql> CREATE TABLE t (i INT, PRIMARY KEY (i)) ENGINE = InnoDB;

mysql> INSERT INTO t (i) VALUES(1),(2),(3);

mysql> START TRANSACTION;

mysql> SELECT * FROM t WHERE i = 2 FOR UPDATE;
+---+
| i |
+---+
| 2 |
+---+

# Session 2:

mysql> START TRANSACTION;

mysql> SELECT * FROM t WHERE i = 2 FOR UPDATE NOWAIT;
ERROR 3572 (HY000): Statement aborted because lock(s) could not be acquired immediately and NOW
```

### Check List 

Tests 

 - Unit test
 

Code changes

Side effects

Related changes


coocood:
`NOWAIT` is an unreserved keyword, we need to add it to `UnReservedKeyword:` section in the `parser.y`.

cfzjywxk:
> From the description in https://dev.mysql.com/doc/refman/8.0/en/select.html#idm140006656103232, it's more appropriate to make NOWAIT its own modifier rather than grouping `FOR UPDATE NOWAIT` as a whole thing.
> 
> ```sql
>     [FOR {UPDATE | SHARE} [OF tbl_name [, tbl_name] ...] [NOWAIT | SKIP LOCKED] 
>       | LOCK IN SHARE MODE]]
> ```

we're currently groupint "FOR UPDATE" as a whole, and "SelectLockType" is defined as single `int` type,   maybe we need to some refactorations on this for 8.0 compatibility ?  There will be some other related work, do this refactoration in the future?
@coocood @kennytm

coocood:
LGTM

cfzjywxk:
@jackysp @tiancaiamao PTAL

kennytm:
> we're currently groupint "FOR UPDATE" as a whole, and "SelectLockType" is defined as single int type, maybe we need to some refactorations on this for 8.0 compatibility ? There will be some other related work, do this refactoration in the future?

@cfzjywxk IMO it's better to turn `SelectLockType` into a struct now, or add a new member to `SelectStmt` indicating whether it is NOWAIT (and SKIP LOCKED in the future). 

To support `SelectLockForUpdateNoWait` you'll need to modify TiDB once now, and later when you refactor for 8.0 compatibility you'll need to modify the code again. So I think it's better to choose a solution which we know is going to last.

coocood:
@kennytm 
We can modify the code again later when we need to.
It's not certain if we need to support those features.

PR #584, opened 2019-10-22
mysql: add some errors for TiDB

jackysp:
Some errors in TiDB do not have an error code. Add parts of them.

Side effects


go.mod1
```diff
@@ -14,3 +14,5 @@ require (
 	github.com/sirupsen/logrus v1.3.0
 	golang.org/x/text v0.3.1-0.20180807135948-17ff2d5776d2
 )
+
+go 1.13
```
kennytm: @tiancaiamao Is it acceptable to add this now?

mysql/errname.go
```diff
@@ -77,7 +77,7 @@ var MySQLErrName = map[uint16]string{
 	ErrTooLongIdent:                             "Identifier name '%-.100s' is too long",
 	ErrDupFieldName:                             "Duplicate column name '%-.192s'",
 	ErrDupKeyName:                               "Duplicate key name '%-.192s'",
-	ErrDupEntry:                                 "Duplicate entry '%-.192s' for key %d",
+	ErrDupEntry:                                 "Duplicate entry '%s' for key '%s'",
```
kennytm: Note that MySQL 8 would truncate the "entry" to 64 characters. I'm not sure if TiDB truncates the output using the `%-.192s` here or manually.

```sql
mysql> create table x(a varchar(512));
Query OK, 0 rows affected (0.03 sec)

mysql> insert into x values (repeat('a', 512)), (repeat('a', 512));
Query OK, 2 rows affected (0.01 sec)
Records: 2  Duplicates: 0  Warnings: 0

mysql> create unique index y on x(a);
ERROR 1062 (23000): Duplicate entry 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' for key 'y'
```

mysql/errname.go
```diff
@@ -943,6 +943,12 @@ var MySQLErrName = map[uint16]string{
 	ErrUnsupportedReloadPlugin:    "Plugin %s isn't loaded so cannot be reloaded",
 	ErrUnsupportedReloadPluginVar: "Reload plugin with different sysVar is unsupported %v",
 	ErrTableLocked:                "Table '%s' was locked in %s by %v",
+	ErrNotExist:                   "Error: key not exist",
+	ErrTxnRetryable:               "Error: KV error safe to retry %s ",
```
kennytm:
```suggestion
	ErrTxnRetryable:               "Error: KV error safe to retry %s",
```

mysql/errname.go
```diff
@@ -943,6 +943,12 @@ var MySQLErrName = map[uint16]string{
...
+	ErrCannotSetNilValue:          "can not set nil value",
```
kennytm:
```suggestion
	ErrCannotSetNilValue:          "cannot set nil value",
```

jackysp: It keeps the same message as TiDB here.

mysql/errname.go
jackysp: It keeps the same message as TiDB here.

mysql/errname.go
```diff
@@ -931,7 +931,7 @@ var MySQLErrName = map[uint16]string{
 	ErrMemExceedThreshold:         "%s holds %dB memory, exceeds threshold %dB.%s",
 	ErrForUpdateCantRetry:         "[%d] can not retry select for update statement",
 	ErrAdminCheckTable:            "TiDB admin check table failed.",
-	ErrTxnTooLarge:                "Transaction is too large",
+	ErrTxnTooLarge:                "Transaction is too large, size: %d",
```
jackysp: Good advice, however, it's actually supposed to use ErrDupEntryWithKeyName instead of this error. It's hard to choose here. The Duplicate Entry error code that TiDB has been using all the time seems to be wrong. Change the error code, I am afraid there will be some compatibility problems. This error is too common. I don't know if I should follow the same message as ErrDupEntryWithKeyName. @kennytm

mysql/errname.go
```diff
@@ -77,7 +77,7 @@ var MySQLErrName = map[uint16]string{
 	ErrTooLongIdent:                             "Identifier name '%-.100s' is too long",
 	ErrDupFieldName:                             "Duplicate column name '%-.192s'",
 	ErrDupKeyName:                               "Duplicate key name '%-.192s'",
-	ErrDupEntry:                                 "Duplicate entry '%-.192s' for key %d",
+	ErrDupEntry:                                 "Duplicate entry '%-.64s' for key '%-.192s'",
```
jackysp: Good advice, however, it's actually supposed to use ErrDupEntryWithKeyName instead of this error. It's hard to choose here. The Duplicate Entry error code that TiDB has been using all the time seems to be wrong. Change the error code, I am afraid there will be some compatibility problems. This error is too common. I don't know if I should follow the same message as ErrDupEntryWithKeyName. @kennytm

tiancaiamao:
LGTM

kennytm:
The PR is LGTM on its own, but pingcap/tidb#12866 would change the error code of some statements which deviates from MySQL, so I don't know if this is acceptable within our compatibility strategy.

jackysp:
We could wait for https://github.com/pingcap/tidb/pull/12866.

PR #587, opened 2019-10-25
Fixed: failed to parse WHERE / GROUP BY clauses in a SELECT statement without a FROM table

### LINKED ISSUE #115: Failed to parse WHERE / GROUP BY clauses in a SELECT statement without a FROM table
These SQLs:

```sql
select 1 where true;
select 1 group by 1;
select 1 having true;
```

All return one row with the value 1 on MySQL 8.0 (also tested on PostgreSQL 9.6 and SQLite 3.24 with the same behavior), but failed to parse on TiDB 2.1.1 and master:

```
mysql> select 1 where true;
ERROR 1105 (HY000): line 1 column 14 near " true" (total length 19)
mysql> select 1 group by 1;
ERROR 1105 (HY000): line 1 column 14 near " by 1" (total length 19)
mysql> select 1 having true;
ERROR 1105 (HY000): line 1 column 15 near " true" (total length 20)
```

----

Table-less queries which are *correctly accepted* by TiDB:

```sql
select 1 order by 1;
select 1 limit 1;
```

zhenghaoz:
Fixed: Failed to parse WHERE / GROUP BY clauses in a SELECT statement without a FROM table

Fixed #115

### Bad SQL Case:

```sql
select 1 where true;
select 1 group by 1;
select 1 having true;
```

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -6022,45 +6033,26 @@ SelectStmtFromDualTable:
...
+				lastEnd = yyS[yypt-3].offset-1
+			} else if $4 != nil {
+				lastEnd = yyS[yypt-2].offset-1
+			} else if $5 != nil {
 				lastEnd = yyS[yypt-1].offset-1
-			} else if $4 != ast.SelectLockNone {
+			} else if $6 != nil {
 				lastEnd = yyS[yypt].offset-1
 			} else {
 				lastEnd = len(src)
```
tangenta: Here the problem is the assumption: there is only one blank character between productions. In fact, there are cases like 
```
select        ((a+1))                   from t
```

```
FAIL: session_test.go:1250: testSessionSuite.TestFieldText

session_test.go:1273:
    c.Assert(result.Fields()[0].ColumnAsName.O, Equals, tt.field)
... obtained string = "((a+1))    "
... expected string = "((a+1))"
```

To calculate the last offset of a Production, you can use `lastEnd := parser.endOffset(&yyS[yypt-4])`, which handles this case elegantly.

tiancaiamao:
The gramma change seems not right, can we have:

```
select 1 where true group by 1 having 1>0
```

tiancaiamao:
I'm trying to fix but not sure it's better to fix, or rollback this PR first. It seems to introduce other problems. 

Could you find a way to make all the tests pass in https://github.com/pingcap/parser/pull/596 @zhenghaoz

PR #592, opened 2019-10-29
Support full-text search

sunxiaoguang:
Add MATCH(col1,col2,...) AGAINST(expr [search_modifier]) syntax support

Signed-off-by: Xiaoguang Sun <sunxiaoguang@zhihu.com>

Support full-text search syntax

Add MATCH(col1,col2,...) AGAINST(expr [search_modifier]) syntax support

### Check List 

Tests 
 - Unit test
New tests added to expression_test.go and parser_test.go 

Code changes
NO
NO 
NO

Side effects
NA

Related changes
NA


sunxiaoguang:
The coverage test seems to be hard to fix. Testing the error path depends on existing code to return error.

ast/expressions.go
```diff
@@ -1273,3 +1273,56 @@ func (n *MaxValueExpr) Accept(v Visitor) (Node, bool) {
...
+	}
+	ctx.WritePlain(") ")
+	ctx.WriteKeyWord("AGAINST")
+	ctx.WritePlain(" (")
+	if err := n.Against.Restore(ctx); err != nil {
+		return errors.Annotate(err, "An error occurred while restore MatchAgainst.Against")
+	}
+	if n.Modifier&FulltextSearchModifierNaturalLanguageMode == FulltextSearchModifierNaturalLanguageMode {
+		ctx.WritePlain(" IN NATURAL LANGUAGE MODE")
+	}
```
kennytm: ... and don't print `IN NATURAL LANGUAGE MODE` in the restored text as it is redundant

ast/dml.go
```diff
@@ -2605,3 +2605,11 @@ func (n *SplitOption) Restore(ctx *RestoreCtx) error {
 	}
 	return nil
 }
+
+type FulltextSearchModifier int
+
+const (
+	FulltextSearchModifierNaturalLanguageMode = 1 << iota
```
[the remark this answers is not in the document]
sunxiaoguang: Makes sense, modes do not mix.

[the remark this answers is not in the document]
sunxiaoguang: Updated, please talk a look.

ast/dml.go
```diff
@@ -2605,3 +2605,24 @@ func (n *SplitOption) Restore(ctx *RestoreCtx) error {
...
 	return nil
 }
+
+type FulltextSearchModifier int
+
+const (
+	FulltextSearchModifierNaturalLanguageMode = 0
+	FulltextSearchModifierBooleanMode         = 1
+	FulltextSearchModifierModeMask            = 0xF
+	FulltextSearchModifierWithQueryExpansion  = 1 << 4
```
kennytm: (why `1 << 4`? 😅)

parser.y
```diff
@@ -3898,6 +3902,14 @@ Expression:
 			$$ = &ast.UnaryOperationExpr{Op: opcode.Not, V: $2}
 		}
 	}
+|	"MATCH" '(' ColumnNameList ')' "AGAINST" '(' StringLiteral FulltextSearchModifierOpt ')'
```
kennytm: Wait, the AGAINST part can be an arbitrary (string) expression, not just string literal.

```
mysql> select match(a) against (1234), match(a) against (now()), match(a) against (@@sql_mode in boolean mode) from x;
+-------------------------+--------------------------+-----------------------------------------------+
| match(a) against (1234) | match(a) against (now()) | match(a) against (@@sql_mode in boolean mode) |
+-------------------------+--------------------------+-----------------------------------------------+
|                       0 |                        0 |                                             0 |
|                       0 |                        0 |                                             0 |
+-------------------------+--------------------------+-----------------------------------------------+
2 rows in set (0.00 sec)
```

ast/dml.go
sunxiaoguang: In case we want to support more modes compare to MySQL later on.

parser.y
sunxiaoguang: Hm, it looks strange to query from variable though.

kennytm: I think it's better to support it for compatibility with MySQL. I guess const-evaluating the expression in TiDB isn't difficult.

sunxiaoguang: Agree, let me fix it.

sunxiaoguang: @kennytm

ast/expressions.go
```diff
@@ -1273,3 +1274,70 @@ func (n *MaxValueExpr) Accept(v Visitor) (Node, bool) {
...
+	fmt.Fprint(w, ")")
+}
+
+func (n *MatchAgainst) Accept(v Visitor) (Node, bool) {
+	newNode, skipChildren := v.Enter(n)
+	if skipChildren {
+		return v.Leave(newNode)
+	}
+	return v.Leave(n)
+}
```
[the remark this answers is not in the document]
sunxiaoguang: Sure, let me try to understand what Accept does 😂

sunxiaoguang:
> LGTM

Thanks, But what does codecov/patch mean? I have no idea what is it about.

kennytm:
@sunxiaoguang it means the coverage of the _new_ code is not enough by codecov's standard. I won't worry about that too much since it is not required to pass.

kennytm:
PTAL @tiancaiamao or @tangenta

tiancaiamao:
LGTM

PR #595, opened 2019-10-30
mysql: add 2 infoschema related errors for TiDB

bb7133:
Some errors in TiDB do not have an error code. Add parts of them.

Side effects

### DIFF
--- a/mysql/errname.go
+++ b/mysql/errname.go
@@ -949,7 +949,7 @@ var MySQLErrName = map[uint16]string{
 	ErrInvalidTxn:                 "invalid transaction",
 	ErrEntryTooLarge:              "entry too large, the max entry size is %d, the size of data is %d",
 	ErrNotImplemented:             "not implemented",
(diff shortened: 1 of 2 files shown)


mysql/errname.go
```diff
@@ -949,6 +949,8 @@ var MySQLErrName = map[uint16]string{
 	ErrInvalidTxn:                 "invalid transaction",
 	ErrEntryTooLarge:              "entry too large, the max entry size is %d, the size of data is %d",
 	ErrNotImplemented:             "not implemented",
+	ErrInfoSchemaExpired:          "Information schema is out of date: schema failed to update in 1 lease",
```
winkyao: "Please make sure TiDB can connect to TiKV nodes"

lonng: I think the error message starting with the lower case is more idiomatic.

bb7133: Addressed

bb7133: @lonng I agree with you, however, most of the existing error message has upper case prefix. So it's better to keep this style.

PR #597, opened 2019-10-31
revert pr 587 failed to parse WHERE / GROUP BY clauses in a SELECT statement without a FROM table

tiancaiamao:
Revert https://github.com/pingcap/parser/pull/587


tiancaiamao:
PTAL @tangenta @kennytm

PR #610, opened 2019-11-02
Add Cloud Spanner style read only transaction

sunxiaoguang:
Add extra syntax for Cloud Spanner style read only transaction

Tests 
 - Unit test
 parser_test.go was modified to include tests for new syntax

Code changes
No
Yes
No


ast/misc.go
```diff
@@ -398,6 +422,13 @@ func (n *BeginStmt) Accept(v Visitor) (Node, bool) {
 		return v.Leave(newNode)
 	}
 	n = newNode.(*BeginStmt)
+	if n.Bound != nil && n.Bound.Timestamp != nil {
+		newTimestamp, ok := n.Bound.Timestamp.Accept(v)
+		if !ok {
+			return n, false
+		}
+		n.Bound.Timestamp = newTimestamp.(ExprNode)
```
tiancaiamao: I guess this should work:

`n.Bound.Timestamp = newTimestamp`

parser.y
```diff
@@ -295,6 +295,7 @@ import (
 	block		"BLOCK"
 	booleanType	"BOOLEAN"
 	boolType	"BOOL"
+	bound		"BOUND"
```
tiancaiamao: According to https://dev.mysql.com/doc/refman/8.0/en/keywords.html

BOUND is not a keyword, so it should be put to the NotKeywordToken section.

parser.y
```diff
@@ -341,6 +342,7 @@ import (
 	event		"EVENT"
 	events		"EVENTS"
 	escape 		"ESCAPE"
+	exact 		"EXACT"
```
tiancaiamao: And `EXACT` is also a NotKeywordToken

parser.y
```diff
@@ -469,12 +471,14 @@ import (
...
 	sqlTsiWeek	"SQL_TSI_WEEK"
 	sqlTsiYear	"SQL_TSI_YEAR"
+	staleness	"STALENESS"
 	start		"START"
 	statsAutoRecalc	"STATS_AUTO_RECALC"
 	statsPersistent	"STATS_PERSISTENT"
 	statsSamplePages	"STATS_SAMPLE_PAGES"
 	status		"STATUS"
 	storage		"STORAGE"
+	strong		"STRONG"
```
tiancaiamao: ditto

parser.y
```diff
@@ -469,12 +471,14 @@ import (
 	sqlTsiSecond	"SQL_TSI_SECOND"
 	sqlTsiWeek	"SQL_TSI_WEEK"
 	sqlTsiYear	"SQL_TSI_YEAR"
+	staleness	"STALENESS"
```
tiancaiamao: ditto

parser.y
```diff
@@ -2208,10 +2213,63 @@ BeginTransactionStmt:
...
+	{
+		$$ = &ast.BeginStmt{}
+	}
 |	"START" "TRANSACTION" "WITH" "CONSISTENT" "SNAPSHOT"
 	{
 		$$ = &ast.BeginStmt{}
 	}
+|	"START" "TRANSACTION" "READ" "ONLY"
+	{
+		$$ = &ast.BeginStmt{
```
tiancaiamao: TiDB should handle this later on, because MySQL also has "start transaction read only" and we should keep the compatibility.

parser.y
```diff
@@ -4482,7 +4540,7 @@ UnReservedKeyword:
 | "WITHOUT" | "RTREE" | "EXCHANGE" | "COLUMN_FORMAT" | "REPAIR" | "IMPORT" | "DISCARD" | "TABLE_CHECKSUM" | "UNICODE"
 | "SQL_TSI_DAY" | "SQL_TSI_HOUR" | "SQL_TSI_MINUTE" | "SQL_TSI_MONTH" | "SQL_TSI_QUARTER" | "SQL_TSI_SECOND" |
 "SQL_TSI_WEEK" | "SQL_TSI_YEAR" | "INVISIBLE" | "VISIBLE" | "TYPE" | "NOWAIT" | "REPLICA" | "LOCATION" | "LABELS"
-| "LOGS" | "HOSTS" | "AGAINST" | "EXPANSION"
+| "LOGS" | "HOSTS" | "AGAINST" | "EXPANSION" | "BOUND" | "EXACT" | "STALENESS" | "STRONG"
```
tiancaiamao: Those belong to NotKeywordToken

parser_test.go
```diff
@@ -4585,6 +4585,22 @@ func (s *testParserSuite) TestFulltextSearch(c *C) {
 	c.Assert(writer.String(), Equals, "MATCH(title,content) AGAINST(\"search\" WITH QUERY EXPANSION)")
 }
 
+func (s *testParserSuite) TestStartTransaction(c *C) {
+	cases := []testCase{
+		{"START TRANSACTION READ WRITE", true, "START TRANSACTION"},
+		{"START TRANSACTION READ ONLY", true, "START TRANSACTION READ ONLY"},
+		{"START TRANSACTION READ ONLY WITH TIMESTAMP BOUND", false, ""},
+		{"START TRANSACTION READ ONLY WITH TIMESTAMP BOUND STRONG", true, "START TRANSACTION READ ONLY WITH TIMESTAMP BOUND STRONG"},
+		{"START TRANSACTION READ ONLY WITH TIMESTAMP BOUND MAX STALENESS '00:00:10'", true, "START TRANSACTION READ ONLY WITH TIMESTAMP BOUND MAX STALENESS '00:00:10'"},
```
tiancaiamao: Is this SQL a valid one?

```
START TRANSACTION READ ONLY WITH TIMESTAMP BOUND MAX STALENESS 1 + 1
```

ast/misc.go
sunxiaoguang: nope, it complains😂
> Node does not implement ExprNode (missing Format method)

parser.y
sunxiaoguang: Ok, I see. I thought none keywords go to UnReservedKeyword. Let me fix them.

parser.y
sunxiaoguang: 'START TRANSACTION READ ONLY' is parsed separately which doesn't specify any Bound therefore can be have different execution plan on TiDB.

In addition it would have the same semantic as 'START TRANSACTION READ ONLY WITH TIMESTAMP BOUND STRONG'. Therefore we can add them together in this PR.

parser_test.go
sunxiaoguang: `STALENESS` should take an expression that can be evaluated to `TIME`.
`TIMESTAMP` would take an expression that can be evaluated to `DATETIME` or `TIMESTAMP` on the other hand.

ast/dml.go
```diff
@@ -2629,3 +2629,16 @@ func (m FulltextSearchModifier) IsNaturalLanguageMode() bool {
...
 	return m&FulltextSearchModifierWithQueryExpansion == FulltextSearchModifierWithQueryExpansion
 }
+
+type TimestampBound struct {
+	Mode      int
+	Timestamp ExprNode
+}
+
+const (
+	TimestampBoundStrong = iota
```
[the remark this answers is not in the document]
sunxiaoguang: Fixed

tiancaiamao:
LGTM

ast/misc.go
```diff
@@ -377,13 +377,37 @@ func (n *ExecuteStmt) Accept(v Visitor) (Node, bool) {
...
-		ctx.WriteKeyWord("START TRANSACTION")
+		if n.ReadOnly {
+			ctx.WriteKeyWord("START TRANSACTION READ ONLY")
+			if n.Bound != nil {
+				switch n.Bound.Mode {
+				case TimestampBoundStrong:
+					ctx.WriteKeyWord(" WITH TIMESTAMP BOUND STRONG")
+				case TimestampBoundMaxStaleness:
+					ctx.WriteKeyWord(" WITH TIMESTAMP BOUND MAX STALENESS ")
+					n.Bound.Timestamp.Restore(ctx)
```
kennytm: The errors returned by `Restore()` should be handled.

sunxiaoguang: Fixed, PTAL.

PR #617, opened 2019-11-06
*: support add/drop primary key

zimulala:
Support add/drop primary key.

Support add/drop primary key type information.

### Check List 

Tests 
 - Unit test

Related changes


model/ddl.go
```diff
@@ -61,10 +61,15 @@ const (
...
+	ActionDropPrimaryKey                ActionType = 33
 )
 
-// AddIndexStr is a string related to the operation of "add index".
-const AddIndexStr = "add index"
+const (
+	// AddIndexStr is a string related to the operation of "add index".
+	AddIndexStr      = "add index"
+	AddPrimaryKeyStr = "add primary key"
+)
```
tangenta: Is it necessary to add another layer like `xxxStr`?

zimulala: I feel that there is no need at present, this two are needed in other packages.

zimulala:
@kennytm 
This PK name has no meaning, but it will report an error if its length exceeds the limit.

PR #621, opened 2019-11-07
model: add hidden to ColumnInfo

wjhuang2016:
To support `hidden column`, we need to add a field in ColumnInfo to indicate if the column is hiddened. A hidden column can't be selected, inserted, updated or deleted.
Besies, `SHOW EXTENDED COLUMNS FROM XXX` can read the hidden column, support it here.

Add a filed in ColumnInfo, update parser to support `EXTENDED` in show columns statement.

### Check List 

Tests 

 - N/A

Code changes


parser_test.go
```diff
@@ -866,6 +866,10 @@ func (s *testParserSuite) TestDBAStmt(c *C) {
 		// for show full columns
 		{"show columns in t;", true, "SHOW COLUMNS IN `t`"},
 		{"show full columns in t;", true, "SHOW FULL COLUMNS IN `t`"},
+		// for show extended columns
+		{`SHOW COLUMNS FROM City;`, true, "SHOW COLUMNS IN `City`"},
+		{`SHOW EXTENDED COLUMNS FROM City;`, true, "SHOW EXTENDED COLUMNS IN `City`"},
+		{`SHOW EXTENDED FIELDS FROM City;`, true, "SHOW EXTENDED COLUMNS IN `City`"},
```
djshow832: How about `show extended full`?

wjhuang2016: Done

djshow832:
LGTM

model/model.go
```diff
@@ -93,6 +93,7 @@ type ColumnInfo struct {
 	types.FieldType     `json:"type"`
 	State               SchemaState `json:"state"`
 	Comment             string      `json:"comment"`
+	Hidden              bool        `json:"hidden"`
```
bb7133: Please add a comment for this attribute.

model/model.go
```diff
@@ -93,6 +93,8 @@ type ColumnInfo struct {
 	types.FieldType     `json:"type"`
 	State               SchemaState `json:"state"`
 	Comment             string      `json:"comment"`
+	// A hidden column is used internally(expression index) and are not accessible by users.
```
bb7133:
```suggestion
	// A hidden column is used internally(by expression index) and is not accessible to the users.
```

PR #612, opened 2019-11-05
parser: add error number for store limit error

AilinKid:
### What problem does this PR solve? 
add store limit error and name for PR [#12779](https://github.com/pingcap/tidb/pull/12779)

### What is changed and how it works?
add error code and error name in mysql package

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below)

Related changes

 - Need to cherry-pick to the release branch

PR #636, opened 2019-11-14
parser : add error number for store limit error (#612)

AilinKid:
cherry-pick #612 to release-3.0

add error number for store limit error in mysql package

### Check List 

Tests 

Release note
 - add error number for store limit error (#612)

### DIFF
--- a/go.mod1
+++ b/go.mod1
@@ -14,3 +14,5 @@ require (
 	github.com/sirupsen/logrus v1.3.0
 	golang.org/x/text v0.3.1-0.20180807135948-17ff2d5776d2
 )
+
+go 1.13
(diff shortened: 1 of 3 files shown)


tangenta:
The error reported in CI says 
```
store/tikv/2pc.go:753:69: time.Since(lockWaitStartTime).Milliseconds undefined (type time.Duration has no field or method Milliseconds)
```
The method `Milliseconds` of `Duration` is added go 1.13, it seems we need to set the go version to 1.13 in go.mod.

AilinKid:
> The error reported in CI says
> 
> ```
> store/tikv/2pc.go:753:69: time.Since(lockWaitStartTime).Milliseconds undefined (type time.Duration has no field or method Milliseconds)
> ```
> 
> The method `Milliseconds` of `Duration` is added go 1.13, it seems we need to set the go version to 1.13 in go.mod.

Thanks very much, let me try it.

PR #645, opened 2019-11-19
[WIP] parser: support select into syntax

lauhg:
support select into syntax
related PR: https://github.com/pingcap/tidb/pull/13603

Most of the code are written during the Hackathon, so it may be messy. 
We have implemented:
* select ... into variables
* select ... into outfile
* select ... into dumpfile

The parser handled some cases of the usage of the select-into with union and subquery statements, it's not completed.  sorry, but, fix me please.

### Check List 

Tests 

 - Unit test


tisonkun:
DONE as https://github.com/pingcap/parser/pull/745

PR #656, opened 2019-12-04
parser: introduce DigestNormalized and deprecate DigestHash

SunRunAway:

part of https://github.com/pingcap/tidb/issues/13871

Introduce `DigestNormalized` and deprecate `DigestHash`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - None

Related changes

 - None


digester.go
```diff
@@ -18,23 +18,39 @@ import (
...
 	digesterPool.Put(d)
 	return
 }
 
+// DigestNormalized generates the digest of a normalized sql.
+// it will generate a hash on a normalized sql.
+// Normalize + DigestNormalized equals to NormalizeDigest.
+//
+// for example: DigestNormalized('select 1') and DigestNormalized('select 2') is not the same
+func DigestNormalized(normalized string) (result string) {
```
lamxTyler: Why not just `digest`? It does not care about if it is normalized.

SunRunAway: See https://github.com/pingcap/tidb/issues/13871#issuecomment-561464681 and https://github.com/pingcap/tidb/issues/13871#issuecomment-561516174

lamxTyler: Oh, so we cannot remove the `DigestHash` or rename it for compatibility reasons? 
The name is quite confusion because it does not say it will normalized the string.

digester.go
```diff
@@ -72,6 +88,19 @@ type sqlDigester struct {
...
 }
 
+func (d *sqlDigester) doDigestNormalized(normalized string) (result string) {
+	hdr := *(*reflect.StringHeader)(unsafe.Pointer(&normalized))
+	b := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
+		Data: hdr.Data,
+		Len:  hdr.Len,
+		Cap:  hdr.Len,
+	}))
+	d.hasher.Write(b)
```
kennytm: Are you sure this is anything different from `d.hasher.Write([]byte(normalized))`? ([`io.WriteString`](https://golang.org/src/io/io.go#L289) also implements this just by a direct `[]byte` cast.)

SunRunAway: Here's an example https://play.golang.org/p/cYoqQ-VuibO, please run it locally.

digester.go
SunRunAway: Oh, the name of this function is suggested by @lysu , It's intended to by called with a normalized SQL string, otherwise it is meaningless.

SunRunAway: I've updated the comment of this function.

PR #658, opened 2019-12-04
DDL : add create sequence support & compatible with mariaDB except OrReplace.

AilinKid:
SUPPORT CREATE SEQUENCE in TiDB !!!
TiDB are compatible with MySQL's protocol, we should be also compatible with SEQUENCE implementation of MariaDB cause MySQL is absent from it.
```
*      CREATE [TEMPORARY] SEQUENCE [IF NOT EXISTS] sequence_name
*      [ INCREMENT [ BY | = ] increment ]
*      [ MINVALUE [=] minvalue | NO MINVALUE | NOMINVALUE ]
*      [ MAXVALUE [=] maxvalue | NO MAXVALUE | NOMAXVALUE ]
*      [ START [ WITH | = ] start ]
*      [ CACHE [=] cache | NOCACHE | NO CACHE]
*      [ CYCLE | NOCYCLE | NO CYCLE]
*      [ ORDER | NOORDER | NO ORDER]
*      [table_options]
```

This PR just put the `CREATE SEQUENCE` forward for discussion and review.  Sequence has many other parser syntax to be filled, come and join us!

Warm yourself up with [MariaDB Sequence](https://mariadb.com/kb/en/library/sequences/) and [TiDB Parser](https://pingcap.com/blog-cn/30mins-become-contributor-of-tidb-20190808/)
### Syntax To Be Supported
- [ ] ```SHOW CREATE SEQUENCE sequence_name;```
- [ ] ```ALTER SEQUENCE [IF EXISTS] sequence_name
[ INCREMENT [ BY | = ] increment ]
[ MINVALUE [=] minvalue | NO MINVALUE | NOMINVALUE ]
[ MAXVALUE [=] maxvalue | NO MAXVALUE | NOMAXVALUE ]
[ START [ WITH | = ] start ] [ CACHE [=] cache ] [ [ NO ] CYCLE ]
[ RESTART [[WITH | =] restart];```
- [ ] ```DROP [TEMPORARY] SEQUENCE [IF EXISTS] [/*COMMENT TO SAVE*/]
    sequence_name [, sequence_name] ...```
- [ ] ```Sequence Function (need discussion)```

### Check List 

Tests 

 - Unit test
 - Integration test


AilinKid:
make parser will occur `reduce conflict`, what can I do with this?

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = []*ast.SequenceOption{$1.(*ast.SequenceOption)}
+	}
+|	SequenceOptionList SequenceOption
+	{
+		$$ = append($1.([]*ast.SequenceOption), $2.(*ast.SequenceOption))
+	}
+
+SequenceOption:
+	"INCREMENT" "BY" NUM
```
kennytm: Consider supporting these additional syntaxes:

* INCREMENT = *num*
* INCREMENT *num*

Also make sure the syntax accepts _negative_ numbers.

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceOptionIncrementBy, IntValue: $3.(int64),}
+	}
+|	"START" "WITH" NUM
```
kennytm: Consider supporting these additional syntax:

* START = *num*
* START *num*

Again, the number can be negative.

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceStartWith, IntValue: $3.(int64),}
+	}
+|	"NOMINVALUE"
```
kennytm: Support:

* NO MINVALUE

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceNoMinValue,}
+	}
+|	"MINVALUE" NUM
```
kennytm: Support:

* MINVALUE = *num*

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceMinValue, IntValue: $2.(int64),}
+	}
+|	"NOMAXVALUE"
```
kennytm: Support

* NO MAXVALUE

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceNoMaxValue,}
+	}
+| 	"MAXVALUE" NUM
```
kennytm: Support

* MAXVALUE = *num*

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceMaxValue, IntValue: $2.(int64),}
+	}
+|	"NOCACHE"
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceNoCache,}
+	}
+|	"CACHE" NUM
```
kennytm: Support

* CACHE = *num*

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceCache, IntValue: $2.(int64),}
+	}
+|	"NOCYCLE"
```
kennytm: Support

* NO CYCLE

parser.y
```diff
@@ -10771,4 +10786,95 @@ LoadStatsStmt:
...
+ *          [START WITH <start>]
+ *	    [MINVALUE <minvalue> | NOMINVALUE]
+ *          [MAXVALUE <maxvalue> | NOMAXVALUE]
+ *          [CACHE <cache>]
+ *          [CYCLE | NO CYCLE]
+ *          [ORDER | NO ORDER]
+ ********************************************************************************************/
+
+CreateSequenceStmt:
+	"CREATE" "SEQUENCE" IfNotExists TableName CreateSequenceOptionListOpt
```
kennytm: Add an `OptTemporary` here since MariaDB allows creating temporary sequences.
```suggestion
	"CREATE" OptTemporary "SEQUENCE" IfNotExists TableName CreateSequenceOptionListOpt
```

Also, MariaDB also allows all table options to be used here, e.g.

```sql
CREATE SEQUENCE xyz START WITH 41 COMMENT 'this is a sequence'
```

so perhaps include `CreateTableOptionListOpt` at the end as well.

tangenta:
> make parser will occur `reduce conflict`, what can I do with this?
> `y.output` is hard to analyze.

@AilinKid That is because of the conflict between `ReservedKeyword` and `MaxValueOrExpression` on `MAXVALUE`. You can search "conflict on" in y.output.

BTW, `MAXVALUE` is a reserved keyword in MySQL.

parser.y
AilinKid: Thanks kennytm, that does a great help to me!

parser.y
AilinKid: Addressed.

parser.y
AilinKid: Does `NUM` won't parse number with a signed symbol?

parser.y
AilinKid: Addressed.

parser.y
AilinKid: Addressed.

parser.y
AilinKid: Addressed.

parser.y
AilinKid: Addressed.

AilinKid:
> > make parser will occur `reduce conflict`, what can I do with this?
> > `y.output` is hard to analyze.
> 
> @AilinKid That is because of the conflict between `ReservedKeyword` and `MaxValueOrExpression` on `MAXVALUE`. You can search "conflict on" in y.output.
> 
> BTW, `MAXVALUE` is a reserved keyword in MySQL.
> https://dev.mysql.com/doc/refman/8.0/en/keywords.html

Thanks tangenta! did me a great favor

parser.y
kennytm: @AilinKid No. That's why `SignedLiteral` is a different rule.

parser.y
```diff
@@ -10771,4 +10788,134 @@ LoadStatsStmt:
...
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceOptionIncrementBy, IntValue: $4.(int64),}
+	}
+|	"START" NUM
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceStartWith, IntValue: $2.(int64),}
+	}
+|	"START" "=" NUM
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceStartWith, IntValue: $3.(int64),}
+	}
```
kennytm: You could use `EqOpt` to merge these rules.

```
|	"START" EqOpt SignedNum
```

parser.y
```diff
@@ -10771,4 +10788,134 @@ LoadStatsStmt:
...
+			SeqOptions: $7.([]*ast.SequenceOption),
+			TblOptions: $8.([]*ast.TableOption),
+		}
+	}
+
+CreateSequenceOptionListOpt:
+	{
+		$$ = []*ast.SequenceOption{}
+	}
+|       SequenceOptionList
```
kennytm: Reindent with tabs

parser.y
```diff
@@ -8639,6 +8655,7 @@ Statement:
 |	CreateUserStmt
 |	CreateRoleStmt
 |	CreateBindingStmt
+|       CreateSequenceStmt
```
kennytm: Reindent with tabs

parser.y
AilinKid: Nice catch!

parser.y
```diff
@@ -10771,4 +10789,127 @@ LoadStatsStmt:
...
+	{
+		$$ = append($1.([]*ast.SequenceOption), $2.(*ast.SequenceOption))
+	}
+
+SequenceOption:
+	"INCREMENT" EqOpt SignedNum
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceOptionIncrementBy, IntValue: $3.(int64),}
+	}
+|	"INCREMENT" "BY" EqOpt SignedNum
```
```suggestion
|	"INCREMENT" "BY" SignedNum
```

The syntax `[ BY | = ]` in https://mariadb.com/kb/en/library/create-sequence/ means that we can either ignore it, choose `BY`, or choose `=`, but not both.

parser.y
```diff
@@ -10771,4 +10789,127 @@ LoadStatsStmt:
...
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceOptionIncrementBy, IntValue: $4.(int64),}
+	}
+|	"START" EqOpt SignedNum
+	{
+		$$ = &ast.SequenceOption{ Tp:ast.SequenceStartWith, IntValue: $3.(int64),}
+	}
+|	"START" "WITH" EqOpt SignedNum
```
```suggestion
|	"START" "WITH" SignedNum
```

ditto

parser.y
```diff
@@ -10771,4 +10789,127 @@ LoadStatsStmt:
...
+	{
+		$$ = $1.(int64)
+	}
+|	'+' NUM
+	{
+		$$ = $2.(int64)
+	}
+|	'-' NUM
+	{
+		$$ = $2.(int64) * -1
```
```suggestion
		$$ = -$2.(int64)
```

😳

AilinKid:
conflict on temporary, shift, and goto state 1948, reduce using rule 343 // temporary: assoc %precedence, prec 27.

seems put `or replace` and `temporary` together will make parser conflict.
I've tried some error way before, it will lead `create temporary ...` parse error, tother statement is ok in parser_test.go.

kennytm:
@AilinKid Let's not support `OR REPLACE` for now. MariaDB includes `OR REPLACE` because [it also supports `CREATE OR REPLACE TABLE`](https://mariadb.com/kb/en/library/create-table/), which is not the case for us nor MySQL 8. So I think we could skip `CREATE OR REPLACE SEQUENCE` until we decide to implement `CREATE OR REPLACE TABLE`.

AilinKid:
> @AilinKid Let's not support `OR REPLACE` for now. MariaDB includes `OR REPLACE` because [it also supports `CREATE OR REPLACE TABLE`](https://mariadb.com/kb/en/library/create-table/), which is not the case for us nor MySQL 8. So I think we could skip `CREATE OR REPLACE SEQUENCE` until we decide to implement `CREATE OR REPLACE TABLE`.

Make sense, thanks a lot kennytm.

AilinKid:
PTAL @kennytm @tiancaiamao

PR #664, opened 2019-12-06
ddl: add expression index support

wjhuang2016:
Support creating expression index.
According to https://dev.mysql.com/doc/refman/8.0/en/create-index.html,
Expression part can't be used in primary index、 fulltext index and foreign key.

Modify yacc to support expression part, rename `IndexColName` to `IndexColNameWithExpr` to hold ExprNode, and add some test.
And add error code related to expression index:https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


ast/ddl.go
```diff
@@ -192,18 +192,27 @@ func (n *DropDatabaseStmt) Accept(v Visitor) (Node, bool) {
 	return v.Leave(n)
 }
 
-// IndexColName is used for parsing index column name from SQL.
-type IndexColName struct {
+// IndexColNameWithExpr is used for parsing index column name from SQL.
+type IndexColNameWithExpr struct {
```
djshow832: Why do you rename it? I think the original name is enough.

wjhuang2016: The original name is `ColName`, but it may be an expression.

ast/ddl_test.go
```diff
@@ -68,12 +68,23 @@ func (ts *testDDLSuite) TestDDLVisitorCover(c *C) {
 }
 
 func (ts *testDDLSuite) TestDDLIndexColNameRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"(a + 1)", "(`a`+1)"},
+		{"(1 * 1 + (1 + 1))", "(1*1+(1+1))"},
```
Deardrops: Adding one test case: (double brackets) 
`((1 * 1 + (1 + 1)))`

ast/ddl.go
Deardrops: I think use `IndexColName` is good enough, because these are only one `IndexColName`, this name does not bring any ambiguous.

wjhuang2016:
> ```
> mysql> create table t (a int, index ((a)));
> ERROR 3762 (HY000): Functional index on a column is not supported. Consider using a regular index instead.
> mysql> create table t (a int, index ((d)));
> ERROR 3762 (HY000): Functional index on a column is not supported. Consider using a regular index instead.
> ```
> 
> This error should reported by parser.

Why should be reported by parser？ It seems that this is not a parser error in MySQL

Deardrops:
This error report by TiDB is also acceptable.

ast/ddl.go
```diff
@@ -192,18 +192,27 @@ func (n *DropDatabaseStmt) Accept(v Visitor) (Node, bool) {
 	return v.Leave(n)
 }
 
-// IndexColName is used for parsing index column name from SQL.
-type IndexColName struct {
+// KeyPartSpecifications is used for parsing index column name or index expression from SQL.
+type KeyPartSpecification struct {
```
zimulala: I think it's better to use the name with the "Index" prefix because our previous index-related structures were prefixed with "Index" instead of "Key".

Deardrops:
The variable naming need more discussion, rest LTGM

parser.y
```diff
@@ -2964,30 +2955,45 @@ CreateIndexStmt:
...
 	}
 
-IndexColName:
-	ColumnName OptFieldLen Order
+KeyPartSpecificationListOpt:
 	{
-		//Order is parsed but just ignored as MySQL did
-		$$ = &ast.IndexColName{Column: $1.(*ast.ColumnName), Length: $2.(int)}
+        	$$ = ([]*ast.KeyPartSpecification)(nil)
+        }
```
tangenta: Please reindent these lines with tab.

parser.y
```diff
@@ -2964,30 +2955,45 @@ CreateIndexStmt:
...
+		$$ = append($1.([]*ast.KeyPartSpecification), $3.(*ast.KeyPartSpecification))
 	}
 
-IndexColNameList:
-	IndexColName
+KeyPartSpecification:
+	ColumnName OptFieldLen Order
 	{
-		$$ = []*ast.IndexColName{$1.(*ast.IndexColName)}
+		//Order is parsed but just ignored as MySQL did
```
tangenta:
```suggestion
		// Order is parsed but just ignored as MySQL did.
```

parser.y
```diff
@@ -898,6 +895,9 @@ import (
 	IndexType			"index type"
 	IndexTypeName			"index type name"
 	IndexTypeOpt			"optional index type"
+	IndexPartSpecification		"Index column name or expression"
+	IndexPartSpecificationList	"List of index column name or expression"
+	IndexPartSpecificationListOpt     "Optional list of index column name or expression"
```
tangenta: It's still not formated... Fine, I think we should develop a format tool for *.y.

kennytm: @tangenta Please create an issue 😄

ast/ddl.go
```diff
@@ -193,18 +193,27 @@ func (n *DropDatabaseStmt) Accept(v Visitor) (Node, bool) {
...
 	Column *ColumnName
 	Length int
+	Expr   ExprNode
 }
 
 // Restore implements Node interface.
-func (n *IndexColName) Restore(ctx *RestoreCtx) error {
+func (n *IndexPartSpecification) Restore(ctx *RestoreCtx) error {
+	if n.Expr != nil {
+		ctx.WritePlainf("(")
```
```suggestion
		ctx.WritePlain("(")
```

ast/ddl.go
```diff
@@ -193,18 +193,27 @@ func (n *DropDatabaseStmt) Accept(v Visitor) (Node, bool) {
...
+		if err := n.Expr.Restore(ctx); err != nil {
+			return errors.Annotate(err, "An error occurred while splicing IndexPartSpecifications")
+		}
+		ctx.WritePlainf(")")
```
```suggestion
		ctx.WritePlain(")")
```

PR #666, opened 2019-12-06
support require clauses in `grant` stmt

lysu:

ref https://github.com/pingcap/tidb/issues/9708

`create user`, `alter user` already support require syntax, `grant` also need that

- make grant support require clause
- fix `and` is optional

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - impl change

Side effects

 - N/A

Related changes

 - Need to cherry-pick to the release branch 3.0

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/666)


ast/misc.go
```diff
@@ -1952,6 +1952,7 @@ type GrantStmt struct {
 	ObjectType ObjectTypeType
 	Level      *GrantLevel
 	Users      []*UserSpec
+	TslOptions []*TslOption
```
kennytm: Is it too late to fix the spelling? It should be "TLS" 😅

lysu:
@jackysp @tiancaiamao  PTAL,  - - decrease codecov is caused by rename... maybe we can merge parse change first and move on other PRS.

ast/misc.go
```diff
@@ -1987,6 +1988,21 @@ func (n *GrantStmt) Restore(ctx *RestoreCtx) error {
...
+		if tlsOptionLen != 0 {
+			ctx.WriteKeyWord(" REQUIRE ")
+		}
+		// Restore `TLSOptions` reversely to keep order the same with original sql
+		for i := tlsOptionLen; i > 0; i-- {
+			if i != tlsOptionLen {
+				ctx.WriteKeyWord(" AND ")
+			}
+			if err := n.TLSOptions[i-1].Restore(ctx); err != nil {
+				return errors.Annotatef(err, "An error occurred while restore CreateUserStmt.TLSOptions[%d]", i)
```
tangenta:
```suggestion
				return errors.Annotatef(err, "An error occurred while restore GrantStmt.TLSOptions[%d]", i)
```

ast/misc.go
```diff
@@ -1987,6 +1988,21 @@ func (n *GrantStmt) Restore(ctx *RestoreCtx) error {
 			return errors.Annotatef(err, "An error occurred while restore GrantStmt.Users[%d]", i)
 		}
 	}
+	if n.TLSOptions != nil {
+		tlsOptionLen := len(n.TLSOptions)
+		if tlsOptionLen != 0 {
+			ctx.WriteKeyWord(" REQUIRE ")
+		}
+		// Restore `TLSOptions` reversely to keep order the same with original sql
```
tangenta: To avoid this unnatural style, we can define the rule `RequireList` with `RequireList RequireListElement` instead of `RequireListElement RequireList`.

lysu:
comment address PTAL @tangenta  thx

ast/misc.go
```diff
@@ -1144,19 +1144,18 @@ func (n *CreateUserStmt) Restore(ctx *RestoreCtx) error {
...
+	for i := 0; i < tlsOptionLen; i++ {
+		if i != 0 {
 			ctx.WriteKeyWord(" AND ")
 		}
-		if err := n.TslOptions[i-1].Restore(ctx); err != nil {
-			return errors.Annotatef(err, "An error occurred while restore CreateUserStmt.TslOptions[%d]", i)
+		if err := n.TLSOptions[i].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore CreateUserStmt.TLSOptions[%d]", i)
 		}
 	}
```
kennytm: With normal iteration direction we could use a `for`/`range` loop instead of `for` loop.

```go
for i, option := range n.TLSOptions {
  if i != 0 {
    ctx.WriteKeyWord(" AND ")
  }
  if err := option.Restore(ctx); err != nil {
    return errors.Annotatef(err, "An error occurred while restore CreateUserStmt.TLSOptions[%d]", i)
  }
}
```

PR #667, opened 2019-12-08
parser: support index advisor

PiotrNewt:
support index advise syntax for [index advisor](https://github.com/pingcap/tidb/blob/master/docs/design/2019-11-05-index-advisor.md), ref [pingcap/tidb#13271](https://github.com/pingcap/tidb/pull/13271)

- Add a `IndexAdviseStmt` for index advisor
- Add new syntax to support index advisor
- Add test for new code of index advisor

### Check List 

Tests 

 - Unit test
 
Code changes

Side effects

- None

Related changes

- None


parser.y
```diff
@@ -399,6 +400,8 @@ import (
 	memory		"MEMORY"
 	merge		"MERGE"
 	minRows		"MIN_ROWS"
+	max_minutes		"MAX_MINUTES"
+	max_idxnum		"MAX_IDXNUM"
```
tangenta: Please format these lines.

parser.y
```diff
@@ -749,6 +754,7 @@ import (
 	GrantStmt			"Grant statement"
 	GrantRoleStmt			"Grant role statement"
 	InsertIntoStmt			"INSERT INTO statement"
+	IndexAdviseStmt		"INDEX ADVISE stetement"
```
tangenta: Ditto.

parser.y
```diff
@@ -10771,4 +10782,70 @@ LoadStatsStmt:
 		}
 	}
 
+/*******************************************************************************************/
+
+IndexAdviseStmt:
+	"INDEX" "ADVISE" LocalOpt "INFILE" stringLit MaxMinutes MaxIndexNum Lines
```
tangenta: Could you please put the `INDEX ADVISE` syntax here?

```
INDEX ADVISE
  [LOCAL]
  INFILE 'file_name'
  [MAX_MINUTES number]
  [MAX_IDXNUM
     [PER_TABLE number] 
     [PER_DB number]
  ]
  [LINES 
    [STARTING BY 'string'] 
    [TERMINATED BY 'string']
  ]
```

BTW, I think `MaxMinutesOpt` and `MaxIndexNumOpt` is better.

parser.y
```diff
@@ -10771,4 +10782,70 @@ LoadStatsStmt:
...
+	{
+		x := &ast.IndexAdviseStmt{
+			Path:	$5,
+		}
+		if $3 != nil {
+			x.IsLocal = true
+		}
+		if $6 != nil {
+			x.MaxMinutes = $6.(uint64)
+		}
```
tangenta: `if $6 != nil` is unnecessary.

parser.y
```diff
@@ -10771,4 +10782,70 @@ LoadStatsStmt:
...
+		if $8 != nil {
+			x.LinesInfo = $8.(*ast.LinesClause)
+		}
+		$$ = x
+	}
+
+MaxMinutes:
+	{
+		$$ = uint64(0)
+	}
```
tangenta: Can we take advantage of `types.UnspecifiedLength`? Thus we can distinguish 0 from an unspecified(default) number.

parser.y
```diff
@@ -10771,4 +10782,70 @@ LoadStatsStmt:
...
+		}
+		if $3 != nil {
+			x.PerDB = $3.(uint64)
+		}
+		$$ = x
+	}
+
+PerTable:
+	{
+		$$ = uint64(0)
```
tangenta: Consider `types.UnspecifiedLength`.

parser.y
```diff
@@ -10771,4 +10782,70 @@ LoadStatsStmt:
...
+	}
+|	"PER_TABLE" NUM
+	{
+		$$ = getUint64FromNUM($2)
+	}
+
+PerDB:
+	{
+		$$ = uint64(0)
```
tangenta: Consider `types.UnspecifiedLength`.

parser.y
PiotrNewt: Changing the name is a good idea!
And...did you mean adding a link to the document? Or write the syntax here directly as a comment in `parser.y`?

parser.y
PiotrNewt: There is indeed some ambiguity here, and using `types.UnspecifiedLength` is also a method to solve it.

But `types.UnspecifiedLength` is `-1`, which means the type of `MaxMinutes` needs to be changed from `uint` to `int`, it may cause unnecessary judgment.

Because in this scenario, **[-∞, 0]** is illegal input. If it is really necessary to specify a default value in *parser*, I think it is possible to define a constant similar to `types.UnspecifiedLength` to solve it, and the constant should be equal to `math.MaxUint64`.

Any suggestions? Or am I understanding it wrong?
(Same for the next few comments)

parser.y
tangenta: https://github.com/pingcap/parser/blob/3611952bf747dcbecc9197f85efe37654de5da5b/parser.y#L3779-L3792

Here is an example. 😄

parser.y
tangenta: Sure, you can create a similar constant using math.MaxUint64.

PiotrNewt: Any suggestions for this one? 🤔

tangenta: IMHO, defining a constant named `UnspecifiedSize` is better.

PiotrNewt: So should I also define it in the same position as `types.UnspecifiedLength`? In the types package

tangenta: How about putting it to `ast/util.go`?

parser.y
```diff
@@ -1146,6 +1152,10 @@ import (
 	EnforcedOrNotOrNotNullOpt	"{[ENFORCED|NOT ENFORCED|NOT NULL]}"
 	Match			"[MATCH FULL | MATCH PARTIAL | MATCH SIMPLE]"
 	MatchOpt		"optional MATCH clause"
+	MaxMinutesOpt	"MAX_MINUTES num(int)"
+	MaxIndexNumOpt	"MAX_IDXNUM clause"
+	PerTable 		"Max index number PER_TABLE"
+	PerDB			"Max index number PER_DB"
```
tangenta: Please format these lines.

PiotrNewt: Oh sorry 
But what tools are helpful? I always seem to be puzzled by `tab` in `parser.y`😂

tangenta: We are planning to develop such a tool. See #670.

The `parser.go` is still in conflict, you can run `make parser` to resolve it.

PR #674, opened 2019-12-12
*: support invisible database / table syntax

zimulala:
Support the invisible database/table. e.g.
```
CREATE TABLE tbl_name (create_definition,...) INVISIBLE
ALTER TABLE tbl_name VISIBLE;
```

It's used for pingcap/br/issues/87.

Add a table_options and a database_options when doing `create/alter database/table`.
The option is `{VISIBLE | INVISIBLE}`.

### Check List 

Tests 

 - Unit test


kennytm:
What is an invisible table? What is its the effect?

tiancaiamao:
What problem does it solve?
Is this MySQL compatible grammar, or could it be added to the comment part?

tangenta:
@zimulala I think the `VISIBILITY` in `CREATE TABLE `a` (`a` INT,`b` INT) VISIBLE` is not needed because it is the default behavior.

kennytm:
Since BR operates on [TiDB's Go API](https://github.com/pingcap/br/blob/8c017e872b4dfff3cfcdc11b4ea3facea5c95e48/pkg/restore/db.go#L58-L90), we could manipulate the state without introducing a new syntax. With this, do we need to introduce the `CREATE TABLE … INVISIBLE` syntax? Like, does INVISIBLE table/schema have any use other than BR/Lightning restore, both of them can use TiDB interals?

One worry I've got is that INVISIBLE table is subtly different from [MariaDB's INVISIBLE columns](https://mariadb.com/kb/en/library/invisible-columns/) (hidden from `SELECT *` and `INSERT INTO`, but you could still explicitly use the columns in queries and DMLs).

For instance if we add a `CreateTableWithModel` [DDL method](https://github.com/pingcap/tidb/blob/cc635dbc9067c60ffbc03d505749483de31d8fed/ddl/ddl.go#L230-L234) instead, we could just set the `table.Schema`'s State to DeleteOnly from the beginning. This API is also more natural for BR, allowing them to cut the `model.TableInfo` → `ConstructResultOfShowCreateTable` to a string → parse into `ast.CreateStmt` → reconstruct `model.TableInfo` detour.

```go
type DDL interface { 
    CreateTable(ctx sessionctx.Context, stmt *ast.CreateTableStmt) error
    CreateTableWithModel(ctx sessionctx.Context, ti *model.TableInfo) error
    ...
}
```

(DM can also be benefited by need this API)

zimulala:
@kennytm Good idea! Let me discuss your proposal with the demand side.

zimulala:
This feature is not currently required

PR #675, opened 2019-12-12
mysql: add foreign key error

crazycs520:
Add error msg

### DIFF
--- a/mysql/errcode.go
+++ b/mysql/errcode.go
@@ -939,6 +939,7 @@ const (
 	ErrSpatialFunctionalIndex                                       = 3760
 	ErrWrongKeyColumnFunctionalIndex                                = 3761
 	ErrFunctionalIndexOnField                                       = 3762
+	ErrReferencedForeignKey                                         = 3780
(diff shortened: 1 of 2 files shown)


mysql/errcode.go
```diff
@@ -939,6 +939,7 @@ const (
 	ErrSpatialFunctionalIndex                                       = 3760
 	ErrWrongKeyColumnFunctionalIndex                                = 3761
 	ErrFunctionalIndexOnField                                       = 3762
+	ErrReferencedForeignKey                                         = 3780
```
tangenta: According to the MySQL [document](https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html),
> Error number: 3780; Symbol: ER_FK_INCOMPATIBLE_COLUMNS; SQLSTATE: HY000
Message: Referencing column '%s' and referenced column '%s' in foreign key constraint '%s' are incompatible.

The error name should be `ErrFKIncompatibleColumns`.

crazycs520: great, done.

tiancaiamao:
LGTM

kennytm:
(CLA is borked)

PR #676, opened 2019-12-12
remove testleak

tangenta:
This is the first step to remove the dependency on TiDB. #677 

- remove testleak

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes
N/A

Side effects
N/A

Related changes

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -15,7 +15,6 @@ package auth
 
 import (
 	. "github.com/pingcap/check"
-	"github.com/pingcap/tidb/util/testleak"
 )
 
 var _ = Suite(&testAuthSuite{})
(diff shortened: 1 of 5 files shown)


test.sh
```diff
@@ -1,11 +1,15 @@
...
-	  GO111MODULE=on go test -p 1 -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... ./...
-} || {
-	  mv go.mod go.mod1
-	  mv go.sum go.sum1
-}
+#!/usr/bin/env bash
 
-mv go.mod go.mod1
-mv go.sum go.sum1
+exitCode=0
```
kennytm: Please just use `set -e`. 

In the current implementation, if `go test` failed the shell script would still `exit 0`.

tangenta: But the command `mv go.mod go.mod1 && mv go.sum go.sum1` should be executed regardless of the test succeeds or fails.

kennytm: @tangenta 

```sh
mv go.mod go.mod1 && mv go.sum go.sum1 || true
```

kennytm: @tangenta You could also use `trap EXIT`, or retain the previous code structure.

tiancaiamao:
I think we can remove the leaktest instead.
The parser doesn't start goroutines. @tangenta

tiancaiamao:
LGTM

PR #679, opened 2019-12-14
remove dependency on TiDB

### LINKED ISSUE #677: remove dependency on TiDB
Currently, the tests in parser are depended on the packages in TiDB like `testleak`, `types` and `parser_driver`. On the other hand, TiDB also depends on the parser. This circular dependency has caused a lot of inconveniences, including
- the code in an old-version TiDB must be imported before using the parser
- `go.mod` and `go.sum` must be named as `go.mod1` and `go.sum1`

In order to solve circular dependency, we can copy the package `testleak` from TiDB to parser and mock a `parser_driver` for tests in parser.

tangenta:
Fix #677.

Copy TiDB's parser_driver and remove the unused code. This forms a minimal mock parser_driver without breaking any existing tests.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N/A

Side effects

N/A

Related changes

N/A


parser_example_test.go
```diff
@@ -16,24 +16,20 @@ package parser_test
...
 	"fmt"
 	"github.com/pingcap/parser"
-	_ "github.com/pingcap/tidb/types/parser_driver"
 )
 
 // This example show how to parse a text sql into ast.
 func Example_parseSQL() {
 
-	// 0. make sure import parser_driver implemented by TiDB(user also can implement own driver by self).
-	// and add `import _ "github.com/pingcap/tidb/types/parser_driver"` in the head of file.
```
kennytm: I don't think we want to recommend `test_driver` for general use. Perhaps move this entire file to `README.md` and keep the content unchanged.

tiancaiamao:
LGTM

PR #680, opened 2019-12-15
parser : support drop sequence syntax.

AilinKid:
TiDB are compatible with MySQL's protocol, we should be also compatible with SEQUENCE implementation of MariaDB cause MySQL is absent from it.

*      DROP [TEMPORARY] SEQUENCE [IF EXISTS] sequence_name [, sequence_name] ...

Linked with sequence project, this PR try to support drop sequence syntax.

- [x] ```SHOW CREATE SEQUENCE sequence_name;```
- [ ] ```ALTER SEQUENCE [IF EXISTS] sequence_name
[ INCREMENT [ BY | = ] increment ]
[ MINVALUE [=] minvalue | NO MINVALUE | NOMINVALUE ]
[ MAXVALUE [=] maxvalue | NO MAXVALUE | NOMAXVALUE ]
[ START [ WITH | = ] start ] [ CACHE [=] cache ] [ [ NO ] CYCLE ]
[ RESTART [[WITH | =] restart];```
- [x] ```DROP [TEMPORARY] SEQUENCE [IF EXISTS]
    sequence_name [, sequence_name] ...```
- [ ] ```Sequence Function (need discussion)```

### Check List 

Tests 

 - Unit test
 - Integration test


ast/ddl.go
```diff
@@ -1089,6 +1089,54 @@ func (n *DropTableStmt) Accept(v Visitor) (Node, bool) {
...
+	}
+	if n.IfExists {
+		ctx.WriteKeyWord("IF EXISTS ")
+	}
+	for i, sequence := range n.Sequences {
+		if i != 0 {
+			ctx.WritePlain(", ")
+		}
+		if err := sequence.Restore(ctx); err != nil {
+			return errors.Annotate(err, "An error occurred while restore DropSequenceStmt.Sequences "+string(i))
```
```suggestion
			return errors.Annotatef(err, "An error occurred while restore DropSequenceStmt.Sequences[%d]", i)
```

AilinKid: Addressed

PR #684, opened 2019-12-16
parser: add explain format for optimizer trace

winoros:

Add one more new explain format to enable optimizer trace.

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -422,6 +422,7 @@ import (
 	nulls		"NULLS"
 	offset		"OFFSET"
 	only		"ONLY"
+	opttrace        "OPT_TRACE"
```
tangenta: To pass the consistency check, we can add `opttrace` to the tokenMap in `misc.go`.

tangenta (commented):
LGTM

misc.go
```diff
@@ -482,6 +482,7 @@ var tokenMap = map[string]int{
 	"OPTIMIZE":                 optimize,
 	"OPTION":                   option,
 	"OPTIONALLY":               optionally,
+	"OPT_TRACE":                opttrace,
```
lamxTyler: Seems that we don't need it? Check https://github.com/pingcap/parser/pull/654/files

ast/misc.go
```diff
@@ -63,11 +63,12 @@ const (
...
 	// Valid formats for explain statement.
-	ExplainFormatROW  = "row"
-	ExplainFormatDOT  = "dot"
-	ExplainFormatHint = "hint"
-	PumpType          = "PUMP"
-	DrainerType       = "DRAINER"
+	ExplainFormatROW   = "row"
+	ExplainFormatDOT   = "dot"
+	ExplainFormatHint  = "hint"
+	ExplainFormatTrace = "opt_trace"
```
eurekaka: Simply `trace` is enough?

parser.y
```diff
@@ -3958,6 +3959,8 @@ ExplainStmt:
 			Stmt:	$2,
 			Format: "row",
 		}
+		startOffset := parser.startOffset(&yyS[yypt])
+		$2.SetText(string(parser.src[startOffset:]))
```
eurekaka: Could you give an example to explain the purpose of these 2 lines?

parser.y
```diff
@@ -3979,6 +3982,8 @@ ExplainStmt:
 			Stmt:	$5,
 			Format: $4,
 		}
+		startOffset := parser.startOffset(&yyS[yypt])
+		$5.SetText(string(parser.src[startOffset:]))
 	}
 |	ExplainSym "FORMAT" "=" ExplainFormatType "FOR" "CONNECTION" NUM
```
eurekaka: We cannot get the optimizer trace from `explain for connection` since the plan may have already been generated?

ast/misc.go
zz-jason: +1

PR #691, opened 2019-12-18
parser : Add model & errcode & errname for supporting sequence.

AilinKid:
Add model & errcode & errname for supporting sequence.

### Check List 

Tests 

 - Unit test
 - Integration test


model/model.go
```diff
@@ -504,6 +507,11 @@ func (t *TableInfo) IsView() bool {
 	return t.View != nil
 }
 
+// IsSequence checks if tableinfo is a sequence
```
zimulala:
```suggestion
// IsSequence checks if TableInfo is a sequence.
```

model/model.go
```diff
@@ -577,6 +585,33 @@ type ViewInfo struct {
...
+)
+
+// SequenceInfo provide meta data describing a DB sequence.
+type SequenceInfo struct {
+	Start      int64  `json:"sequence_start"`
+	Cache      bool   `json:"sequence_cache"`
+	Order      bool   `json:"sequence_order"`
+	Cycle      bool   `json:"sequence_cycle"`
+	MinValue   int64  `json:"sequence_minvalue"`
+	MaxValue   int64  `json:"sequence_maxvalue"`
```
zimulala:
```suggestion
	MaxValue   int64  `json:"sequence_max_value"`
```

PR #694, opened 2019-12-20
parser: fix panic when MEMORY_QUOTA overflows int64

eurekaka:

```
mysql> select /*+ memory_quota(18446744073709551612 MB) */ 1;
ERROR 2013 (HY000): Lost connection to MySQL server during query
```

Check integer overflow when parsing MEMORY_QUOTA hint.

### Check List 

Tests 

 - Unit test

Code changes

N/A

Side effects

N/A

Related changes


eurekaka:
@tangenta thanks for the comments. `MemoryQuota` uses `-1` to represent "no limit", so I think it should better to be `int64`. If the upper-bound is not checked when converting `MB / GB` to `Byte`, the panic would also happen in https://github.com/pingcap/parser/blob/master/parser.y#L7073 ?

tiancaiamao:
LGTM

PR #701, opened 2019-12-26
add yacc file formatter

### LINKED ISSUE #670: format tool for yacc file
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, due to the lack of yacc file format tool, almost all PRs have format issues in `parser.y` when they are first submitted. In order to improve efficiency, we should develop a format tool.

**Describe the feature you'd like:**

`make fmt` should format `parser.y` file.

tangenta:
Fix #670.

Use goyacc's parser to generate code from AST.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


parser.y
```diff
@@ -2178,44 +2181,42 @@ SplitSyntaxOption:
 		}
 	}
 
-/*******************************************************************************************/
```
kennytm: Should we keep this?

parser.y
```diff
@@ -3164,22 +3176,6 @@ DatabaseOptionList:
...
- *      CREATE TABLE Persons
- *      (
- *          P_Id int NOT NULL,
- *          LastName varchar(255) NOT NULL,
- *          FirstName varchar(255),
- *          Address varchar(255),
- *          City varchar(255),
- *          PRIMARY KEY (P_Id)
- *      )
- *******************************************************************/
```
kennytm: 😕

Makefile
```diff
@@ -28,12 +28,14 @@ parser: bin/goyacc
...
-	GO111MODULE=on go build -o bin/goyacc goyacc/main.go
+bin/goyacc: goyacc/main.go goyacc/format_yacc.go
+	GO111MODULE=on go build -o bin/goyacc goyacc/main.go goyacc/format_yacc.go
 
-fmt:
+fmt: bin/goyacc
 	@echo "gofmt (simplify)"
 	@gofmt -s -l -w . 2>&1 | awk '{print} END{if(NR>0) {exit 1}}'
+	@bin/goyacc -fmt -fmtout parser_golden.y parser.y
+	@(git diff --no-index --quiet parser.y parser_golden.y && rm parser_golden.y) || (mv parser_golden.y parser.y && >&2 echo "formatted parser.y" && exit 1)
```
kennytm: I suggest using `--exit-code` instead of `--quiet`

goyacc/format_yacc.go
```diff
@@ -0,0 +1,581 @@
...
+		switch def.Case {
+		case StartIdentifierCase:
+			err = handleStart(formatter, def)
+		case UnionDefinitionCase:
+			err = handleUnion(formatter, def)
+		case LCURLRCURLCase:
+			err = handleProlog(formatter, def)
+		case ReservedWordTagNameListCase:
+			err = handleReservedWordTagNameList(formatter, def)
+		}
```
kennytm: how about `ReservedWordTagCase`?

goyacc/format_yacc.go
```diff
@@ -0,0 +1,581 @@
...
+	var splitter, beforeComment string
+	switch layout {
+	case spanStringLayout:
+		splitter, beforeComment = " ", ""
+	case divStringLayout:
+		splitter, beforeComment = "\n", ""
+	case divNewLineStringLayout:
+		splitter, beforeComment = "\n", "\n"
+	default:
+		panic(errors.WithStack(errors.Errorf("unsupported stringLayout: %v", layout)))
```
kennytm:
```suggestion
		panic(errors.Errorf("unsupported stringLayout: %v", layout))
```

Errorf should contain the stack already.

goyacc/format_yacc.go
```diff
@@ -0,0 +1,581 @@
...
+func printSingleName(f format.Formatter, name *parser.Name, maxCharLength int) error {
+	if hasComments(name) {
+		_, err := f.Format("\n%s\n", strings.Join(name.Token.Comments, "\n"))
+		if err != nil {
+			return err
+		}
+	}
+	if name.LiteralStringOpt != nil && name.LiteralStringOpt.Token != nil {
+		strLit := fmt.Sprintf(" %s", name.LiteralStringOpt.Token.Val)
+		_, err := f.Format("%-*s%s\n", maxCharLength, name.Token.Val, strLit)
```
kennytm: This is the same as?

```go
_, err := f.Format("%-*s %s\n", maxCharLength, name.Token.Val, name.LiteralStringOpt.Token.Val)
```

goyacc/format_yacc.go
```diff
@@ -0,0 +1,581 @@
...
+	for ri := rule.RuleItemList; ri != nil; ri = ri.RuleItemList {
+		switch ruleItemType(ri.Case) {
+		case identRuleItemType, strLiteralRuleItemType:
+			term := fmt.Sprintf(" %s", ri.Token.Val)
+			if ri == firstRuleItem {
+				term = term[1:]
+			}
+			cmt := getTokenComment(ri.Token, divStringLayout)
+
+			if _, err := f.Format(escapePercent(cmt)); err != nil {
```
kennytm: I think `f.Format("%s", cmt)` should be enough. If we use `/* ... */` instead of `// ...` it should handle multi-line comments fine? (ditto for everything involving `escapePercent()`)

tangenta: `cmt` contains characters like `\n`, which is used by `format.Formatter` (when `%i` is enabled) to determine whether it should print an indention or not.

To use the `%i`/`%u` mechanism, the string contains `\n` must be passed into `Format` as the first argument. So it's different between
```
f.Format("%s", cmt)
```
and 
```
f.Format(cmt)
```

parser.y
tangenta: The comments without any code in its next line __are ignored__ by goyacc/parser, I can't find them in AST  :joy:

BTW, I think it's acceptable to remove these meaningless comments.

goyacc/format_yacc.go
kennytm: @tangenta I understand the difference between `Format(s)` vs `Format("%s", s)`, what I'm saying is that perhaps we could avoid `Format(s)`.

parser.y
```diff
@@ -3839,8 +3842,10 @@ DropTableStmt:
 	}
 
 OptTemporary:
-	  /* empty */ { $$ = false; }
-	| "TEMPORARY" 
+	/* empty */
+	{	$$ = false 
+	}
```
kennytm: 🤔 is this acceptable?

goyacc/format_yacc.go
```diff
@@ -0,0 +1,579 @@
...
+	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
+
+	parser "github.com/cznic/parser/yacc"
+	"github.com/cznic/strutil"
+	"github.com/pingcap/errors"
+	"github.com/pingcap/parser/format"
+	gofmt "go/format"
```
djshow832: Move this import up.

tiancaiamao:
LGTM

PR #703, opened 2019-12-27
Remove abandon command in Makefile

Deardrops:

After merging #679, `cpmod` in Makefile is abandon, it should be removed.

Removing `cpmod` in Makefile.

### Check List 

Tests 

 - None

Code changes

 - None

Side effects

 - None

Related changes

 - None


reafans:
LGTM

PR #706, opened 2019-12-31
distinguish 'convert to' keyword in table option of charset

tangenta:
Fix the parser part in pingcap/tidb#14284.

Related to pingcap/tidb#14304.

According to MySQL [document](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-character-set),  the statement
```
ALTER TABLE tbl_name CONVERT TO CHARACTER SET charset_name;
```
> also changes the collation of all character columns. 

However, without the `CONVERT TO` keyword, both charset and collation of the columns should remain unchanged.

So the parser should recognize if there is the 'CONVERT TO' keyword in a table option.

Utilize the field `UintValue` of TableOption to record whether a charset table option has the 'convert to' keyword before it.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


ast/ddl.go
```diff
@@ -1823,8 +1828,15 @@ func (n *TableOption) Restore(ctx *RestoreCtx) error {
...
-		ctx.WritePlain("= ")
+		if n.UintValue == TableOptionCharsetWithConvertTo {
+			ctx.WriteKeyWord("CONVERT TO ")
+		} else {
+			ctx.WriteKeyWord("DEFAULT ")
+		}
+		ctx.WriteKeyWord("CHARACTER SET ")
+		if n.UintValue == TableOptionCharsetWithoutConvertTo {
+			ctx.WriteKeyWord("= ")
+		}
```
kennytm: Perhaps make the equal sign unnecessary

ast/ddl_test.go
```diff
@@ -453,8 +453,8 @@ func (ts *testDDLSuite) TestAlterTableSpecRestore(c *C) {
 		{"ROW_FORMAT = tokudb_uncompressed", "ROW_FORMAT = TOKUDB_UNCOMPRESSED"},
 		{"shard_row_id_bits 1", "SHARD_ROW_ID_BITS = 1"},
 		{"shard_row_id_bits = 1", "SHARD_ROW_ID_BITS = 1"},
-		{"CONVERT TO CHARACTER SET utf8", "DEFAULT CHARACTER SET = UTF8"},
-		{"CONVERT TO CHARSET utf8", "DEFAULT CHARACTER SET = UTF8"},
+		{"CONVERT TO CHARACTER SET utf8", "CONVERT TO CHARACTER SET UTF8"},
+		{"CONVERT TO CHARSET utf8", "CONVERT TO CHARACTER SET UTF8"},
```
kennytm: Please include some test cases involving DEFAULT CHARACTER SET.

ast/ddl.go
tangenta: If we remove this equal sign, we should also remove the equal sign in `COLLATE = XXX`, and update a lot of tests.

kennytm: okay 😞

bb7133: I should say the syntax of MySQL is really weird.

PR #711, opened 2020-01-06
Rewrite special comment parser

kennytm:

Rewrite the special comment parser (`/*! ... */` and `/*+ ... */`) to better match MySQL's behavior. 

This is the first step to solve https://github.com/pingcap/tidb/issues/14030#issuecomment-567791327.

* Removed `specialCommentScanner`. Instead:
    1. When encountering `/*!` or `/*!12345` or `/*T!` or `/*T!012345`, drop this substring and continue parsing as if outside a comment. Use a flag to indicate we should skip the stray `*/`.
    2. When encountering `/*+`, read the entire comment as a string token. This token will be parsed into `[]*ast.TableOptimizerHint` using the new *hint parser* on-demand. All errors coming from that parsers are demoted to warnings.

* The `/*+ ... */` is only parsed after keywords SELECT, INSERT, REPLACE, UPDATE, DELETE, CREATE and PARTITION, otherwise it is treated as an ordinary comment. Therefore, some SQL previously failing like `SELECT 1 FROM /*+ hash_agg() */ t1` now becomes successful and ignoring the hint. 

    * Unlike MySQL, however, statements like `SELECT 1 FOR UPDATE /*+ max_execution_time(60) */` would still produce an error. We could fix it later?

* And we now have two parsers, the main `parser.y`, and the optimizer-hint specific `hintparser.y`

    1. All hint-related rules from deleted from `parser.y`. This also deletes the related TiDB-specific keywords.
    2. goyacc is modified to support generating two parsers. Also make goyacc to exit with error on conflict, so we don't need to run the parser twice on the same .y file.
    3. In the new hint parser, all optimizer hints from MySQL 5.7+ are recognized for completeness (though all of them are ignored with warnings).

* The structure `ast.TableOptimizerHint` is *not yet changed*, so we are still API-compatible.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


hintparserimpl.go
```diff
@@ -0,0 +1,147 @@
...
+	yyVAL  *yyhintSymType
+}
+
+// ParseHint parses an optimizer hint (the interior of `/*+ ... */`).
+func ParseHint(input string, sqlMode mysql.SQLMode) ([]*ast.TableOptimizerHint, []error) {
+	// yyhintDebug = 2
+
+	hp := hintParser{cache: make([]yyhintSymType, 50)}
+	hp.lexer.reset(input)
+	hp.lexer.SetSQLMode(sqlMode)
```
tangenta: Can we reuse the struct `hintParser` instead of allocating it every time `ParseHint` is called?

lexer.go
```diff
@@ -388,82 +334,98 @@ func startWithDash(s *Scanner) (tok int, pos Pos, lit string) {
...
+		s.r.inc()
 		return s.scan()
 	}
-	tok = int('/')
-	return
+	// otherwise it is just a normal star.
+	return '*', pos, "*"
 }
 
 func sqlOffsetInComment(comment string) int {
```
tangenta: Please remove this unused function.

kennytm:
PTAL @tangenta @tiancaiamao cc @lonng 

The integration test is failing because the warnings produced by invalid hint is changed.

tiancaiamao:
LGTM
Good job! @kennytm

tiancaiamao:
CI failed here https://github.com/pingcap/tidb/blob/master/expression/integration_test.go#L4407

PR #713, opened 2020-01-09
parser: support sequence function & next value for

AilinKid:
Support sequence function `nextval`, `lastval` and `setval`.
Support `next value for sequence` syntax. 

Add function registration in `functions.go. `

### Check List 

Tests 

 - Integration test

Related changes


ast/functions_test.go
```diff
@@ -73,6 +73,9 @@ func (ts *testFunctionsSuite) TestFuncCallExprRestore(c *C) {
 		{"MASTER_POS_WAIT(@log_name, @log_pos, @timeout, @channel_name)", "MASTER_POS_WAIT(@`log_name`, @`log_pos`, @`timeout`, @`channel_name`)"},
 		{"JSON_TYPE('[123]')", "JSON_TYPE('[123]')"},
 		{"bit_and(all c1)", "BIT_AND(`c1`)"},
+		{"nextval('seq')", "NEXTVAL('seq')"},
+		{"lastval('seq')", "LASTVAL('seq')"},
+		{"setval('seq')", "SETVAL('seq')"},
```
kennytm: According to the RFC the input to these functions should be the sequence name, not a string.

```go
 		{"nextval(seq)", "NEXTVAL(`seq`)"}, 
```

AilinKid: Addressed, PTAL @kennytm

ast/functions_test.go
```diff
@@ -200,3 +203,15 @@ func (ts *testDMLSuite) TestWindowFuncExprRestore(c *C) {
...
+func (ts *testDMLSuite) TestNextValueForSequence(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"next value for seq", "NEXTVAL(`seq`)"},
+		{"next value for sequence", "NEXTVAL(`sequence`)"},
+		{"NeXt vAluE for seQuEncE2", "NEXTVAL(`seQuEncE2`)"},
+	}
+	extractNodeFunc := func(node Node) Node {
+		return node.(*SelectStmt).Fields.Fields[0].Expr
+	}
+	RunNodeRestoreTestWithoutCompareAST(c, testCases, "select %s", extractNodeFunc)
```
kennytm: Both `NEXT VALUE FOR seq ` and `nextval(seq)` should produce the same AST. This means `nextval` and friends should be special-cased to only accept ObjectNameExpr (the `seq` in `nextval(seq)` should never mean a column named `seq`)

AilinKid: Exactly right~ I see, got it!

AilinKid: Exactly right, got it~

ast/functions_test.go
```diff
@@ -76,6 +76,9 @@ func (ts *testFunctionsSuite) TestFuncCallExprRestore(c *C) {
 		{"nextval(seq)", "NEXTVAL(`seq`)"},
 		{"lastval(seq)", "LASTVAL(`seq`)"},
 		{"setval(seq, 100)", "SETVAL(`seq`, 100)"},
+		{"next value for seq", "NEXTVAL(`seq`)"},
+		{"next value for sequence", "NEXTVAL(`sequence`)"},
+		{"NeXt vAluE for seQuEncE2", "NEXTVAL(`seQuEncE2`)"},
```
kennytm: Please add test cases for the `` `db`.`tbl` `` form.

ast/expressions.go
```diff
@@ -443,6 +444,47 @@ func (n *CompareSubqueryExpr) Accept(v Visitor) (Node, bool) {
...
+
+// Format the ExprNode into a Writer.
+func (n *ObjectNameExpr) Format(w io.Writer) {
+	result := n.Name.Name.L
+	if n.Name.Schema.L != "" {
+		result = n.Name.Schema.L + "." + result
+	}
+	name := strings.Replace(result, ".", "`.`", -1)
+	fmt.Fprintf(w, "`%s`", name)
+}
```
tangenta: Could we to avoid using strings.Replace()?
```go
dbName, tbName := n.Name.Schema.L, n.Name.Name.L
if dbName == "" {
	fmt.Fprintf(w, "`%s`", tbName)
} else {
	fmt.Fprintf(w, "`%s`.`%s`", dbName, tbName)
}
```

ast/expressions.go
```diff
@@ -443,6 +444,47 @@ func (n *CompareSubqueryExpr) Accept(v Visitor) (Node, bool) {
 	return v.Leave(n)
 }
 
+// ObjectNameExpr represents a table-level name expression, such as sequence/table/view etc.
+type ObjectNameExpr struct {
```
tangenta: I think renaming to `TableNameExpr` is better.

AilinKid: Fine, addressed.

ast/expressions.go
AilinKid: OK, addressed.

ast/functions_test.go
AilinKid: Make sense, addressed.

AilinKid:
> /build

Is it useful for parser repo?

PR #715, opened 2020-01-15
*: remove TiDB only terror from parser repo

lysu:

see more details in https://github.com/pingcap/tidb/pull/14482

WIP

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/715)


mysql/const.go
```diff
@@ -657,21 +653,6 @@ func FormatSQLModeStr(s string) string {
...
-	for i, length := 0, len(strs); i < length; i++ {
-		mode, ok := Str2SQLMode[strs[i]]
-		if !ok && strs[i] != "" {
-			return sqlMode, newInvalidModeErr(strs[i])
-		}
-		sqlMode = sqlMode | mode
-	}
-	return sqlMode, nil
-}
-
```
kennytm: This function is used in tools and should not be removed.

lysu: thanks....I forgot tools use this package...let me take a look...

tiancaiamao: All exported functions and global variable are difficult to change

lysu:
will take more smaller step with #734, so close this one

Issue #600, opened 2019-10-31
Fix TiDB behavior: table encryption


tangenta:
Index issue: #598 

### SQL example
```sql
CREATE TABLE t (a INT) ENCRYPTION = {'Y' | 'N'}
```

### Current implementation
`TableOptionEncryption` is ignored by TiDB.

### Expected implementation
Throw a warning only when encryption equals to ‘Y’.

Issue #599, opened 2019-10-31
Fix TiDB behavior: database encryption


tangenta:
Index issue: #598 

### SQL example
```sql
CREATE DATABASE test DEFAULT ENCRYPTION = 'N';
```

### Current implementation
`DatabaseOptionEncryption` is ignored by TiDB.

### Expected implementation
- Ignore it when encryption equals to ‘n’.
- Throw a warning when encryption equals to ‘y’.
- Report an error when encryption equals to other values. 

Please refer to corresponding MySQL errors.

PR #717, opened 2020-01-17
Parser: Database and Table encryption (#599 #600)

zhaox1n:

Support database encryption [#599](<https://github.com/pingcap/parser/issues/599>) [#600](<https://github.com/pingcap/parser/issues/600>) 

### Bad SQL Case

```sql
create database t default encryption = 'X'
```

### Check List 

Tests 

 - Unit test


tangenta:
@zhaox1n Thanks for your contribution! We are not supposed to throw errors in parser, because there are several projects other than TiDB depending on it. We can add this check to

https://github.com/pingcap/tidb/blob/a7076a52635c59e40a8bf771ff27a97f5da08370/planner/core/preprocess.go#L363

If you have more questions, feel free to ask here or contact me on slack :)
> Slack: channel #sig-parser in the [tidbcommunity](https://pingcap.com/tidbslack) slack workspace.

zhaox1n:
@tangenta Thanks for your reply. The way to solve this issue is just to simulate the mysql-server implementation. If this issue should be fixed in tidb, obviously this issue should not be displayed in the parser module. In my opinion, the parser module is used to be compatible with mysql lexical and grammatical, and then output AST. Is this correct?

parser.y
```diff
@@ -3123,6 +3123,13 @@ DatabaseOption:
 	}
 |	DefaultKwdOpt "ENCRYPTION" EqOpt stringLit
 	{
+		if "Y" == strings.ToUpper($4) {
+			yylex.AppendError(yylex.Errorf("The ENCRYPTION clause is parsed but ignored by all storage engines."))
+			parser.lastErrorAsWarn()
+		} else if "N" != strings.ToUpper($4) {
+			yylex.AppendError(ErrWrongValue.GenWithStackByArgs("argument (should be Y or N)", $4))
+			return 1
+		}
```
kennytm: Better use a `switch`/`case` here (maybe get rid of `strings.ToUpper` too), or create a dedicated rule for `'Y'`/`'y'`/`'N'`/`'n'` in place of `stringLit` (since the same also applies to TableOption).

```go
switch $4 {
case "Y", "y":
    yylex.AppendError(...)
    parser.lastErrorAsWarn()
case "N", "n":
    break
default:
    yylex.AppendError(...)
    return 1
}
```

zhaox1n: I have modified the implementation of the code.

parser.y
```diff
@@ -1131,6 +1131,7 @@ import (
 	MaxIndexNumOpt                         "MAX_IDXNUM clause"
 	PerTable                               "Max index number PER_TABLE"
 	PerDB                                  "Max index number PER_DB"
+	EncryptionOpt                          "Encryption option 'Y' or 'N'"
```
tangenta: We can move `EncryptionOpt` declaration from `%type	<item>` to `%type	<ident>` to avoid unnecessary type conversion like `$4.(string)`.

``` 
%union {
	offset int // offset
	item interface{}
	ident string
	expr ast.ExprNode
	statement ast.StmtNode
}
```

PR #726, opened 2020-01-20
add json_objectagg's null key errcode

hg2990656:
MySQL document: https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html#error_er_json_document_null_key

### DIFF
--- a/mysql/errcode.go
+++ b/mysql/errcode.go
@@ -900,7 +900,7 @@ const (
 	ErrInvalidJSONPathWildcard                                      = 3149
 	ErrInvalidJSONContainsPathType                                  = 3150
 	ErrJSONUsedAsKey                                                = 3152
-	ErrJSONKeyNULL                                                  = 3158
+	ErrJSONDocumentNULLKey                                          = 3158
 	ErrBadUser                                                      = 3162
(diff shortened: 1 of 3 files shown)


hg2990656:
@zz-jason 
In the [#11154](https://github.com/pingcap/tidb/pull/11154), I need to add two error code, one is key null error and the other is type support error.But I found there is already a errorcode ErrUnsupportedType = 8108 in TiDB self-defined errors.So I am not sure whether should I add another error code just for json_objectagg such as ErrorUnspportSecondArugumentType or I can use now error code ErrUnsupportedType for this error.

hg2990656:
@tangenta 
Whether should I use ErrUnsupportedType = 8108 in TiDB self-defined to print error directly instead of adding another ErrUnsupportedSecondArugumentType for json_objectagg because the value type is not supported completely?

zz-jason:
> @tangenta
> Whether should I use ErrUnsupportedType = 8108 in TiDB self-defined to print error directly instead of adding another ErrUnsupportedSecondArugumentType for json_objectagg because the value type is not supported completely?

I prefer adding another error code for it, actually we are adding error code for each TiDB internal error.

PR #727, opened 2020-01-20
add json_objetagg's unsupported second argument type errorcode

hg2990656:
In [#11154](https://github.com/pingcap/tidb/pull/11154), when json_objectagg append json which contains string key and any kinds of value to the result chunk, the value type can be any types but appendBinary only support few types, so many types will not supported now in json_objectagg result and I add the error code for this.

### DIFF
--- a/mysql/errcode.go
+++ b/mysql/errcode.go
@@ -1036,6 +1036,7 @@ const (
 	ErrWarnMemoryQuotaOverflow             = 8063
 	ErrWarnOptimizerHintParseError         = 8064
 	ErrWarnOptimizerHintInvalidInteger     = 8065
+	ErrUnsupportedSecondArgumentType       = 8066
 
 	// Error codes used by TiDB ddl package
 	ErrUnsupportedDDLOperation        = 8200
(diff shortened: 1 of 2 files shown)


mysql/errname.go
```diff
@@ -1063,6 +1063,7 @@ var MySQLErrName = map[uint16]string{
 	ErrPrivilegeCheckFail:                  "privilege check fail", // this error message should begin lowercased to be compatible with the test
 	ErrInvalidWildCard:                     "Wildcard fields without any table name appears in wrong place",
 	ErrMixOfGroupFuncAndFieldsIncompatible: "In aggregated query without GROUP BY, expression #%d of SELECT list contains nonaggregated column '%s'; this is incompatible with sql_mode=only_full_group_by",
+	ErrUnsupportedSecondArgumentType:       "The second argument type is not supported now, it will be supported later",
```
zz-jason:
```suggestion
	ErrUnsupportedSecondArgumentType:       "JSON_OBJECTAGG: unsupported second argument type %v",
```

mysql/errcode.go
```diff
@@ -1036,6 +1036,7 @@ const (
 	ErrWarnMemoryQuotaOverflow             = 8063
 	ErrWarnOptimizerHintParseError         = 8064
 	ErrWarnOptimizerHintInvalidInteger     = 8065
+	ErrUnsupportedSecondArgumentType       = 8066
```
tangenta: Is this error only applicable in `JSON_OBJECTAGG`?

hg2990656: Yes

Issue #714, opened 2020-01-09
Staticcheck report


mahjonp:
## Staticcheck Report

1. What did you do?

```bash
$ go get honnef.co/go/tools/cmd/staticcheck
$ GO111MODULE=on staticcheck ./...

ast/advisor.go:17:2: should not use dot imports (ST1001)
ast/ast.go:21:2: should not use dot imports (ST1001)
ast/base.go:99:6: type resultSetNode is unused (U1000)
ast/ddl.go:19:2: should not use dot imports (ST1001)
ast/ddl.go:2700:6: should omit comparison to bool constant, can be simplified to !n.Constraint.Enforced (S1002)
ast/ddl.go:2753:3: this value of col is never used (SA4006)
ast/ddl.go:2760:3: this value of constraint is never used (SA4006)
ast/dml.go:19:2: should not use dot imports (ST1001)
ast/dml.go:70:2: field resultSetNode is unused (U1000)
ast/dml.go:177:2: field resultSetNode is unused (U1000)
ast/dml.go:774:2: field resultSetNode is unused (U1000)
ast/dml.go:1065:2: field resultSetNode is unused (U1000)
ast/dml.go:1889:2: field resultSetNode is unused (U1000)
ast/expressions.go:23:2: should not use dot imports (ST1001)
ast/expressions.go:1183:12: this value of ok is never used (SA4006)
ast/functions.go:22:2: should not use dot imports (ST1001)
ast/misc.go:24:2: should not use dot imports (ST1001)
ast/misc_test.go:195:2: this value of pwd is never used (SA4006)
ast/stats.go:18:2: should not use dot imports (ST1001)
auth/auth.go:23:2: should not use dot imports (ST1001)
charset/charset.go:108:2: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (S1008)
digester.go:117:11: should use d.buffer.String() instead of string(d.buffer.Bytes()) (S1030)
digester.go:124:15: should use d.buffer.String() instead of string(d.buffer.Bytes()) (S1030)
digester.go:247:2: redundant return statement (S1023)
digester.go:363:22: func (*tokenDeque).pushBack is unused (U1000)
goyacc/format_yacc.go:486:2: field readBytes is unused (U1000)
lexer_test.go:219:4: string literal contains the Unicode control character U+0005, consider using the '\x05' escape sequence (ST1018)
[...]

PR #730, opened 2020-01-21
change all dot import, according to staticcheck

rubb1sh:

#714 

change all dot import.
### Check List 

Tests 

 - Unit test

### DIFF
--- a/ast/advisor.go
+++ b/ast/advisor.go
@@ -14,7 +14,7 @@
 package ast
 
 import (
-	. "github.com/pingcap/parser/format"
+	"github.com/pingcap/parser/format"
 )
 
 var _ StmtNode = &IndexAdviseStmt{}
@@ -31,7 +31,7 @@ type IndexAdviseStmt struct {
 }
 
 // Restore implements Node Accept interface.
-func (n *IndexAdviseStmt) Restore(ctx *RestoreCtx) error {
+func (n *IndexAdviseStmt) Restore(ctx *format.RestoreCtx) error {
 	ctx.WriteKeyWord("INDEX ADVISE ")
 	if n.IsLocal {
 		ctx.WriteKeyWord("LOCAL ")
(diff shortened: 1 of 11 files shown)


kennytm:
In the future please put substantial new changes on a new PR, otherwise we gotta restart the review cycle and delay merging.

ast/base.go
```diff
@@ -95,7 +95,3 @@ type funcNode struct {
 
 // functionExpression implements FunctionNode interface.
 func (fn *funcNode) functionExpression() {}
-
-type resultSetNode struct {
-	resultFields []*ResultField
-}
```
kennytm: PTAL @XuHuaiyu pingcap/tidb#4988. Is it safe to remove this inner struct given that nothing on `ast.ResultSetNode` relies on ResultFields anyway?

lexer_test.go
```diff
@@ -216,7 +216,7 @@ func (s *testLexerSuite) TestscanString(c *C) {
 		{`'disappearing\ backslash'`, "disappearing backslash"},
 		{"'한국의中文UTF8およびテキストトラック'", "한국의中文UTF8およびテキストトラック"},
 		{"'\\a\x90'", "a\x90"},
-		{`"\aèàø»"`, `aèàø»`},
```
kennytm: If staticcheck complains about this it is definitely wrong.

model/model_test.go
```diff
@@ -315,7 +315,7 @@ func (testModelSuite) TestUnmarshalCIStr(c *C) {
 	c.Assert(ci.O, Equals, str)
 	c.Assert(ci.L, Equals, "aabb")
 
-	buf, err = json.Marshal(ci)
+	buf, _ = json.Marshal(ci)
```
kennytm:
```suggestion
	buf, err = json.Marshal(ci)
	c.Assert(err, IsNil)
```

rubb1sh:
sorry, this is my first pr. something is wrong. And i didn't notice commit in common pr, i think commit should in my repo, but don't know why commit appears in this pr, sorry. i will make right.

kennytm:
That's okay. Since you filed the PR using the `parser` branch in your repo, GitHub will track all changes to that branch *in your repo* and update the PR accordingly. If you want to revert to the previous state (only change all dot imports), run:

```sh
git checkout parser
git checkout -b parser-new                   # save the changes in a new branch
git checkout parser
git reset --hard 19ad50b3                    # revert to previous state, drop the new commit
git push self parser --force-with-lease      # assume your repo is named "self"
```

rubb1sh:
ok, but i think i can recommit. So this commit replace that is wrong. right?

PR #731, opened 2020-02-02
parser: support expression restored string in default value.

AilinKid:
Now the default value in `columnInfo` has been evaluated.
but in some cases, the default value is dynamic, needing to be evaluated every time call it.
It's a little bit like a generated column, so we store the expression restored string in default value.

So we need a field to judge the default string is whether the `common_string` or `expr_string`.

after this pr:
```
create sequence seq
create table t(a int default next value for seq)
```
the `a`'s default value will be a string equal to `"nextval(seq)"`
the  field `DefaultIsExpr` in columnInfo of `a` will be true.

linked TiDB PR:

### Check List 

Tests 

 - Integration test

Code changes

 - Has added fields in columnInfo

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -81,12 +81,12 @@ const (
 
 // ColumnInfo provides meta data describing of a table column.
 type ColumnInfo struct {
-	ID                  int64               `json:"id"`
-	Name                CIStr               `json:"name"`
-	Offset              int                 `json:"offset"`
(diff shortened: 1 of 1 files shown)


AilinKid:
> LGTM.
> 
> BTW is it also `true` for `DEFAULT CURRENT_TIMESTAMP`?

Not really, CURRENT_TIMESTAMP has mysql flag to judge it.

PR #611, opened 2019-11-04
*: support flashback truncate table.

crazycs520:
### What problem does this PR solve? 
This is for PR: https://github.com/pingcap/tidb/pull/13109

Add syntax :
```sql
flashback table t until timestamp '2019-11-04 11:56:49.497 +0800 CST' to t1; 
```

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change

Side effects

 - Possible performance regression

Related changes

PR #733, opened 2020-02-03
*: refine flashback syntax by remove until timestamp statement.

crazycs520:
PR https://github.com/pingcap/parser/pull/611 add the `flashback` syntax. But it was hard to use with the `until timestamp` statement. 

And currently, we can't support `flashback` to any timestamp, even in the GC safe time. 

After think over and talk with PM, we decided to remove the `until timestamp` statement in the `flashback` syntax. Now the `flashback` syntax is like below:

```SQL
FLASHBACK TABLE table_name [TO] [other_name]
```

### Check List 

Tests 

 - Unit test


crazycs520:
@AilinKid  Since this PR doesn't compatible, so it can't pass the integration test, it need merge with TiDB PR at the same time.

AilinKid:
> @AilinKid Since this PR doesn't compatible, so it can't pass the integration test, it need merge with TiDB PR at the same time.

Oh, got

PR #734, opened 2020-02-03
 support register new terror

lysu:

make terror support register new error code and error msg and prepare for move error code/msg out of parser

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - impl

Side effects

 - n/a

Related changes

 - n/a

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/734)

### DIFF
--- a/terror/terror.go
+++ b/terror/terror.go
@@ -57,68 +57,51 @@ const (
 type ErrClass int
 
 // Error classes.
-const (
-	ClassAutoid ErrClass = iota + 1
-	ClassDDL
-	ClassDomain
-	ClassEvaluator
-	ClassExecutor
-	ClassExpression
-	ClassAdmin
-	ClassKV
-	ClassMeta
-	ClassOptimizer
-	ClassParser
-	ClassPerfSchema
-	ClassPrivilege
-	ClassSchema
-	ClassServer
-	ClassStructure
-	ClassVariable
-	ClassXEval
-	ClassTable
-	ClassTypes
-	ClassGlobal
-	ClassMockTikv
-	ClassJSON
-	ClassTiKV
-	ClassSession
-	ClassPlugin
-	ClassUtil
+var (
+	ClassAutoid     = RegisterErrorClass(1, "autoid")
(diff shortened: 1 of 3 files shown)


terror/terror.go
```diff
@@ -144,6 +128,15 @@ func (ec ErrClass) NotEqualClass(err error) bool {
 // New creates an *Error with an error code and an error message.
 // Usually used to create base *Error.
 func (ec ErrClass) New(code ErrCode, message string) *Error {
+	if ErrClassToMySQLCodes == nil {
+		ErrClassToMySQLCodes = make(map[ErrClass]map[ErrCode]struct{})
```
kennytm: Any chance this function is called simultaneously from two different goroutines?

lysu: no this, function only can be called in `init()` or `global variable initializer`,  I will add some comments for this :D

tiancaiamao: But there is no need for this lazy initialization @lysu

terror/terror.go
```diff
@@ -291,21 +284,20 @@ func (e *Error) getMySQLErrorCode() uint16 {
...
 
 var (
-	// ErrClassToMySQLCodes is the map of ErrClass to code-map.
-	ErrClassToMySQLCodes map[ErrClass]map[ErrCode]uint16
+	// ErrClassToMySQLCodes is the map of ErrClass to code-set.
+	ErrClassToMySQLCodes map[ErrClass]map[ErrCode]struct{}
 )
 
 func init() {
-	ErrClassToMySQLCodes = make(map[ErrClass]map[ErrCode]uint16)
```
tiancaiamao: We can keep this way to initialize `ErrClassToMySQLCodes`.

lysu:
==  it has some race problem need fix

lysu:
@jackysp @zimulala  hi, I have do some change, add a new `Synthesize`, because I found TiDB will create terror base on TiKV's error code in cop select_result.go, we can not register TiKV error code directly in current logic(maybe it's not right..),  so add a new method just create terror but didn't register code - -

PR #735, opened 2020-02-05
terror: expose args from Error

sticnarf:

Required by https://github.com/pingcap/tidb/pull/14626, the `args` of `Error` need be exposed.

Add a method to `Error` that returns the `args`.

### Check List 

Tests 

 - No need for tests because it's too trivial.

Code changes

### DIFF
--- a/terror/terror.go
+++ b/terror/terror.go
@@ -177,6 +177,11 @@ func (e *Error) Code() ErrCode {
 	return e.code
 }
 
+// Args returns the inner args
+func (e *Error) Args() []interface{} {
+	return e.args
+}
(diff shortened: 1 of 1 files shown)


terror/terror.go
```diff
@@ -177,6 +177,11 @@ func (e *Error) Code() ErrCode {
 	return e.code
 }
 
+// Args returns the inner args
```
zimulala:
```suggestion
// Args returns the inner args.
```

sticnarf:
Closed as pingcap/tidb#14626 is closed

PR #743, opened 2020-02-12
parser: add WEIGHT_STRING() function

bb7133:
This PR is the implementation of `WEIGHT_STRING()` in parser, you can check the [MySQL manual](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_weight-string) for more details.

Here we only implement a part of the full syntax: `WEIGHT_STRING(str [AS {CHAR|BINARY}(N)])`, according to MySQL manual:

> WEIGHT_STRING() is a debugging function intended for internal use. Its behavior can change without notice between MySQL versions. It can be used for testing and debugging of collations, especially if you are adding a new collation. 

We can adjust this function on our own needs.

Related issue: https://github.com/pingcap/tidb/issues/3580
Related TiDB PR: https://github.com/pingcap/tidb/pull/14792

Add `WEIGHT_STRING()` in parser.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/ast/functions.go
+++ b/ast/functions.go
@@ -226,6 +226,7 @@ const (
 	CharLength      = "char_length"
 	CharacterLength = "character_length"
 	FindInSet       = "find_in_set"
+	WeightString    = "weight_string"
 
 	// information functions
 	Benchmark      = "benchmark"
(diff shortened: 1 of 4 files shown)


ast/functions.go
```diff
@@ -420,6 +421,21 @@ func (n *FuncCallExpr) Restore(ctx *format.RestoreCtx) error {
...
 			}
 		}
+	case WeightString:
+		if err := n.Args[0].Restore(ctx); err != nil {
+			return errors.Annotatef(err, "An error occurred while restore FuncCallExpr.(WEIGHT_STRING).Args[0]")
+		}
+		if len(n.Args) == 3 {
+			ctx.WritePlain(" ")
+			ctx.WriteKeyWord("AS")
+			ctx.WritePlain(" ")
```
kennytm:
```go
ctx.WriteKeyWord(" AS ")
```

tiancaiamao:
Please address comment @bb7133

ast/functions.go
bb7133: Addressed, thank you.

bb7133:
PTAL @wjhuang2016

parser.y
```diff
@@ -5858,6 +5860,32 @@ FunctionCallNonKeyword:
...
+		}
+	}
+|	weightString '(' Expression "AS" "BINARY" FieldLen ')'
+	{
+		$$ = &ast.FuncCallExpr{
+			FnName: model.NewCIStr($1),
+			Args:   []ast.ExprNode{$3, ast.NewValueExpr("BINARY"), ast.NewValueExpr($6)},
+		}
+	}
+/** The following syntax is defined in MySQL parser, but not described in MySQL reference manual, we do not support it for now.
```
wjhuang2016: I don't think we need to support it, since we don't support `flag`.
It's strange to support a grammer not described in MySQL reference manual but don't support a grammer  described in MySQL reference manual.

bb7133: So what's you suggesion, remove this?

wjhuang2016: Yeah,

bb7133: Done.

PR #747, opened 2020-02-18
refine table optimizer hint and make it more generic

lonng:
Signed-off-by: Lonng <heng@lonng.org>

The struct `TableOptimizerHint` is used to save optimizer information, but its field count growth as various types of hint adding.  

1. This PR introduces a new `HintData` to save the generic hint payload and remove the extra fields to reduce the struct size.
2. This PR adds a new hint `TIME_RANGE`, which is used to hint the time range of `inspection_result/inspection_summary` system table.

### Check List 

Tests 

 - Unit test


ast/misc.go
```diff
@@ -2198,17 +2198,31 @@ type TableOptimizerHint struct {
 	// Table hints has no schema info
 	// It allows only table name or alias (if table has an alias)
 	HintName model.CIStr
+	// HintData is the payload of the hint. The actual type of this field
+	// is defined differently as according `HintName. Define as following:
```
tangenta:
```suggestion
	// is defined differently as according `HintName`. Define as following:
```

PR #749, opened 2020-02-23
parser, ast: add `SetCollationExpr`

bb7133:
To support the syntax like the followings:

- `select 'a' collate utf8mb4_general_ci`
- `select a collate utf8_general_ci from t`

In which, the "`collate`" syntax is ignored before. 

In this PR a `SetCollatorExpr` is added to keep it. This ExprNode is equivalent to the `Item_func_set_collation` in MySQL,
it is used to specify the collation for the result of another
expression.

### Check List 

Tests 

 - Unit test
 - Integration test
 
Code changes

Side effects

 - NA

Related changes


bb7133:
PTAL @kennytm @tangenta @zimulala

PR #761, opened 2020-03-05
*: set charset and collate for string literals correctly

wjhuang2016:
set charset and collate for string literal correctly

In `NewValueExpr`, set the charset and collate for string literal.
The charset and collate are from parser, which is the connection charset and connection collation.
See also https://dev.mysql.com/doc/refman/8.0/en/charset-literal.html

### Check List 

Tests 

 - No code
Tests are added in TiDB, need to merge together.

Code changes

Side effects

Related changes

### DIFF
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -65,7 +65,7 @@ type ValueExpr interface {
 }
 
 // NewValueExpr creates a ValueExpr with value, and sets default field type.
-var NewValueExpr func(interface{}) ValueExpr
+var NewValueExpr func(interface{}, string, string) ValueExpr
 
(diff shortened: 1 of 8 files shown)


ast/expressions.go
```diff
@@ -65,7 +65,7 @@ type ValueExpr interface {
 }
 
 // NewValueExpr creates a ValueExpr with value, and sets default field type.
-var NewValueExpr func(interface{}) ValueExpr
+var NewValueExpr func(interface{}, string, string) ValueExpr
```
kennytm:
```suggestion
var NewValueExpr func(value interface{}, charset string, collation string) ValueExpr
```

PR #763, opened 2020-03-09
parser: set text for explainable statements

eurekaka:

SPM in TiDB uses the `Text()` method of `SelectStmt` in `ExplainStmt` to do pattern matching, but this field is empty now.

Call `SetText()` for those explainable statements.

### Check List 

Tests 

 - Unit test

Code changes

N/A

Side effects

N/A

Related changes

N/A

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -11688,6 +11688,8 @@ yynewstate:
 				Stmt:   yyS[yypt-0].statement,
 				Format: "row",
 			}
+			startOffset := parser.startOffset(&yyS[yypt])
+			yyS[yypt-0].statement.SetText(string(parser.src[startOffset:]))
 		}
 	case 400:
 		{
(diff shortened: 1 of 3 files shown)


kennytm:
Can you change SPM to use `Restore()` instead? 🤔

eurekaka:
The restored SQL may not be as strictly same as the original text? this may cause SQL binding not being able to be used.

kennytm:
right... but then `.Text()` will also treat `SELECT 1 FROM a.b;` and ``select 1 from `a`.`b`;`` as different

eurekaka:
Looks like only `.Restore()` would add `` ` ` ``, while the `.Text()` function returns exactly same string as the input.

eurekaka:
We extract select text from `explain` explicitly in tidb code now, so this PR is not needed anymore and I close it.

PR #765, opened 2020-03-11
parser: add autoid cache option

AilinKid:
By now, the auto-increment cache in TiDB is quite big for inserting performance.

In some cases, some table is used as a tool to generate some id, the autoid cache will miss when tidb crash or restart,  which leading autoid increment quickly.

The related to pingcap/tidb#15409

That's why we support the auto-increment cache option.

Add `auto_increment_cache` table option.
Make some test about it.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


tiancaiamao:
Is there a TiDB PR for this change?

AilinKid:
> Is there a TiDB PR for this change?

Yes，it's still work in process.

tiancaiamao:
> > Is there a TiDB PR for this change?
> 
> Yes，it's still work in process.

After it's done, please link that PR with this one, so we can merge them together. @AilinKid

parser.y
```diff
@@ -4884,6 +4885,7 @@ UnReservedKeyword:
 |	"ADVISE"
 |	"ASCII"
 |	"AUTO_INCREMENT"
+|	"AUTO_INCREMENT_CACHE"
```
bb7133: Please rename it to `AUTO_ID_CACHE`, as you mentioned, it doesn't affect AutoIncrement column only.

bb7133:
Please address the comment and fix CI.

bb7133:
LGTM, but let's hold it DNM until the TiDB PR is ready to merge.

PR #766, opened 2020-03-11
parser: support group_concat as window function in parser

zhaox1n:

group_concat as a window function cannot be resolved in the parser. The detail you can see in TIDB [#10996](https://github.com/pingcap/tidb/issues/10996)

### Check List 

Tests 

 - Unit test
 - Integration test


zhaox1n:
Is there anything wrong with the license checker? @zz-jason

zz-jason:
> Is there anything wrong with the license checker? @zz-jason

I think so.

Issue #768, opened 2020-03-12
Support `set config ...` and `show config ...` syntaxes


qw4990:
## Feature Request

**Is your feature request related to a problem? Please describe:**

No.

**Describe the feature you'd like:**

After implementing this proposal(https://github.com/pingcap/tidb/pull/13660), TiDB, TiKV, and PD all support to change configs dynamically online.
Now we can update configs dynamically according to `pd-ctl`, which is not convenient to our user, so we need to support updating configs by SQL.
After discussing, we decide to support these syntaxes:
1. set config {component or instance} key=val
2. show config where ...

For example: 
1. set config TiDB Log.Level="info"
2. set config '127.0.0.1:3306' Log.Level="info"
3. show config where type="TiDB"

qw4990:
Close due to finish.

PR #770, opened 2020-03-13
parser: support `set config ...` and `show config ...` syntaxes

qw4990:
Implement https://github.com/pingcap/parser/issues/768.

### Check List 

Tests 

 - Unit test


ast/misc.go
```diff
@@ -817,6 +817,41 @@ func (n *SetStmt) Accept(v Visitor) (Node, bool) {
...
+	Name     string // the variable name
+	Value    ExprNode
+}
+
+func (n *SetConfigStmt) Restore(ctx *format.RestoreCtx) error {
+	ctx.WriteKeyWord("SET CONFIG ")
+	if n.Type != "" {
+		ctx.WriteKeyWord(n.Type)
+	} else {
+		ctx.WritePlainf("'%s'", n.Instance)
```
```suggestion
		ctx.WriteString(n.Instance)
```

ast/misc.go
```diff
@@ -817,6 +817,41 @@ func (n *SetStmt) Accept(v Visitor) (Node, bool) {
...
+	}
+	ctx.WritePlainf(" %s=", n.Name)
```
```suggestion
	ctx.WritePlain(" ")
	ctx.WriteKeyWord(n.Name)
	ctx.WritePlain(" = ")
```

parser.y
```diff
@@ -7726,6 +7727,18 @@ SetStmt:
...
 		$$ = &ast.SetStmt{Variables: assigns}
 	}
+|	"SET" "CONFIG" "TIDB" VariableName EqOrAssignmentEq SetExpr
+	{
+		$$ = &ast.SetConfigStmt{Type: "TiDB", Name: $4, Value: $6}
+	}
+|	"SET" "CONFIG" identifier VariableName EqOrAssignmentEq SetExpr
+	{
+		$$ = &ast.SetConfigStmt{Type: $3, Name: $4, Value: $6}
+	}
```
kennytm: Perhaps better use a single rule with `Identifier` and store the type in lower-case like

```suggestion
|	"SET" "CONFIG" Identifier VariableName EqOrAssignmentEq SetExpr
	{
		$$ = &ast.SetConfigStmt{Type: strings.ToLower($3), Name: $4, Value: $6}
	}
```

qw4990:
Thanks, @kennytm , all comments have been addressed now.

tiancaiamao:
LGTM

ast/misc.go
```diff
@@ -817,6 +817,43 @@ func (n *SetStmt) Accept(v Visitor) (Node, bool) {
 	return v.Leave(n)
 }
 
+// SetConfigStmt ...
```
XuHuaiyu: unfinished comment

parser_test.go
```diff
@@ -1025,6 +1025,15 @@ func (s *testParserSuite) TestDBAStmt(c *C) {
 		{"set names utf8, @@session.sql_mode=1;", true, "SET NAMES 'utf8', @@SESSION.`sql_mode`=1"},
 		{"set @@session.sql_mode=1, names utf8, charset utf8;", true, "SET @@SESSION.`sql_mode`=1, NAMES 'utf8', NAMES 'utf8'"},
 
+		// for set/show config
+		{"set config TIKV LOG.LEVEL='info'", true, "SET CONFIG TIKV LOG.LEVEL = 'info'"},
+		{"set config PD LOG.LEVEL='info'", true, "SET CONFIG PD LOG.LEVEL = 'info'"},
+		{"set config TIDB LOG.LEVEL='info'", true, "SET CONFIG TIDB LOG.LEVEL = 'info'"},
+		{"set config '127.0.0.1:3306' LOG.LEVEL='info'", true, "SET CONFIG '127.0.0.1:3306' LOG.LEVEL = 'info'"},
```
XuHuaiyu: It'll be better to add some invalid test cases. Such as `set config tipd log.level = 'info'`, etc.

ast/misc.go
qw4990: fixed.

parser_test.go
qw4990: It's designed to let the `TiDB` check if the `type` is valid instead of `Parser`, which is more extendable, for example, if we want to support changing `TiFlash` configs online, we don't need to modify the `Parser` again.

parser.y
```diff
@@ -8616,6 +8625,10 @@ ShowTargetFilterable:
 	{
 		$$ = &ast.ShowStmt{Tp: ast.ShowDatabases}
 	}
+|	"CONFIG"
```
kennytm: Wait, given that this is listed under `ShowTargetFilterable`, do we expect the following to work?

```sql
SHOW CONFIG LIKE 'foo%';
```

qw4990: Yes, `show config like '%log%'` and `show config where name='log.level'` both are ok.

kennytm:
Please fix conflict so this can be merged.

qw4990:
@kennytm updated, PTAL.

misc.go
```diff
@@ -204,6 +204,7 @@ var tokenMap = map[string]int{
 	"BYTE":                       byteType,
 	"CACHE":                      cache,
 	"CANCEL":                     cancel,
+	"CONFIG":                     config,
```
kennytm: this should be placed after line 236.

qw4990: fixed, PTAL 😃

PR #777, opened 2020-03-16
lexer: replace version comments with feature-ids mechanism

tangenta:
Executable special comments like `/*T!40000 xxx */` are not suitable for TiDB. Due to the nature of branch-release, new features can be cherry-picked to different major versions TiDB. As a result, a TiDB with a higher version may not support the features in a TiDB with a lower version. 

To solve this problem, this PR gives a mechanism to manage the features that need to support forward compatibility:
- For the features that need to be downgrade compatible, assign a unique identifier to each of them, called `feature-id`.
- The parsers in different release branches maintain their own `featureMap`, which contains a set of currently supported feature-ids.
- The pattern `/*T![feature_id] xxx */` is used to wrap the key part during logical synchronization.
- To decide whether a statement with special comments should be parsed, parser can search `feature_id` in `featureMap`.
- In rare cases, a comment can have multiple feature ids like `/*T![feature_id1,feature_id2,...] xxx */`. It can only be parsed if all feature IDs in the comment can be found in the `featureMap`.

Add `scanFeatureIDs()` to scan feature ids in special comments.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


misc.go
```diff
@@ -886,6 +866,23 @@ var hintTokenMap = map[string]int{
...
 
+type FeaturesMap map[string]struct{}
+
+func (m FeaturesMap) containsAll(features []string) bool {
+	allFound := true
+	for _, f := range features {
+		_, found := m[f]
+		allFound = allFound && found
+	}
+	return allFound
```
kennytm:
```suggestion
	for _, f := range features {
		if _, found := m[f]; !found {
			return false
		}
	}
	return true
```

misc.go
```diff
@@ -886,6 +866,23 @@ var hintTokenMap = map[string]int{
 	"MATERIALIZATION": hintMaterialization,
 }
 
+type FeaturesMap map[string]struct{}
```
kennytm: does this type need to be public?

tangenta: @kennytm I came up with an idea: how about putting the `FeaturesMap` into a part of the parser driver? Let users decide what features should be parsed.

tiancaiamao:
The parser should not care about the detail of what `/*T!40000 xxx */` means.
It just parses that value as an opaque data, and return it to TiDB.

If we change to `FeaturesMap` that means the code bind with the parser.
It's no longer a black box for the parser.

lexer.go
```diff
@@ -748,21 +748,66 @@ func (s *Scanner) scanDigits() string {
...
 		} else {
 			break
 		}
 	}
-	return
+}
+
+func (s *Scanner) scanFeatureIDs() (featureIDs []string) {
+	pos := s.r.pos()
+	const init, expectAlpha, alpha = 0, 1, 2
```
AilinKid: state name should more clear?

tangenta:
@tiancaiamao Do you have any suggestions to deal with the problem mentioned in description?

How about putting the `FeaturesMap` into a part of the parser driver? Let users decide what should be parsed.

tiancaiamao:
> Do you have any suggestions to deal with the problem mentioned in the description?

The idea sounds good. I just have concerns about the compatibility issue.
When the TiDB uses this new feature, what happens if the customer rolling updates from an old cluster?

> How about putting the FeaturesMap into a part of the parser driver? Let users decide what should be parsed.

Much better.

tangenta:
IMO, it doesn't affect rolling updates.

Currently, only `auto_random` feature is using the special comments like `/*T!40000 xxx */`. I think compatibility is not the problem, because it is an experimental feature :) 

> In later TiDB versions, the syntax or semantics of AUTO_RANDOM might change.
https://pingcap.com/docs/dev/reference/sql/attributes/auto-random/

kennytm:
in the worst case we could support both `/*T!12345 ... */` and `/*T![abc,def] ... */` since the two syntax are mutually exclusive.

misc.go
```diff
@@ -886,6 +866,23 @@ var hintTokenMap = map[string]int{
...
+	for _, f := range features {
+		if _, found := m[f]; !found {
+			return false
+		}
+	}
+	return true
+}
+
+// featureMap is a set of feature IDs supported by current parser.
+var featuresMap = featureMap{
```
tiancaiamao: Let the user to fill this map. You can provide a function like `RegistFeature`.
Or just let the parser return the parsed feature list strings and let the user to handle them.

tiancaiamao:
If none of our users is actually using `/*T!40000 xxx */`, it doesn't affect rolling updates.

lexer.go
```diff
@@ -360,9 +360,9 @@ func startWithSlash(s *Scanner) (tok int, pos Pos, lit string) {
 			break
 		}
 		s.r.inc()
-		// in '/*T!', try to consume the 5 to 6 digit version string.
-		commentVersion := s.scanVersionDigits(5, 6)
-		if commentVersion <= CommentCodeCurrentVersion {
+		// in '/*T!', try to match the pattern '/*T![feature1,feature2,...]'.
+		features := s.scanFeatureIDs()
+		if len(features) != 0 && SpecialCommentsController.ContainsAll(features) {
```
kennytm: OK so there's this thing: 

Before this PR, `SELECT /*T! now() */;` is valid because the "version" of `/*T!` without 5 digits following it is always 0. 

So this line should be changed to

```suggestion
		if SpecialCommentsController.ContainsAll(features) {
```

lexer.go
```diff
@@ -748,21 +748,66 @@ func (s *Scanner) scanDigits() string {
...
+		case obtainChar:
+			if isIdentChar(ch) {
+				b.WriteRune(ch)
+				state = obtainChar
+				break
+			} else if ch == ',' {
+				featureIDs = append(featureIDs, b.String())
+				b.Reset()
+				state = expectChar
+				fmt.Print(",")
```
tiancaiamao: remove debug log

lexer.go
```diff
@@ -360,9 +360,9 @@ func startWithSlash(s *Scanner) (tok int, pos Pos, lit string) {
...
 		s.r.inc()
-		// in '/*T!', try to consume the 5 to 6 digit version string.
-		commentVersion := s.scanVersionDigits(5, 6)
-		if commentVersion <= CommentCodeCurrentVersion {
+		// in '/*T!', try to match the pattern '/*T![feature1,feature2,...]'.
+		features := s.scanFeatureIDs()
+		if SpecialCommentsController.ContainsAll(features) {
 			s.inBangComment = true
 			return s.scan()
 		}
```
tiancaiamao: What happens if the feature id is not included in the list?

tangenta: It will parse it as normal comments. For now, the behavior is:
`FeatureMap: {"xxx", "yyy"}`

```
/*T![xxx,yyy] part_of_sql */  # all the feature ids matched
/*T![xxx,zzz] part_of_sql */  # one feature id mismatched
/*T![xxx part_of_sql */       # no feature id is detected
/*T!part_of_sql */            # no feature id detected
```
are equivalant to 
```
part_of_sql

[xxx part_of_sql
part_of_sql
```
respectively.

tiancaiamao:
LGTM
Please fix the integration test.

tangenta:
This PR should be merged together with pingcap/tidb#15412.

PR #778, opened 2020-03-16
parser: transfer `ServerStatus` to readable message

clark1013:
- to solve issue in TiDB: https://github.com/pingcap/tidb/issues/7367
- PR in TiDB: https://github.com/pingcap/tidb/pull/15413

- Add a map transfer server status to string

### Check List 

Tests 

 - Unit test


clark1013:
> The status is a bit-field. You can't convert the status into a string by a single map lookup (e.g. the status can be `0x0003` = `in transaction; autocommit`).

copy that. I will update this PR.

mysql/const.go
```diff
@@ -58,6 +58,54 @@ const (
...
 )
 
+// ServerStatus2Str convert server status to string.
+// Param state is a bit-field. (e.g. 0x0003 = "in transaction; autocommit").
+func ServerStatus2Str(state uint16) string {
+	// l collect server status strings.
+	var l []string
+	// status is a slice of all defined server status in ascending order.
+	var status []uint16
+	status = append(
```
tiancaiamao: Every time this function is called, it makes an allocation here?
How about moving the definition out so the allocation is avoided

mysql/const.go
```diff
@@ -58,6 +58,54 @@ const (
...
+		ServerStatusCursorExists,
+		ServerStatusLastRowSend,
+		ServerStatusDBDropped,
+		ServerStatusNoBackslashEscaped,
+		ServerStatusMetadataChanged,
+		ServerStatusWasSlow,
+		ServerPSOutParams,
+	)
+	// mapServerStatus is the map for server status to string.
+	var mapServerStatus = map[uint16]string{
```
tiancaiamao: ditto

clark1013:
The CI failed at `planner`. I am not sure what has I did wrong. Expect someone could give me a guideline.
image

tiancaiamao:
> The CI failed at `planner`. I am not sure what has I did wrong. Expect someone could give me a guideline.
> image

Maybe it's casued by the latest parser rather than this change.

clark1013:
> > The CI failed at `planner`. I am not sure what has I did wrong. Expect someone could give me a guideline.
> > image
> 
> Maybe it's casued by the latest parser rather than this change.

OK, I will try to update my branch to master. If it still fail. I will ignore it.

clark1013:
> This change can be done in TiDB.
> If the function is rarely used, it's not necessary to provide a global function for it.

Thank you for your review. About this problem, my starting point is packaging server status in `mysql` package, as like `Command2Str` do so. Assume that server status extend, there will be no need to update code in TiDB.  So should I close this PR and transfer the change to TiDB?

clark1013:
As mentioned in https://github.com/pingcap/tidb/pull/15413. I will make the change in TiDB. This PR will be closed.

PR #791, opened 2020-04-01
parser: sequence order option parse warning

AilinKid:
add sequence `order` option parse warning when create sequence.

just like `temporary` do, add the parse warning when scan to this.

### Check List 

Tests 
 - No code


tangenta:
I think it is better to throw the warning in TiDB. WDYT @AilinKid @kennytm

kennytm:
If we do so this should apply to all existing warnings of the same kind (STATS_AUTO_RECALC, STATS_SAMPLE_PAGES, SECONDARY_ENGINE, ENCRYPTION, ...)

PR #792, opened 2020-04-01
opcode,ast: fix invalid restore of '4 DIV 2' (became '4DIV2')

kennytm:

It was found that in pingcap/tidb#15948 the Restore() involving the `DIV` operator was incorrect: `4 DIV 2` by default was restored without spaces as `4DIV2`.

Here we refactored the `opcode` module to include whether the operator is a keyword. Spaces around keyword-like ops are enforced.

### Check List 

Tests 

 - Unit test

Code changes

    * I don't know why the map `opcode.Ops` was exported. It is used nowhere. This PR makes it private.

Side effects

Related changes

    * needs to cherry-pick to all previous releases.

### DIFF
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -156,19 +156,27 @@ type BinaryOperationExpr struct {
 	R ExprNode
 }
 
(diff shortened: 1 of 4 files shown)


tiancaiamao:
LGTM

PR #796, opened 2020-04-07
parser: remove sequence order option support

AilinKid:
Since the order option is not really implemented in TiDB (just syntax compatible)

The warnings message will be ignored by some users, which may migrate their business from Oracle etc to TiDB, consequently, leading the inconsistency.

So better remove it support from the very beginning.

remove order option in yy and ast.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


AilinKid:
do not merge now

tangenta:
Report an error in TiDB can be another choice.

AilinKid:
> Report an error in TiDB can be another choice.

sounds good!

PR #803, opened 2020-04-09
terror: print the stack in terror.Log()

tiancaiamao:
In TiDB, `terror.Log(err)` is called when we expect the error to be nil and get rid of the annoying golint check.
Sometimes the error is not nil as expected, so it's better to print the stack to find why.

> [2020/04/08 21:09:00.837 +08:00] [ERROR] [terror.go:363] ["encountered error"] [error="[ddl:1273]Unknown collation: '

PTAL @lysu @kennytm


tiancaiamao:
> we can only get stack when calling terror.Log from this pr.

That's enough for us to modify the code (identify the incorrect `terror.Log` call)

terror/terror.go
```diff
@@ -360,6 +361,8 @@ func Call(fn func() error) {
 // Log logs the error if it is not nil.
 func Log(err error) {
 	if err != nil {
-		log.Error("encountered error", zap.Error(err))
+		var buf [4096]byte
+		sz := runtime.Stack(buf[:], false)
+		log.Error("encountered error", zap.Error(err), zap.String("stack", string(buf[:sz])))
```
kennytm, line 366: Perhaps easier to just do

```suggestion
		log.Error("encountered error", zap.Error(errors.Trace(err)))
```

tiancaiamao: Yes, we need it!
We need the call stack to identify the misuse of `terror.Log()`.  @kennytm

kennytm: @tiancaiamao Ah ok. In that case `errors.Trace` isn't suitable since it won't replace the error's stack if it already exists. We should use [`errors.WithStack`](https://pkg.go.dev/github.com/pingcap/errors?tab=doc#WithStack)

tiancaiamao: Done

PR #804, opened 2020-04-09
charset: return a default value for GetCollationByName("")

tiancaiamao:

I see a annoying error log in TiDB like this:

> [2020/04/09 22:26:46.422 +08:00] [ERROR] [terror.go:366] ["encountered error"] [error="[ddl:1273]Unknown collation: ''"] 

and it's caused by here

https://github.com/pingcap/tidb/blob/2a2bf377aed3bc16ce5873a30f4c6d9b15dc6c0e/executor/simple.go#L533

Maybe return a default value is better for `GetCollationByName("")`.

Let `GetCollationByName("")` return a default value.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/charset/charset.go
+++ b/charset/charset.go
@@ -170,6 +170,9 @@ func GetCollations() []*Collation {
 }
 
 func GetCollationByName(name string) (*Collation, error) {
+	if name == "" {
+		name = mysql.DefaultCollationName
+	}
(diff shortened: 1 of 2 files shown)


tiancaiamao:
PTAL @bb7133 @wjhuang2016

Issue #798, opened 2020-04-07
Add builtin function prototype `MVCC_GET_BY_HEX` in parser


imtbkcat:
## Feature Request
This is sub issue of https://github.com/pingcap/tidb/issues/15938

**Is your feature request related to a problem? Please describe:**

To implement getting mvcc info through SQL, TiDB need to support 4 builtin function.
`MVCC_GET_BY_HEX`  is one of them.

**Describe the feature you'd like:**

add `MVCC_GET_BY_HEX` into parser.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

- Add function name in to `functions.go`
- Add test case.

zz-jason:
<img width="996" alt="Screen Shot 2020-04-07 at 16 20 21" src="https://user-images.githubusercontent.com/5268763/78646512-b9519280-78eb-11ea-968f-262929e3d395.png">

Can we only add one builtin function like `get_mvcc_info(method string, value string)` to handle all these scenarios?

imtbkcat:
@zz-jason yes, it will be much clean. so we could close other three issue, and keep one issue for UCP contributor.

hsqlu:
@zz-jason @imtbkcat If we using the single method `get_mvcc_info`, it means the parameters passed in have to be in order and have different meanings depends on the first method name?  Is it a little bit complicated?

zz-jason:
Actually, there are some MySQL builtin functions like this. I think the complication can be accepted.

zz-jason:
Actually, there are some MySQL builtin functions like this. I think the complication can be accepted.

kennytm:
you could make `get_mvcc_info()` return a table like [`json_table()`](https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html#function_json-table), or just return a JSON type and let the user to `get_mvcc_info()->>'$.hex'`.

PR #805, opened 2020-04-10
Add builtin function prototype `get_mvcc_info` in parser

hsqlu:
Signed-off-by: Qiannan Lyu <lvqiannan@gmail.com>

#798 

### Check List 

Tests 

 - Unit test

Code changes

Related changes


parser_test.go
```diff
@@ -1298,6 +1298,7 @@ func (s *testParserSuite) TestBuiltin(c *C) {
 		{`SELECT tidb_is_ddl_owner();`, true, "SELECT TIDB_IS_DDL_OWNER()"},
 		{`SELECT tidb_decode_plan();`, true, "SELECT TIDB_DECODE_PLAN()"},
 		{`SELECT tidb_decode_key('abc');`, true, "SELECT TIDB_DECODE_KEY('abc')"},
+		{`SELECT get_mvcc_info('hex', '0xabc');`, true, "SELECT GET_MVCC_INFO('hex', '0xabc')"},
```
kennytm, line 1301: isn't it more natural to write `get_mvcc_info('hex', 0xabc)` 🤔

imtbkcat: @kennytm This value is a hex decoded key, not integer value, I think use `'7480000000000000FF2E5F698000000000FF0000010380000000FF05F5E10003000000FF0000000000000000F7'` is better.

kennytm: ok. then the input should be `'0abc'` 🙃.

also note that in MySQL `0xabc` is not an integer, but a binary blob.

```
mysql> select 0x416464206275696C74696E2066756E6374696F6E2070726F746F7479706520606765745F6D7663635F696E666F6020696E20706172736572;
+--------------------------------------------------------------------------------------------------------------------+
| 0x416464206275696C74696E2066756E6374696F6E2070726F746F7479706520606765745F6D7663635F696E666F6020696E20706172736572 |
+--------------------------------------------------------------------------------------------------------------------+
| Add builtin function prototype `get_mvcc_info` in parser                                                           |
+--------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select x'416464206275696C74696E2066756E6374696F6E2070726F746F7479706520606765745F6D7663635F696E666F6020696E20706172736572';
+---------------------------------------------------------------------------------------------------------------------+
| x'416464206275696C74696E2066756E6374696F6E2070726F746F7479706520606765745F6D7663635F696E666F6020696E20706172736572' |
+---------------------------------------------------------------------------------------------------------------------+
| Add builtin function prototype `get_mvcc_info` in parser                                                            |
+---------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

PR #808, opened 2020-04-13
support gbk charset

spongedu:
1. Support charset gbk in parser; 
2. fix description for charsetlatin and charsetbin1 and charsetbin

We're using parser to develop some our systems and found that charset gbk is not supported yet in parser, which makes some SQL fail to work. 
As a MySQL compatible parser, I think we should support gbk as MySQL's parser do, and leave the charset-compatible issues to `TiDB` layer.

Add gbk into supported charsets, and add related tests

### Check List 

Tests 

 - Unit test
 - Manual test


charset/charset.go
```diff
@@ -56,17 +56,19 @@ var charsetInfos = []*Charset{
 	{CharsetUTF8, CollationUTF8, make(map[string]*Collation), "UTF-8 Unicode", 3},
 	{CharsetUTF8MB4, CollationUTF8MB4, make(map[string]*Collation), "UTF-8 Unicode", 4},
 	{CharsetASCII, CollationASCII, make(map[string]*Collation), "US ASCII", 1},
-	{CharsetLatin1, CollationLatin1, make(map[string]*Collation), "Latin1", 1},
-	{CharsetBin, CollationBin, make(map[string]*Collation), "binary", 1},
+	{CharsetLatin1, CollationLatin1, make(map[string]*Collation), "cp1252 West European", 1},
+	{CharsetBin, CollationBin, make(map[string]*Collation), "Binary pseudo charset", 1},
+	{CharsetGBK, CollationGBKChineseCi, make(map[string]*Collation), " GBK Simplified Chinese", 1},
```
kennytm, line 61:
```suggestion
	{CharsetGBK, CollationGBKChineseCi, make(map[string]*Collation), "GBK Simplified Chinese", 1},
```

kennytm (commented):
Rest LGTM.

Why do we support GBK rather than its superset GB18030?

spongedu:
> Rest LGTM.
> 
> Why do we support GBK rather than its superset GB18030?

Because we just met `gbk` in our cases.  maybe gb18030 should be supported as well

tiancaiamao:
PTAL @bb7133 @wjhuang2016

charset/charset.go
```diff
@@ -56,17 +56,19 @@ var charsetInfos = []*Charset{
 	{CharsetUTF8, CollationUTF8, make(map[string]*Collation), "UTF-8 Unicode", 3},
 	{CharsetUTF8MB4, CollationUTF8MB4, make(map[string]*Collation), "UTF-8 Unicode", 4},
 	{CharsetASCII, CollationASCII, make(map[string]*Collation), "US ASCII", 1},
-	{CharsetLatin1, CollationLatin1, make(map[string]*Collation), "Latin1", 1},
-	{CharsetBin, CollationBin, make(map[string]*Collation), "binary", 1},
+	{CharsetLatin1, CollationLatin1, make(map[string]*Collation), "cp1252 West European", 1},
+	{CharsetBin, CollationBin, make(map[string]*Collation), "Binary pseudo charset", 1},
+	{CharsetGBK, CollationGBKChineseCi, make(map[string]*Collation), "GBK Simplified Chinese", 1},
```
tangenta, line 61: https://en.wikipedia.org/wiki/GBK_(character_encoding)
```suggestion
	{CharsetGBK, CollationGBKChineseCi, make(map[string]*Collation), "GBK Simplified Chinese", 2},
```

> A character is encoded as 1 or 2 bytes.

wjhuang2016 (commented):
`GetSupportedCharsets`  need to change too.

spongedu:
@wjhuang2016  PTAL :)

PR #809, opened 2020-04-13
fix type conversion panic from NUM to int64

tangenta:

The parser is panicked when it try to parse the following text SQLs:
```sql
recover table by job 18446744073709551612;
recover table t 18446744073709551612;
admin check index t idx (0, 18446744073709551612);
create user abc@def with max_queries_per_hour 18446744073709551612;
```

Add a rule `Int64Num` representing 64-bit signed integers to avoid casting to int64 directly.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


goyacc/format_yacc.go
```diff
@@ -39,9 +39,7 @@ func Format(inputFilename string, goldenFilename string) (err error) {
 	if err = yFmt.Setup(goldenFilename); err != nil {
 		return err
 	}
-	defer func() {
-		err = yFmt.Teardown()
-	}()
```
AilinKid, line 44: why change this? here will store the err about Teardown() failure

tangenta: If we store the error of `Teardown()`, the original error will be overridden.

parser.y
```diff
@@ -1157,6 +1157,7 @@ import (
 	OptCharset                             "Optional Character setting"
 	OptCollate                             "Optional Collate setting"
 	IgnoreLines                            "Ignore num(int) lines"
+	Int64Num                               "A number that can be converted to signed 64 bits integer"
```
zz-jason, line 1160: what's the difference with `num`?

tangenta: It checks if it is ok to cast to `int64`.

goyacc/format_yacc.go
kennytm: maybe only replace error when it was nil.

```go
teardownErr := yFmt.Teardown()
if err == nil {
    err = teardownErr
}
```

PR #821, opened 2020-04-22
parser: support using dash('-') in config item names

### LINKED ISSUE #820: Support using dash('-') in config item names
## Feature Request

**Is your feature request related to a problem? Please describe:**

After this issue(https://github.com/pingcap/parser/issues/768), the parser supports `set config name=value` syntax, but the name doesn't support using `-`.
We should support it since there are some config items like `label-property.reject-leader`, `auto-compaction` which use `-` in their names.

**Describe the feature you'd like:**

Support using dash(`-`) in config item names, for examples:
1. `set config pd auto-compation=true`
2. `set config pd property.reject-leader.key='zone'`

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

qw4990:
Close #820
After this issue(https://github.com/pingcap/parser/issues/768), the parser supports `set config name=value` syntax, but the name doesn't support using `-`.
We should support it since there are some config items like `label-property.reject-leader`, `auto-compaction` which use `-` in their names.

Introduce a new token `ConfigItemName` to support `-`.

### Check List 

Tests 

 - Unit test


SunRunAway:
Should it be cherry-picked to release-4.0 branch and update the dependency of tidb repo?

tiancaiamao:
Dash(-) is not a valid char as identifier
When used, it should be warp with `

```
`sth-sth`
```

https://dev.mysql.com/doc/refman/5.7/en/identifier-qualifiers.html

PR #823, opened 2020-04-22
parser: add auto_random table option support

AilinKid:
supply a way to rebase auto_random from outside —— by auto_random table option

support auto_random table option support
the lined TiDB PR: [#16750](https://github.com/pingcap/tidb/pull/16750)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - add auto_random to table option

Side effects

(we will add it into the comment feature ID management)

Related changes


tiancaiamao:
Is it possible that there are more than one columns declared as auto random? @AilinKid

AilinKid:
> Is it possible that there are more than one column declared as auto random? @AilinKid

No, only one specially with primary key

PR #825, opened 2020-04-22
Remove unnecessary `.(string)` casting and `$$ = $1` from parser.y

kennytm:

1. If a rule always produces a `string`, make their type `<ident>` so we don't need to do `$x.(string)`.
2. Implicitly every expression without an action gets `$$ = $1`. So we remove all such actions.

`parser.go` is slightly reduced from 4,084,324 bytes to 4,079,780 bytes. No observable performance gain or loss.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


tiancaiamao:
Good job!

> If a rule always produces a string, make their type <ident> so we don't need to do $x.(string).

This could avoid the `interface{} => string` conversion and it's beneficial for the performance.

LGTM

PR #828, opened 2020-04-24
parser: rename auto rand table option

AilinKid:
According to some team partners' advice, should better rename table option `auto_rand` to `auto_rand_base`, which means the base ID setting.

change the table option name to `auto_rand_base`

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - None

Related changes


kennytm:
why call it `AUTO_RAND_BASE` not <code>AUTO_RAND<strong>OM</strong>_BASE</code>?

AilinKid:
> why call it `AUTO_RAND_BASE` not `AUTO_RAND**OM**_BASE`?

ok

PR #836, opened 2020-04-27
Fix `set charset` statement

wjhuang2016:
`set charset` statement is not equal to `set names` statement, we need to distinguish them.
Related TiDB [PR](https://github.com/pingcap/tidb/pull/16984).

If the statement is `set charset`, we use `SetCharset` instead of `SetNames` to mark it.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


tiancaiamao:
LGTM

PR #845, opened 2020-05-02
parser: support TableOptimizerHints in InsertStmt

Yisaer:
Ref https://github.com/pingcap/tidb/issues/16929

Support `Insert /*+ OptimizerHint */  .... ` for InsertStmt.

### Check List 

Tests 

 - Unit test
Code changes

Related changes


parser.y
```diff
@@ -5275,16 +5275,19 @@ NotKeywordToken:
...
-		ts := &ast.TableSource{Source: $5.(*ast.TableName)}
+		ts := &ast.TableSource{Source: $6.(*ast.TableName)}
 		x.Table = &ast.TableRefsClause{TableRefs: &ast.Join{Left: ts}}
-		if $7 != nil {
-			x.OnDuplicate = $7.([]*ast.Assignment)
+		if $8 != nil {
+			x.OnDuplicate = $8.([]*ast.Assignment)
+		}
+		if $2 != nil {
+			x.TableHints = $2.([]*ast.TableOptimizerHint)
```
kennytm, line 5290: what about this?

```sql
insert /*+ qb_name(qb1) */ into a select /*+ qb_name(qb2) */ 1;
```

MySQL 8.0 currently treats this as the same as `/*+ qb_name(qb2) qb_name(qb1) */`.

parser.y
```diff
@@ -5303,11 +5306,21 @@ InsertValues:
...
-		$$ = &ast.InsertStmt{Columns: $2.([]*ast.ColumnName), Select: $5.(*ast.SelectStmt)}
+		x := &ast.InsertStmt{Columns: $2.([]*ast.ColumnName), Select: $5.(*ast.SelectStmt)}
+		st := $5.(*ast.SelectStmt)
+		if st.SelectStmtOpts.TableHints != nil {
+			x.TableHints = st.SelectStmtOpts.TableHints
+		}
+		$$ = x
 	}
 |	'(' ColumnNameListOpt ')' UnionStmt
 	{
```
kennytm, line 5326: also, how about this?

```sql
insert /*+ qb_name(qb1) */ into a 
    select /*+ qb_name(qb2) */ 1 union select /*+ qb_name(qb3) */ 2;
```

parser.y
Yisaer: good catch. What about ignoring `TableOptimizerHint` when `TableHints` is defined in `InsertValues`?

parser.y
Yisaer: Do you have any suggestion? I'm not sure how to solve this case properly.

kennytm: i think you should *not* lift the hint from SELECT/UNION into INSERT in the parser. leave everything separate and let TiDB handle the propagation.

Yisaer: That's a good idea. We shall only handle the insert hint in its own stmt struct.

parser.y
Yisaer: We would store the hints in each own stmtNode and solve the hints priority in TiDB.

ast/dml.go
```diff
@@ -1379,6 +1381,17 @@ func (n *InsertStmt) Restore(ctx *format.RestoreCtx) error {
 	} else {
 		ctx.WriteKeyWord("INSERT ")
 	}
+
+	if n.TableHints != nil && len(n.TableHints) != 0 {
+		ctx.WritePlain("/*+ ")
+		for i, tableHint := range n.TableHints {
+			if err := tableHint.Restore(ctx); err != nil {
+				return errors.Annotatef(err, "An error occurred while restore UpdateStmt.TableHints[%d]", i)
```
SunRunAway, line 1389:
```suggestion
				return errors.Annotatef(err, "An error occurred while restore InsertStmt.TableHints[%d]", i)
```

XuHuaiyu:
@Yisaer Should we cherry-pick this commit to release-4.0?

Yisaer:
@XuHuaiyu I think we can.

XuHuaiyu:
@Yisaer Ok, may you file a PR?

PR #850, opened 2020-05-09
model: add IsCommonHandle field on TableInfo

coocood:
To support clustered index.

https://github.com/pingcap/tidb/pull/17044


coocood:
@kennytm PTAL

Issue #874, opened 2020-05-27
Syntax error when function name includes a dot


kyleconroy:
## Bug Report

1. What did you do?

```go
package main

import (
	"fmt"

	"github.com/pingcap/parser"
	_ "github.com/pingcap/tidb/types/parser_driver"
)

const query = `
SELECT foo.bar('baz');
`

func main() {
	p := parser.New()
	stmts, warn, err := p.Parse(query, "", "")
	fmt.Println("stmt", stmts)
	fmt.Println("warn", warn)
	fmt.Println("err", err)
}
```

```
$ go run main.go
```

2. What did you expect to see?

```
stmt [0xc000270000]
warn []
err <nil>
```

I tested `SELECT foo.bar('baz');` on both MySQL 5.7 and 8. The query parsed and returned the expected "function does not exist" error.

```
mysql> select foo.bar('baz');
ERROR 1305 (42000): FUNCTION foo.bar does not exist
```

3. What did you see instead?

The query fails to parse.

```
stmt []
warn []
err line 2 column 16 near "('baz');
" 
```

4. What version of TiDB SQL Parser are you using?

My `go.mod` has the following contents:

```
go 1.13

require (
	github.com/pingcap/parser v3.1.1+incompatible
	github.com/pingcap/tidb v1.1.0-beta.0.20200426094148-df4c9f05c49c
)
```

PR #876, opened 2020-05-28
let function names like a.b() be parsed successfully

tangenta:
Fix issue #874.

Add a rule for `FunctionCallGeneric`.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below):
```sql
mysql root@127.0.0.1:test> select t.t();
(1305, 'FUNCTION test.t.t does not exist')
```

Code changes

Side effects

Related changes


parser.y
```diff
@@ -6443,6 +6443,10 @@ FunctionCallGeneric:
 	{
 		$$ = &ast.FuncCallExpr{FnName: model.NewCIStr($1), Args: $3.([]ast.ExprNode)}
 	}
+|	Identifier '.' Identifier '(' ExpressionListOpt ')'
+	{
+		$$ = &ast.FuncCallExpr{FnName: model.NewCIStr($1 + "." + $3), Args: $5.([]ast.ExprNode)}
```
kennytm, line 6448: This is inappropriate to use `$1 + "." + $3` as the function name, since each identifier itself can contain a dot. 

Counter example on MySQL:

```sql
create schema s;
use s;
create function `s.a` () returns int deterministic return 1;
create function `a` () returns int deterministic return 2;

select `s`.`a`();  -- value is 2
select `s.a`();    -- value is 1 
```

ast/functions_test.go
```diff
@@ -227,3 +227,17 @@ func (ts *testDMLSuite) TestWindowFuncExprRestore(c *C) {
...
 	RunNodeRestoreTest(c, testCases, "select %s from t", extractNodeFunc)
 }
+
+func (ts *testFunctionsSuite) TestGenericFuncRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"s.a()", "`s`.A()"},
+		{"`s`.`a`()", "`s`.A()"},
+		{"now()", "NOW()"},
+		{"`s`.`now`()", "`s`.NOW()"},
+		{"generic_func()", "GENERIC_FUNC()"},
```
tangenta, line 237: The restored SQL text is very strange because the function name is written as keyword:
```
ctx.WriteKeyWord(n.FnName.O)
```

- If we use `WriteName()`, it restores `now()` to `` `now`() ``, which is invalid in MySQL:
```
FUNCTION test.now does not exist. Check the 'Function Name Parsing and Resolution' section in the Reference Manual
```

kennytm: Perhaps we need one more field to indicate whether the function is a name or keyword (according to rules in https://dev.mysql.com/doc/refman/8.0/en/function-resolution.html)

ast/functions_test.go
```diff
@@ -227,3 +227,19 @@ func (ts *testDMLSuite) TestWindowFuncExprRestore(c *C) {
...
 }
+
+func (ts *testFunctionsSuite) TestGenericFuncRestore(c *C) {
+	testCases := []NodeRestoreTestCase{
+		{"s.a()", "`s`.`a`()"},
+		{"`s`.`a`()", "`s`.`a`()"},
+		{"now()", "NOW()"},
+		{"`s`.`now`()", "`s`.`now`()"},
+		// FIXME: expectSQL should be `generic_func()`.
+		{"generic_func()", "GENERIC_FUNC()"},
```
tangenta, line 238: To identify these functions(without a dot), the first branch of `FunctionCallGeneric` rule needs to be changed from `identifier` to `Identifier`, and use `CheckInTokenMap` to filter generic functions. 
```
FunctionCallGeneric:
	identifier '(' ExpressionListOpt ')'		identifier '(' ExpressionListOpt ')'
```
However, such change also causes a lot of shift/reduce and reduce/reduce conflicts (with FunctionCallKeyword and FunctionCallNonKeyword).

tangenta: One more thing: hundreds of test cases still need to be updated even if the conflicts are resolved.

kennytm: 🤔 That's gonna break a lot of test cases.

misc.go
```diff
@@ -133,6 +133,12 @@ func init() {
 	initTokenFunc("'\"", startString)
 }
 
+// CheckInTokenMap indicates whether the target string is contained in tokenMap.
+func CheckInTokenMap(target string) bool {
```
tiancaiamao, line 137: Is it necessary to expose a new function, will it be used outside the package?

tangenta: I forgot to unexport it after trying to use it in `FuncCallExpr.Restore()`. Updated.

tiancaiamao:
LGTM
Please update parser.go @tangenta

PR #883, opened 2020-06-02
Add more fields in DiffSchema for exchanging partition

zhaox1n:
While exchanging partition affects 2 tables, we need to expand `DiffSchema`. Related PR in TIDB [#17149](https://github.com/pingcap/tidb/pull/17149)

### Check List 

Tests 

 - No code

Side effects

 - Increased code complexity

### Release note 

 - No release note


model/ddl.go
```diff
@@ -429,4 +429,9 @@ type SchemaDiff struct {
 	OldTableID int64 `json:"old_table_id"`
 	// OldSchemaID is the schema ID before rename table, only used by rename table DDL.
 	OldSchemaID int64 `json:"old_schema_id"`
+
+	// AffectedTableID is the table ID of partition table when exchanging partition
+	AffectedTableID int64 `json:"affected_table_id"`
```
tiancaiamao, line 434: Model data is persistented, so new fields of the struct is hard to remove for compatibility reasons once introduced.

Can we reuse `OldTableID` `OldSchemID` when the ActionType is AlterTableExchangePartition ?

zhaox1n: In theory, the fields are not enough  to use . For exchanging partition, the IDs of partition and `nt` are exchanged. nt need to clear the old table id , so it used `OldTableID` and `OldSchemaID`. Meanwhile there is no field to record the schemID and TableID of pt, it needs to expand.

zimulala: > Meanwhile there is no field to record the schemID and TableID of pt, it needs to expand.

Could we use `SchemaID` and `TableID` to record the schemID and TableID of `pt`?

zhaox1n: PTAL @tiancaiamao @zimulala

model/ddl.go
```diff
@@ -429,4 +429,14 @@ type SchemaDiff struct {
 	OldTableID int64 `json:"old_table_id"`
 	// OldSchemaID is the schema ID before rename table, only used by rename table DDL.
 	OldSchemaID int64 `json:"old_schema_id"`
+
+	OtherAffectedSchema []*SchemaAffect `json:"other_affected_schema"`
```
zimulala, line 433:
```suggestion
	OtherAffectedSchemas []*SchemaAffect `json:"other_affected_schema"`
```

zimulala: I don't think the "SchemaAffect" using "Schema" as a unit is particularly good. "Schema" in "OtherAffectedSchema" and "SchemaID" is a bit difficult to understand.

model/ddl.go
```diff
@@ -429,4 +429,14 @@ type SchemaDiff struct {
 	OldTableID int64 `json:"old_table_id"`
 	// OldSchemaID is the schema ID before rename table, only used by rename table DDL.
 	OldSchemaID int64 `json:"old_schema_id"`
+
+	OtherAffectedSchemas []*SchemaAffect `json:"other_affected_schema"`
+}
+
+// SchemaAffect is used when a ddl affects multi tables.
+type SchemaAffect struct {
```
kennytm, line 437:
```suggestion
type AffectedSchema struct {
```

model/ddl.go
```diff
@@ -429,4 +429,14 @@ type SchemaDiff struct {
 	OldTableID int64 `json:"old_table_id"`
 	// OldSchemaID is the schema ID before rename table, only used by rename table DDL.
 	OldSchemaID int64 `json:"old_schema_id"`
+
+	AffectOpts []*AffectOption `json:"other_affect_options"`
+}
+
+// AffectOption is used when a ddl affects multi tables.
```
zimulala, line 436:
```suggestion
// AffectedOption is used when a ddl affects multi tables.
```

model/ddl.go
```diff
@@ -429,4 +429,14 @@ type SchemaDiff struct {
...
+type AffectOption struct {
```
zimulala, line 437:
```suggestion
type AffectedOption struct {
```

model/ddl.go
```diff
@@ -429,4 +429,14 @@ type SchemaDiff struct {
 	OldTableID int64 `json:"old_table_id"`
 	// OldSchemaID is the schema ID before rename table, only used by rename table DDL.
 	OldSchemaID int64 `json:"old_schema_id"`
+
+	AffectOpts []*AffectOption `json:"other_affect_options"`
```
zimulala, line 433:
```suggestion
	AffectedOpts []*AffectedOption `json:"other_affected_options"`
```

zimulala:
PTAL @tiancaiamao @kennytm

tiancaiamao:
LGTM

PR #890, opened 2020-06-08
model: correct the usage of json.Unmarshal in DecodeArgs

### LINKED ISSUE #889: `DecodeArgs` relies on a wrong behavior of `json.Unmarshal`
## Bug Report

In Go 1.15 (master branch), TiDB's DDL cannot work correctly and DDL related unit-test all failed.

The root cause is `Job.DecodeArgs` relies on the **wrong behavior** of `json.Unmarshal`, and this bug is fixed in Go 1.15.

The docs read:
> To unmarshal a JSON array into a slice, Unmarshal resets the slice length to zero and then appends each element to the slice.

This means the value in `args` doesn't matter, and the value in `args` will be replaced by an `interface{}(nil)`. 

https://github.com/pingcap/parser/blob/fb519914dddd049f695e582181e135f947a5d93a/model/ddl.go#L284-L288

See https://github.com/golang/go/issues/39427 for a more detailed discussion.

bobotu:
Fixed #889

Using `json.Decoder` is more complex.

```go
if len(job.RawArgs) == 0 {
	return nil
}

decoder := json.NewDecoder(bytes.NewReader(job.RawArgs))
token, err := decoder.Token()
if err != nil {
	return errors.Trace(err)
}
if del, ok := token.(json.Delim); !ok || del != '[' {
	return nil
}

var i int
for decoder.More() {
	if i == len(args) {
		break
	}
	if err := decoder.Decode(args[i]); err != nil {
		return errors.Trace(err)
	}
	i++
}
job.Args = args[:i]
return nil
```

### Check List 

Tests 
 - Run TiDB's unit-test with Go 1.15 

Side effects

### DIFF
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -287,15 +287,18 @@ func (job *Job) DecodeArgs(args ...interface{}) error {
 		return err
 	}
 
-	for i := 0; ; i++ {
-		if i == len(rawArgs) || i == len(args) {
-			job.Args = args[:i]
-			return nil
-		}
+	sz := len(rawArgs)
+	if sz > len(args) {
+		sz = len(args)
+	}
+
+	for i := 0; i < sz; i++ {
 		if err := json.Unmarshal(rawArgs[i], args[i]); err != nil {
 			return err
 		}
 	}
+	job.Args = args[:sz]
+	return nil
 }
 
 // String implements fmt.Stringer interface.
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -282,9 +282,20 @@ func (job *Job) Decode(b []byte) error {
 
 // DecodeArgs decodes job args.
 func (job *Job) DecodeArgs(args ...interface{}) error {
-	job.Args = args
-	err := json.Unmarshal(job.RawArgs, &job.Args)
-	return errors.Trace(err)
+	var rawArgs []json.RawMessage
+	if err := json.Unmarshal(job.RawArgs, &rawArgs); err != nil {
+		return err
+	}
+
+	for i := 0; ; i++ {
+		if i == len(rawArgs) || i == len(args) {
+			job.Args = args[:i]
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


model/ddl.go
```diff
@@ -282,9 +282,20 @@ func (job *Job) Decode(b []byte) error {
...
+
+	for i := 0; ; i++ {
+		if i == len(rawArgs) || i == len(args) {
+			job.Args = args[:i]
+			return nil
+		}
+		if err := json.Unmarshal(rawArgs[i], args[i]); err != nil {
+			return err
+		}
+	}
```
kennytm, line 298:
```suggestion
	if len(rawArgs) > len(args) {
		rawArgs = rawArgs[:len(args)]
	}
	for i, rawArg := range rawArgs {
		if err := json.Unmarshal(rawArg, args[i]); err != nil {
			return err
		}
	}
	job.Args = args
	return nil
```

bobotu: We need to ensure if `len(rawArgs) == 0` then `len(job.Args) == 0`.

bobotu:
@kennytm The existing test case already covered this `DecodeArgs`, I have run UT with Go 1.15 to ensure it is correct.

model/ddl.go
```diff
@@ -282,9 +282,23 @@ func (job *Job) Decode(b []byte) error {
 
 // DecodeArgs decodes job args.
 func (job *Job) DecodeArgs(args ...interface{}) error {
-	job.Args = args
-	err := json.Unmarshal(job.RawArgs, &job.Args)
-	return errors.Trace(err)
+	var rawArgs []json.RawMessage
+	if err := json.Unmarshal(job.RawArgs, &rawArgs); err != nil {
+		return err
```
wjhuang2016, line 287: How about errors.Trace(err) ？

bobotu: Done~

PR #894, opened 2020-06-10
types: make the EvalType of TypeNull as ETInt

XuHuaiyu:
fix https://github.com/pingcap/tidb/issues/16027

After https://github.com/pingcap/tidb/pull/15512, TypeNull is defined as fixedLen. But the commit does not handle all the scenarios for TypeNull such https://github.com/pingcap/tidb/blob/master/expression/chunk_executor.go#L118.
We define the EvalType for TypeNull as EvalInt because the element length of TypeNull equals to TypeLonglong.

Define the EvalType of TypeNull as ETInt.

### Check List 

Tests 

Test in tidb will cover this

Code changes

Side effects

N/A

Related changes


XuHuaiyu:
Wrong modification.

PR #901, opened 2020-06-15
parser: remove check constraint warnings in 4.0-itai

AilinKid:
remove check constraint warnings in parser.

remove check constraint warnings in parser.

### Check List 

Tests 

 - Unit test
 - Integration test


AilinKid:
Because 4.0-itai use TiDB master as the integration test branch, so there is some errors for these features which was not cherry-picked to parser-4.0-itai
```
# github.com/pingcap/tidb/infoschema [github.com/pingcap/tidb/types.test]
infoschema/builder.go:63:58: undefined: model.ActionExchangeTablePartition
```

PR #902, opened 2020-06-15
parser: remove check constraint waring in column option

AilinKid:
 remove check constraint waring in column option

 remove check constraint waring in column option

### Check List 

Tests 

 - Unit test
 - Integration test


AilinKid:
```
# github.com/pingcap/tidb/infoschema [github.com/pingcap/tidb/infoschema.test]
infoschema/builder.go:63: undefined: model.ActionExchangeTablePartition
```

PR #854, opened 2020-05-13
parser: add new agg function `APPROX_COUNT_DISTINCT`

solotzg:
Signed-off-by: Tong Zhigao <tongzhigao@pingcap.com>

### What problem does this PR solve? 

- https://github.com/pingcap/tidb/issues/14632

### What is changed and how it works?

- Add new agg function `APPROX_COUNT_DISTINCT` to compute approximate result like count distinct.

- https://github.com/pingcap/tidb/pull/17175

### Check List 

Tests 

 - Unit test

PR #906, opened 2020-06-18
parser: add new agg function `APPROX_COUNT_DISTINCT` (#854)

solotzg:
cherry-pick #854 to release-4.0

- https://github.com/pingcap/tidb/issues/14632

- Add new agg function `APPROX_COUNT_DISTINCT` to compute approximate result like count distinct.

- https://github.com/pingcap/tidb/pull/17175

### Check List 

Tests 

 - Unit test


parser.y
```diff
@@ -6384,6 +6387,10 @@ SumExpr:
 			$$ = &ast.AggregateFuncExpr{F: $1, Args: args}
 		}
 	}
+|	builtinApproxCountDistinct '(' ExpressionList ')'
+	{
+		$$ = &ast.AggregateFuncExpr{F: $1, Args: $3.([]ast.ExprNode), Distinct: false}
```
XuHuaiyu, line 6392: `Distinct: false` can be removed?

solotzg: Maybe other logic use this field.

PR #911, opened 2020-06-30
ast/functions: add tidb_decode_base64 function

xiongjiwei:

`tidb_decode_base64` use to decode base64 encoded key

refer issue https://github.com/pingcap/tidb/issues/15951

### DIFF
--- a/ast/functions.go
+++ b/ast/functions.go
@@ -328,7 +328,8 @@ const (
 	JSONLength        = "json_length"
 
 	// TiDB internal function.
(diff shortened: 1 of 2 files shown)


kennytm:
@imtbkcat could we name the function <code>tidb_decode_base64<strong>_key</strong></code>? the current function name sounds like a strange alias of [`from_base64`](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_from-base64).

xiongjiwei:
@kennytm @AilinKid PTAL

PR #912, opened 2020-07-02
parser: add function n10_to_36

lzmhhh123:
Add a function n10_to_36.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


zz-jason:
how about changing the function to `convert_decimal_to(dst_scale, decimal_value)`, only support 36 as the `dst_scale`?

kennytm:
Why do you need such function when [`CONV`](https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_conv) does the same thing? It is even supported on TiDB.

```sql
mysql> select conv(123456, 10, 36);
+----------------------+
| conv(123456, 10, 36) |
+----------------------+
| 2N9C                 |
+----------------------+
1 row in set (0.00 sec)

mysql> select conv('2n9c', 36, 10);
+----------------------+
| conv('2n9c', 36, 10) |
+----------------------+
| 123456               |
+----------------------+
1 row in set (0.00 sec)
```

PR #913, opened 2020-07-03
parser: support new syntax for extended statistics

eurekaka:

Parser changes of https://github.com/pingcap/tidb/issues/18330

Add new SQL syntax `CREATE STATISTICS` / `DROP STATISTICS` / `ADMIN RELOAD STATISTICS` support.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

N/A

Related changes


kennytm:
Current syntax:

```sql
CREATE STATISTICS [IF NOT EXISTS] statistics_name
    [ '(' CARDINALITY ')' | '(' DEPENDENCY ')' | '(' CORRELATION ')' ] 
    ON table_name '(' column_name ',' column_name [',' ...] ')'
```

For comparison, [PostgreSQL](https://www.postgresql.org/docs/current/sql-createstatistics.html) (and [CockroachDB](https://www.cockroachlabs.com/docs/stable/create-statistics.html)):

```sql
CREATE STATISTICS [IF NOT EXISTS] statistics_name
    [ '(' { NDISTINCT | DEPENDENCIES | MCV } [',' ...] ')' ]
    ON column_name ',' column_name [',' ...] 
    FROM table_name
```

[Transact-SQL](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-statistics-transact-sql)

```sql
CREATE STATISTICS statistics_name
    ON table_name '(' column_name ',' column_name [',' ...] ')'
    [ WHERE predicate ]
    [ WITH options ]
```

[SAP HANA](https://help.sap.com/viewer/4fe29514fd584807ac9f2a04f6754767/2.0.05/en-US/43a36b3923294924a0bae80fd4b370dc.html)

```sql
CREATE STATISTICS statistics_name
    ON table_name [ '(' column_name ',' column_name [',' ...] ')' ] FOR { DEFAULT | EXTENDED } STORAGE
    [ TYPE statistics_type ]
    [ statistics_properties ',' statistics_properties [',' ...] ]
    [ [ NO ] INITIAL REFRESH ]
```

So the current syntax looks like a strange mix of PostgreSQL and Transact-SQL. 

Question: 

* why not allow specifying multiple statistics?

    ```sql
    create statistics foo (cardinality, correlation) on tbl (c1, c2);
    ```

eurekaka:
@kennytm the idea of extended statistics is mainly borrowed from PostgreSQL, but we plan to support different types of statistics, i.e, correlation. The definition of table and columns, i.e, `ON t(a,b,c)` is consistent with MySQL syntax style, e.g, `create index idx on t(a,b,c)`.

eurekaka:
CI failure should be fixed after https://github.com/pingcap/tidb/pull/18787 is merged.

parser_test.go
```diff
@@ -704,6 +704,7 @@ func (s *testParserSuite) TestDMLStmt(c *C) {
 		{"admin reload bindings", true, "ADMIN RELOAD BINDINGS"},
 		{"admin show telemetry", true, "ADMIN SHOW TELEMETRY"},
 		{"admin reset telemetry_id", true, "ADMIN RESET TELEMETRY_ID"},
+		{"admin reload statistics", true, "ADMIN RELOAD STATISTICS"},
```
qw4990, line 707: Is this statement used to update extended statistics?
Why not use a similar syntax to `analyze ...`, for example, `analyze statistics stat1`, `analyze statistics table1`?

eurekaka: This is used to reload the stats cache for extended statistics from `mysql.stats_extended`, normally this command is not needed, it is a remedial approach provided for DBA to guarantee the stats cache is consistent with records in `mysql.stats_extended`.

qw4990: OK

PR #915, opened 2020-07-05
Support global index creation

zhaox1n:
- Add `Global` option in IndexInfo
- Support  global index creation in parser

```sql
CREATE INDEX idx ON t ( a ) GLOBAL;
create table t1 (a int , b int , unique index  unq (b) global) partition by hash(a) partitions 2;
```

### Check List 

Tests 

 - Unit test

Related changes

 - None


zhaox1n:
@tiancaiamao PTAL

PR #916, opened 2020-07-06
parser: support set operator EXCEPT and INTERSECT

lzmhhh123:
refer to [sql-server EXCEPT and INTERSECT](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-except-and-intersect-transact-sql?view=sql-server-ver15), we decide to support extra set operators `except` and `intersect`, similar to `union`.

### Check List 

Tests 

 - Unit test
 - Integration test (fail now

Code changes

Side effects

 - Increased code complexity

Related changes


kennytm:
(Probably more appropriate to refer to the [MariaDB docs](https://mariadb.com/kb/en/intersect/), which at least share a closer dialect with us)

kennytm:
Both SQL server and MariaDB specify that INTERSECT has higher precedence than UNION and EXCEPT. This PR currently generates a flat list, which is fine, but you need to ensure TiDB recognizes (A &cap; B &cup; C &cap; D) == ((A &cap; B) &cup; (C &cap; D)).

> **Transact-SQL**
> 
> If EXCEPT or INTERSECT is used together with other operators in an expression, it's evaluated in the context of the following precedence:
> 1. Expressions in parentheses
> 2. The INTERSECT operator
> 3. EXCEPT and UNION evaluated from left to right based on their position in the expression



> **MariaDB**
>
> `INTERSECT` has higher precedence than `UNION` and `EXCEPT`. If possible it will be executed linearly but if not it will be translated to a subquery in the `FROM` clause:
>
> ```sql
> (select a,b from t1)
> union
> (select c,d from t2)
> intersect
> (select e,f from t3)
> union
> (select 4,4);
> ```
>
> will be translated to:
>
> ```sql
> (select a,b from t1)
> union
> select c,d from
>   ((select c,d from t2)
>    intersect
>    (select e,f from t3)) dummy_subselect
> union
> (select 4,4)
> ```

lzmhhh123:
> Both SQL server and MariaDB specify that INTERSECT has higher precedence than UNION and EXCEPT. This PR currently generates a flat list, which is fine, but you need to ensure TiDB recognizes (A ∩ B ∪ C ∩ D) == ((A ∩ B) ∪ (C ∩ D)).

Yeah, you are right. I'll make the priority correct in TiDB.

ast/dml.go
```diff
@@ -1067,52 +1073,64 @@ func (n *UnionSelectList) Accept(v Visitor) (Node, bool) {
...
 // See https://dev.mysql.com/doc/refman/5.7/en/union.html
-type UnionStmt struct {
+// See https://mariadb.com/kb/en/intersect/
+// See https://mariadb.com/kb/en/except/
+type SetOprStmt struct {
 	dmlNode
 	resultSetNode
 
-	SelectList *UnionSelectList
+	OptType    SetOprType
```
kennytm, line 1093:
```suggestion
	OprType    SetOprType
```
?

IMO it's clearer to spell it out as `SetOperator` than `SetOprType`.

parser_test.go
```diff
@@ -3818,6 +3818,36 @@ func (s *testParserSuite) TestUnion(c *C) {
...
+		{"(select c1 from t1) intersect (select c2 from t2) order by c1 intersect select c3 from t3", false, ""},
+		{"(select c1 from t1) intersect (select c2 from t2) limit 1 intersect select c3 from t3", false, ""},
+		{"(select c1 from t1) intersect select c2 from t2 intersect (select c3 from t3) order by c1 limit 1", true, "(SELECT `c1` FROM `t1`) INTERSECT SELECT `c2` FROM `t2` INTERSECT (SELECT `c3` FROM `t3`) ORDER BY `c1` LIMIT 1"},
+		{"select (select 1 intersect select 1) as a", true, "SELECT (SELECT 1 INTERSECT SELECT 1) AS `a`"},
+		{"select * from (select 1 intersect select 2) as a", true, "SELECT * FROM (SELECT 1 INTERSECT SELECT 2) AS `a`"},
+		{"insert into t select c1 from t1 intersect select c2 from t2", true, "INSERT INTO `t` SELECT `c1` FROM `t1` INTERSECT SELECT `c2` FROM `t2`"},
+		{"insert into t (c) select c1 from t1 intersect select c2 from t2", true, "INSERT INTO `t` (`c`) SELECT `c1` FROM `t1` INTERSECT SELECT `c2` FROM `t2`"},
+		{"select 2 as a from dual intersect select 1 as b from dual order by a", true, "SELECT 2 AS `a` INTERSECT SELECT 1 AS `b` ORDER BY `a`"},
```
kennytm, line 3850: please add a test case involving a mixture of `intersect`, `except`, `union` and `union all`

lzmhhh123: Done.

misc.go
```diff
@@ -377,6 +377,7 @@ var tokenMap = map[string]int{
 	"INT8":                     int8Type,
 	"INTEGER":                  integerType,
 	"INTERNAL":                 internal,
+	"INTERSECT":                intersect,
```
fzhedu, line 380: do `except` need to be inserted here?

parser_test.go
```diff
@@ -3818,6 +3819,43 @@ func (s *testParserSuite) TestUnion(c *C) {
 		{"insert into t select c1 from t1 union select c2 from t2", true, "INSERT INTO `t` SELECT `c1` FROM `t1` UNION SELECT `c2` FROM `t2`"},
 		{"insert into t (c) select c1 from t1 union select c2 from t2", true, "INSERT INTO `t` (`c`) SELECT `c1` FROM `t1` UNION SELECT `c2` FROM `t2`"},
 		{"select 2 as a from dual union select 1 as b from dual order by a", true, "SELECT 2 AS `a` UNION SELECT 1 AS `b` ORDER BY `a`"},
+		// except
+		{"select c1 from t1 except select c2 from t2", true, "SELECT `c1` FROM `t1` EXCEPT SELECT `c2` FROM `t2`"},
```
fzhedu, line 3823: would `except/union/inserset` be nested in subqueries?

parser.y
```diff
@@ -7825,109 +7827,111 @@ SelectLockOpt:
...
 		}
-		$$ = union
+		$$ = setOpr
 	}
-|	UnionClauseList "UNION" UnionOpt SelectStmtFromTable OrderByOptional SelectStmtLimit SelectLockOpt
+|	SetOprClauseList SetOpr SelectStmtFromTable OrderByOptional SelectStmtLimit SelectLockOpt
 	{
-		st := $4.(*ast.SelectStmt)
-		union := $1.(*ast.UnionStmt)
-		st.IsAfterUnionDistinct = $3.(bool)
```
fzhedu, line 7873: can we ignore this?

parser_test.go
lzmhhh123: It a type of data source. Could be nested in subqueries.

misc.go
lzmhhh123: `except` already exists here.

parser.y
lzmhhh123: I used the `afterSetOprType` instead of it.

lzmhhh123:
@fzhedu PTAL again~

PR #927, opened 2020-07-13
parser: add a state field for partitionDefinition

AilinKid:
To Close Issue https://github.com/pingcap/tidb/issues/18496

Related TiDB PR: https://github.com/pingcap/tidb/pull/18495

add state field for partition difinition.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


AilinKid:
PTAL @tiancaiamao @tangenta

model/model.go
```diff
@@ -719,10 +719,30 @@ func (pi *PartitionInfo) GetNameByID(id int64) string {
...
-	Comment  string   `json:"comment,omitempty"`
+	ID       int64       `json:"id"`
+	Name     CIStr       `json:"name"`
+	LessThan []string    `json:"less_than"`
+	Comment  string      `json:"comment,omitempty"`
+	State    SchemaState `json:"state"`
+}
+
+// Clone clones ConstraintInfo.
+func (ci *PartitionDefinition) Clone() PartitionDefinition {
```
zimulala, line 730: It seems that we don't use it. So I think we can remove it.

PR #930, opened 2020-07-15
terror: add terror api to support add and product error workaround automaticly

imtbkcat:
TiDB has a lot of error with error code, writing workaround toml file manual will be a huge work. 
So we need an api which can add workaround easily and produce toml file by code.

In TiDB, error registered in this form:
`errAccessDenied  = terror.ClassServer.New(errno.ErrAccessDenied, errno.MySQLErrName[errno.ErrAccessDenied])`

To add workaround, just append behind the `New` function.
`errAccessDenied = terror.ClassServer.New(errno.ErrAccessDenied, errno.MySQLErrName[errno.ErrAccessDenied]).SetWorkaround("this may caused by incorrect password")`

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes


terror/terror.go
```diff
@@ -363,3 +376,34 @@ func Log(err error) {
...
+		}
+		_, err = file.WriteString(errorMsg)
+		if err != nil {
+			return err
+		}
+		_, err = file.WriteString(description)
+		if err != nil {
+			return err
+		}
+		_, err = file.WriteString(workaround)
```
kennytm, line 403:
```suggestion
		_, err := fmt.Fprintf(
			"[error.%v]\nerror = '''%v'''\ndescription = '''%v'''\nworkaround = '''%v'''\n\n", 
			code, e.message, e.description, e.workaround)
```

terror/terror.go
```diff
@@ -179,6 +183,15 @@ func (e *Error) Code() ErrCode {
 	return e.code
 }
 
+// SetWorkaround is a decorator like method which add a workaround to
+// error which is convenient for user to search.
+func (e *Error) SetWorkaround(workaround, description string) *Error {
```
SunRunAway, line 188: What does `description` mean? Could you give an example usage of this method?

imtbkcat: error messge is enough, I will remove it.

imtbkcat: Example is like:
`ErrUnknown = terror.ClassOptimizer.New(mysql.ErrUnknown, mysql.MySQLErrName[mysql.ErrUnknown]).SetWorkaround("here is workaround")`

bb7133: IMHO it's better to move setting `description` to another `SetDescription` just like `SetWorkaround`, so that developers can choose if they want to set the `description` explicitly.

terror/terror.go
```diff
@@ -179,6 +183,14 @@ func (e *Error) Code() ErrCode {
 	return e.code
 }
 
+// SetWorkaround is a decorator like method which add a workaround to
+// error which is convenient for user to search.
+func (e *Error) SetWorkaround(workaround string) *Error {
+	e.workaround = workaround
+	errCodeMap[e.code] = e
```
SunRunAway, line 190: If some `Error` does not have been called `SetWorkaround` then it can not be exported by `ExportErrorCodeAndWorkaround`. Is it by design?

imtbkcat: Yes, but I find  this is a bad design, we can't find whether we lose some error workaround. How about add this code into `New` function?

cfzjywxk: We could add errors into `errCodeMap` in the `new` function so there is no missing, `SetWorkaround` could be an option for refactoring work.

imtbkcat:
@SunRunAway @cfzjywxk @bb7133  all comments have been addressed, PTAL

terror/terror.go
```diff
@@ -161,12 +165,14 @@ func (ec ErrClass) Synthesize(code ErrCode, message string) *Error {
...
-	code    ErrCode
-	message string
-	args    []interface{}
-	file    string
-	line    int
+	class       ErrClass
+	code        ErrCode
+	message     string
+	workaround  string
+	description string
```
SunRunAway, line 172:
```suggestion
```

imtbkcat:
@SunRunAway  addressed.

kennytm:
Please fix gofmt.

```
gofmt (simplify)
terror/terror.go
make: *** [Makefile:25: fmt] Error 1
```

bb7133:
LGTM

PR #931, opened 2020-07-15
parser: more meaningful words for placement rules

xhebox:

Follow https://github.com/pingcap/tidb/issues/18030#issuecomment-654943694.

However, "constraints" is used for "label". Because it is not necessarily related to region/locality.

### Check List

Tests

 - Unit test
 - Integration test

Code changes

Related changes


parser.y
```diff
@@ -1382,7 +1383,7 @@ PlacementCountOpt:
 	}
 
 PlacementLabelOpt:
-	"LABEL" "=" stringLit
+	"CONSTRAINTS" "=" stringLit
```
kennytm, line 1386: Question: what if we use `EqOpt` instead of `'='` here 🤔 

```sql
ALTER TABLE t ALTER PARTITION p ADD PLACEMENT POLICY CONSTRAINTS '+zone=zone1' ROLE LEADER REPLICAS 1
```

xhebox: It should be fine. As that `PlacementSpec` must start with one of unreserved keywords `CONSTRAINTS`, `ROLE` or `REPLICAS`... Empty should not have an effect. I just had a test with the above statement, and it is passed.
But I do not know if I should use `EqOpt` instead of `"="`.

parser.y
```diff
@@ -1449,9 +1450,9 @@ PlacementOpts:
 	}
 
 PlacementSpec:
-	"ADD" "PLACEMENT" PlacementOpts
+	"ADD" "PLACEMENT" "POLICY" PlacementOpts
```
zimulala, line 1453: Could we remove `POLICY`? I think `ADD PLACEMENT CONSTRAINTS...` is OK. I think this statement is relatively long, if not necessary, we can remove this field.

xhebox: @djshow832 what do you think? You have not updated the RFC doc yet... For me, both makes sense.

parser.y
xhebox: After some thoughts, I will stick to `"="`. `=` can be a separator, visually, which can help you read the SQL statement.

parser.y
djshow832: You can discuss it under the issue https://github.com/pingcap/tidb/issues/18030 @zimulala

zimulala:
@xhebox 
Please fix the conflicts.

parser.y
zimulala: Fine.

xhebox:
@zimulala Resolved.

nolouch:
Should we update the [RPC documents](https://github.com/pingcap/tidb/pull/18199)? cc @djshow832

xhebox:
@nolouch RFC documents are slower than my contribution progress until today. The syntax just changed again.

Still, there are lots of things not decided. And there may be more problems on RFC documents when I am trying to implement this project.

The conclusion is that just comment any inconsistency or problem on documents...

nolouch:
@xhebox got it, thanks.

PR #939, opened 2020-07-24
parser: create integer type field with specified length should have warings

AilinKid:
before
```
mysql> create table t(a bigint(2));
```

after 
```
mysql> create table t(a bigint(2));
Query OK, 0 rows affected, 1 warning (0.01 sec)

mysql> show warnings;
+---------+------+------------------------------------------------------------------------------+
| Level   | Code | Message                                                                      |
+---------+------+------------------------------------------------------------------------------+
| Warning | 1681 | Integer display width is deprecated and will be removed in a future release. |
+---------+------+------------------------------------------------------------------------------+
1 row in set (0.00 sec)


```

This is PR is part of the functionality above, it need TiDB PR to check the same warning and rewrite the error code and message.

### Check List 

Tests 

 - Integration test


parser.y
```diff
@@ -9634,6 +9634,10 @@ NumericType:
 		// TODO: check flen 0
 		x := types.NewFieldType($1.(byte))
 		x.Flen = $2.(int)
+		if $2.(int) != types.UnspecifiedLength {
+			yylex.AppendError(yylex.Errorf("Integer display width is deprecated and will be removed in a future release."))
```
zimulala, line 9638: Could we use `ErrWarnDeprecatedIntegerDisplayWidth ` directly?

AilinKid: it can't use a `terror` class directly in scanner here,  `ErrWarnDeprecatedIntegerDisplayWidth` is used in TiDB to append warnings in statement context to replace the raw warning, it contains the error code 1681, error message. 
the raw message is like: "You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use near ''%s" Integer display width is deprecated and will be removed in a future release" and the fault error code is 1064

AilinKid:
Cause ast.UnionStmt is removed in newer parser, but TiDB still referred the old version parser. That's why there is integation error.

PR #946, opened 2020-07-27
parser: compatible with `select...for share` statement

### LINKED ISSUE #945: Incompatible for select ... for share statements.
## Bug Report

1. What did you do?

```sql
mysql> CREATE TABLE t (i INT, PRIMARY KEY (i)) ENGINE = InnoDB;
mysql> INSERT INTO t (i) VALUES(1),(2),(3);
mysql> SELECT * FROM test.t FOR SHARE;
```

```
+---+
| i |
+---+
| 1 |
+---+
| 2 |
+---+
| 3 |
+---+
```
3. What did you see instead?

```
Error Code: 1064. You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 30 near "SHARE" 
```

    `v0.0.0-20200623082809-b74301ac298b`

In MySQL 8.0, `SELECT ... LOCK IN SHARE` has been replaced by `SELECT ... FOR SHARE`.

cncal:

Close #945.

Update `parser.y`.

### Check List 

Tests 

 - Unit test
 - Integration test


kennytm:
Thanks for the PR!

Since `FOR SHARE` and `LOCK IN SHARE MODE` are [equivalent](https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html), I think they should be parsed into the same enum value (remove `SelectLockInShareMode` from the `iota` block, and make its value same as `SelectLockForShare` for backward compatibility). Similarly, the restored text of both `FOR SHARE` and `LOCK IN SHARE MODE` can be the same.

Additionally, `FOR SHARE` can be followed by `NOWAIT` and `SKIP LOCKED` like `FOR UPDATE`, should we also support them in this PR?

cncal:
Can't agree more and I'll try it.

cncal:
PTAL @kennytm.

nullnotnil:
When this merges, it will likely break https://github.com/pingcap/tidb/pull/19506

(Currently TiDB parses but ignores `LOCK IN SHARE MODE`. It will need to do the same with `FOR SHARE`).

ast/dml.go
```diff
@@ -467,24 +468,27 @@ type SelectLockInfo struct {
 }
 
 // String implements fmt.Stringer.
-func (n SelectLockType) String() string {
+func (n SelectLockType) String() (string, error) {
```
kennytm, line 471: Returning the error makes it not implement Stringer anymore. Please keep returning string only.

parser_test.go
```diff
@@ -54,7 +54,7 @@ func (s *testParserSuite) TestSimple(c *C) {
 		"current_timestamp", "current_user", "database", "databases", "day_hour", "day_microsecond",
 		"day_minute", "day_second", "decimal", "default", "delete", "desc", "describe",
 		"distinct", "distinctRow", "div", "double", "drop", "dual", "else", "enclosed", "escaped",
-		"exists", "explain", "false", "float", "fetch", "for", "force", "foreign", "from",
+		"exists", "explain", "false", "float", "for", "force", "foreign", "from",
```
kennytm, line 57: put the `fetch` back

```suggestion
		"exists", "explain", "false", "float", "fetch", "for", "force", "foreign", "from",
```

parser_test.go
```diff
@@ -746,7 +752,6 @@ func (s *testParserSuite) TestDMLStmt(c *C) {
 		{"admin reload bindings", true, "ADMIN RELOAD BINDINGS"},
 		{"admin show telemetry", true, "ADMIN SHOW TELEMETRY"},
 		{"admin reset telemetry_id", true, "ADMIN RESET TELEMETRY_ID"},
-		{"admin reload statistics", true, "ADMIN RELOAD STATISTICS"},
```
kennytm, line 749: keep this test case

kennytm (commented):
rest LGTM

PTAL @tangenta

parser_test.go
cncal: Sorry, I must have made a mistake when merging master.

tangenta (approved):
LGTM

ti-srebot (dismissed):

tangenta:
@cncal Could you regenerate `parser.go` to help merge this?

cncal:
@tangenta ok, I'll do it later.

kennytm:
ping @cncal

cncal:
@kennytm conflicts resolved.

PR #947, opened 2020-07-28
parser: substitute `state` field with `addingDefinition` in partitionInfo

AilinKid:
Linked with https://github.com/pingcap/tidb/pull/18495

When we add state field in partitionInfo, there is much to do with checking the partition visiability in DML, SQL Planer and so on.

While we can use additional field `addingDefinition` to store the mid state partition definition, it will be finnally to moved to normal partition definitions when it becomes public.  With that, there is no burden for adjustment in DML, SQL Planner and so on.

Use additional field `addingDefinition` to sustitute `state` field in partitionInfo.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


AilinKid:
the integration error comes from the newer commit remove of `TypeDecimal`

model/model.go
```diff
@@ -36,6 +36,8 @@ const (
 	// StateWriteOnly means we can use any write operation on this schema element,
 	// but outer can't read the changed data.
 	StateWriteOnly
+	// StateReplica means this schema element is waiting tiflash replica to be finished.
```
djshow832, line 39:
```suggestion
	// StateReplica means we're waiting tiflash replica to be finished.
```

PR #948, opened 2020-07-28
quickstart.md: get dependency with git hash instead of version tag

tangenta:

Using a version tag to `go get` the dependency is not the correct way. 

```console
go get github.com/pingcap/parser@v4.0.0-rc.1: github.com/pingcap/parser@v4.0.0-rc.1: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v4
```

Under ideal circumstances, the path should be something like 

```
github.com/pingcap/parser/v4@master
``` 

when the new major version is not backward compatible. However, this is still under discussion due to potential compatibility problems.

As a workaround, we can specify the git hash manually.

See the changes.

### Check List 

Tests 

N/A

Code changes

N/A

Side effects

N/A

Related changes

N/A

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -17,10 +17,10 @@ go mod init colx && touch main.go
 
 ## Import Dependencies
 
-First of all, you need to use `go get` to fetch the dependencies:
+First of all, you need to use `go get` to fetch the dependencies(for v4.0.2, the git hash is `3a18f1e`):
 
 ```bash
-go get -v github.com/pingcap/parser@v4.0.0-rc.1
+go get -v github.com/pingcap/parser@3a18f1e
 ```
 
 > **NOTE**
@@ -28,7 +28,7 @@ go get -v github.com/pingcap/parser@v4.0.0-rc.1
 > You may want to use advanced API on expressions (a kind of AST node), such as numbers, string literals, booleans, nulls, etc. It is strongly recommended to use the `types` package in TiDB repo with the following command:
 >
 > ```bash
-> go get -v github.com/pingcap/tidb/types/parser_driver@v4.0.0-rc.1
+> go get -v github.com/pingcap/tidb/types/parser_driver@328b6d0
 > ```
 > and import it in your golang source code:
 > ```go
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -17,18 +17,18 @@ go mod init colx && touch main.go
 
 ## Import Dependencies
 
-First of all, you need to use `go get` to fetch the dependencies:
+First of all, you need to use `go get` to fetch the dependencies(for v4.0.2, the git hash is `3a18f1e`):
```
djshow832, line 20: What about give the approach of fetching the git hash?

tangenta: Done.

docs/quickstart.md
```diff
@@ -17,18 +17,18 @@ go mod init colx && touch main.go
 
 ## Import Dependencies
 
-First of all, you need to use `go get` to fetch the dependencies:
+First of all, you need to use `go get` to fetch the dependencies through git hash. The git hashes are available in [release page](https://github.com/pingcap/parser/releases). Take `release-4.0.2` as an example:
```
kennytm, line 20:
```suggestion
First of all, you need to use `go get` to fetch the dependencies through git hash. The git hashes are available in [release page](https://github.com/pingcap/parser/releases). Take `v4.0.2` as an example:
```

docs/quickstart.md
```diff
@@ -17,18 +17,18 @@ go mod init colx && touch main.go
...
+go get -v github.com/pingcap/parser@3a18f1e
 ```
 
 > **NOTE**
 >
 > You may want to use advanced API on expressions (a kind of AST node), such as numbers, string literals, booleans, nulls, etc. It is strongly recommended to use the `types` package in TiDB repo with the following command:
 >
 > ```bash
-> go get -v github.com/pingcap/tidb/types/parser_driver@v4.0.0-rc.1
+> go get -v github.com/pingcap/tidb/types/parser_driver@328b6d0
```
kennytm, line 31: should this be consistent with line 23 (3a18f1e)?

tangenta: This is TiDB v4.0.2.

https://github.com/pingcap/tidb/releases/tag/v4.0.2

PR #952, opened 2020-07-30
parser: append replica-only state to previous one strictly

AilinKid:

The new added replica-only state should strictly be appended to the last one.
Otherwise, it will cause a rolling-update failure.
Referring issue: https://github.com/pingcap/tidb/issues/18873

append replica-only state to previous one strictly

### Check List 

Tests 

 - Integration test

```
Manual test like what the issue described:
1: starting a TiDB (v4.0.0) server.
2: rolling update it to the TiDB master (commit 4170007fb60f3ad1ebb11027bbb6ac97c408929d).
3: checking this process without any error mentioned in the issue above.
```

Related changes

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -45,8 +45,7 @@ const (
 	// StateReplica means we're waiting tiflash replica to be finished.
 	StateReplicaOnly
 	/*
-	 * New states must be appended after the last one, since the value is assigned according
-	 * to the strict position. Otherwise it will cause rolling update failure.
+	 *  keep the values consistent across versions.
 	 */
 )
 
--- a/model/model.go
+++ b/model/model.go
@@ -45,7 +45,7 @@ const (
 	// StateReplica means we're waiting tiflash replica to be finished.
 	StateReplicaOnly
 	/*
-	 * Follower adding state must be appended to the last one, since the value is assigned according
+	 * New states must be appended after the last one, since the value is assigned according
 	 * to the strict position. Otherwise it will cause rolling update failure.
 	 */
 )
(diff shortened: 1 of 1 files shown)


model/model.go
```diff
@@ -36,14 +36,18 @@ const (
...
 	// StateWriteReorganization means we are re-organizing whole data after write only state.
 	StateWriteReorganization
 	// StateDeleteReorganization means we are re-organizing whole data after delete only state.
 	StateDeleteReorganization
 	// StatePublic means this schema element is ok for all write and read operations.
 	StatePublic
+	// StateReplica means we're waiting tiflash replica to be finished.
+	StateReplicaOnly
+	/*
+	 * Follower adding state must be appended to the last one, since the value is assigned according
```
djshow832, line 48:
```suggestion
	 * New states must be appended after the last one, since the value is assigned according
```

model/model.go
```diff
@@ -36,14 +36,17 @@ const (
...
 	// StateWriteReorganization means we are re-organizing whole data after write only state.
 	StateWriteReorganization
 	// StateDeleteReorganization means we are re-organizing whole data after delete only state.
 	StateDeleteReorganization
 	// StatePublic means this schema element is ok for all write and read operations.
 	StatePublic
+	// StateReplica means we're waiting tiflash replica to be finished.
+	StateReplicaOnly
+	/*
+	 *  keep the values consistent across versions.
```
tangenta, line 48:
```suggestion
	 *  Please add the new state at the end to keep the values consistent across versions.
```

PR #954, opened 2020-07-30
correct calculation of field lengths for Enum and Set types

qw4990:
Related to https://github.com/pingcap/tidb/issues/18870.

Correct calculation of field lengths for Enum and Set types

Now, when creating tables, field lengths for Enum and Set type columns are set to -1, which is not correct.

This PR fixes their field length as rules below:
1. enum_flen = max(element_flen)
2. set_flen = sum(element_flen) + number_of_elements - 1

These rules are from MySQL.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/types/field_type_test.go
+++ b/types/field_type_test.go
@@ -241,12 +241,38 @@ func (s *testFieldTypeSuite) TestHasCharsetFromStmt(c *C) {
 	}
 }
 
-func (s *testFieldTypeSuite) TestEnumFlen(c *C) {
+func (s *testFieldTypeSuite) TestEnumSetFlen(c *C) {
 	p := parser.New()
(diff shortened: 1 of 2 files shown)


types/field_type_test.go
```diff
@@ -241,6 +241,40 @@ func (s *testFieldTypeSuite) TestHasCharsetFromStmt(c *C) {
 	}
 }
 
+func (s *testFieldTypeSuite) TestEnumSetFlen(c *C) {
+	p := parser.New()
+	// enum_flen = max(ele_flen)
```
wjhuang2016, line 246: Please put the comments in the parser.

qw4990:
TiDB PR https://github.com/pingcap/tidb/pull/18897

types/field_type_test.go
qw4990: updated.

PR #955, opened 2020-07-31
correct calculation of field lengths for Enum and Set types

qw4990:
Cherry pick for https://github.com/pingcap/parser/pull/954.

---

Related to https://github.com/pingcap/tidb/issues/18870.

Correct calculation of field lengths for Enum and Set types

Now, when creating tables, field lengths for Enum and Set type columns are set to -1, which is not correct.

This PR fixes their field length as rules below:
1. enum_flen = max(element_flen)
2. set_flen = sum(element_flen) + number_of_elements - 1

These rules are from MySQL.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -10033,6 +10033,13 @@ StringType:
 	{
 		x := types.NewFieldType(mysql.TypeEnum)
 		x.Elems = $3.([]string)
+		fieldLen := -1 // enum_flen = max(ele_flen)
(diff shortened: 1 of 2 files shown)


qw4990:
TiDB PR https://github.com/pingcap/tidb/pull/18909

qw4990:
PTAL @kennytm

PR #961, opened 2020-08-04
parser: support `DROP PLACEMENT` clause

xhebox:

It's a part of project "Placement rules in SQL". Support `DROP PLACEMENT` syntax according to the RFC document.

### Check List

Tests

 - Unit test
 - Integration test

Code changes

Related changes


parser.y
```diff
@@ -1187,6 +1187,7 @@ import (
 	PlacementCountOpt                      "Placement rules count option"
 	PlacementLabelOpt                      "Placement rules label option"
 	PlacementRoleOpt                       "Placement rules role option"
```
tangenta, line 1189: The 'opt' means 'optional'... not table 'option'.

Please check other rules that end with 'Opt': `AnalyzeOptionListOpt`, `AlterTableSpecListOpt`, etc.

Could you file another PR to correct these names?

kennytm: OTOH we've got things like `CreateViewSelectOpt` and `IndexNameAndTypeOpt`...

PR #965, opened 2020-08-05
*: remove redundant checks

xhebox:

Remove check on count and constraints option. Because the syntax becomes complex, and TiDB side already does a lot of checks. Especially, empty constraints are allowed in alter clauses now. It is better to leave all these works to TiDB. 

Also fix related tests, including changing meaningful `+zone=xx` into meaningless `str`. Since the parser does not check the contents anymore.

### Check List

Tests

 - Unit test
 - Integration test


parser.y
```diff
@@ -1434,7 +1424,7 @@ PlacementOptions:
 	{
 		spec := $1.(*ast.PlacementSpec)
 		if spec.Replicas > 0 {
-			yylex.AppendError(yylex.Errorf("Duplicate placement option REPLICAS"))
+			yylex.AppendError(yylex.Errorf("Duplicated placement option REPLICAS"))
```
djshow832, line 1427:
```suggestion
			yylex.AppendError(yylex.Errorf("Duplicate placement option REPLICAS"))
```
Duplicate is also an adj.

xhebox: reverted.

PR #966, opened 2020-08-06
model: add ChangeStateInfo to ColumnInfo

zimulala:
This PR is used to support column type changing.

Add ChangeStateInfo to ColumnInfo.


model/model.go
```diff
@@ -96,7 +102,8 @@ type ColumnInfo struct {
 	State               SchemaState `json:"state"`
 	Comment             string      `json:"comment"`
 	// A hidden column is used internally(expression index) and are not accessible by users.
-	Hidden bool `json:"hidden"`
+	Hidden           bool `json:"hidden"`
+	*ChangeStateInfo `json: "change_state_info"`
```
kennytm, line 106:
```suggestion
	*ChangeStateInfo `json:"change_state_info"`
```

zimulala: Done.

zimulala:
PTAL @kennytm @tangenta

zimulala:
PTAL @AilinKid  @tangenta

PR #967, opened 2020-08-10
*: forbid `REPLICAS = 0`

xhebox:

Forbid `REPLICAS = 0`.

See https://github.com/pingcap/tidb/pull/19065#discussion_r467671565.

### Check List

Tests

 - Unit test
 - Integration test


parser.y
```diff
@@ -1383,7 +1384,12 @@ PlacementRole:
 PlacementCount:
 	"REPLICAS" "=" LengthNum
 	{
-		$$ = $3
+		cnt := $3.(uint64)
+		if cnt == math.MaxUint64 {
+			yylex.AppendError(yylex.Errorf("Invalid placement option REPLICAS, it is not allowed to be MaxUint64"))
+			return 1
+		}
+		$$ = cnt
```
djshow832, line 1392: How about 0 stands for no declaration, and forbid it to be set to 0?

xhebox: Done

PR #969, opened 2020-08-11
parser.y: add create binding for digest syntax for improving SPM

iosmanthus:

This pull request adds `create binding for digest xxx with hints xxx` to enable create a `SQL` plan binding with only its normalized version's digest and `SQL` hints.

Add a new rule for `CreateBindingStmt` and distinguish each branch using `BindingType` enum.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


zz-jason:
how to find the original sql according to the digest?

iosmanthus:
@zz-jason In TiDB, we store binding via a `cache` struct in the following manner:
```
Normalized SQL Hash => []Binding { ... }
```
The `Binding` struct contains the original SQL and database name which is used to check the hash confliction but it unlikely happen while the digest is a `sha256` one. We could just ignore the confliction and use the digest directly to seek/update the bindings. But I am still working on it because in a **global** binding it needs to flush to store first via some internal `SQL` execution but in the end, it always updates to the `cache` map.

zz-jason:
then how to get the target digest?

iosmanthus:
@zz-jason  The new syntax looks like: `CREATE BINDING FOR DIGEST 'some_hash' WITH HINTS /*+ ... */
The hints are extracted from the hint comment above and the digest are extracted from 'some_hash' part.

zz-jason:
> @zz-jason The new syntax looks like: `CREATE BINDING FOR DIGEST 'some_hash' WITH HINTS /*+ ... */`. The hints are extracted from the hint comment above and the digest is extracted from `'some_hash'` part.

I know, the question is, as a user, how do I know the digest of a SQL?

iosmanthus:
@tangenta PTAL

parser.y
```diff
@@ -11154,21 +11156,36 @@ CreateBindingStmt:
...
 		x := &ast.CreateBindingStmt{
+			BindingTp:   ast.BindingForStmt,
 			OriginNode:  setOprStmt,
 			HintedNode:  hintedSetOprStmt,
 			GlobalScope: $2.(bool),
 		}
 
+		$$ = x
+	}
+|	"CREATE" GlobalScope "BINDING" "FOR" "DIGEST" stringLit "USING" hintComment
```
zz-jason, line 11175: can uses create a biding with:

```sql
create binding for
	select ...
using
	/*+ USE_INDEX(@`sel_1` `mysql`.`t` ) */
```

breeswish: This is a good idea to complement the syntax. However it is not a required feature to make it available in UI. I guess SQL binding may be less frequently used when there is a UI for it. Maybe we can defer it until necessary?

breeswish (approved):

breeswish:
@crazycs520 @kennytm Please help take a look when you are free. Thanks!

kennytm (commented):
rest lgtm

iosmanthus:
@kennytm @crazycs520 @breeswish @tangenta PTAL

breeswish:
@iosmanthus Please fix CI

zz-jason:
I'm going to close this PR since it hasn't been updated for a long time. Feel free to reopen it once you wish to advance this PR in the future.

PR #974, opened 2020-08-14
[DNM] Add reference id for drop column with composite index

blacktear23:
https://github.com/pingcap/tidb/issues/19196
If we drop column with composite index, we should create a temporary composite index without dropped column. The temporary index's name is origin index name append with a randomized string, so we need a field to track which origin index is related to this new temporary index.

Just add a new field `ReferenceID` to trace origin index

### Check List 

Tests 

 - No code

Related changes

 - No


blacktear23:
No need to add this field. Just close this PR

PR #976, opened 2020-08-15
fix: use OptCharsetWithOptBinary for ENUM/SET

wangggong:
See also: https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy#L6617

Signed-off-by: Wang Ruichao <wangruichao2014@xiaochuankeji.cn>

fix https://github.com/pingcap/tidb/issues/19014

According to the implement in MySQL, add `BINARY` support for `ENUM/SET`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes


tangenta:
Is it enough to support the syntax? Is the behavior in TiDB as expected?

```
create table t (c enum('a', 'A') binary);
insert into t values ('a');
insert into t values ('A');
insert into t values ('b');
select * from t;
```

wangggong:
> Is it enough to support the syntax? Is the behavior in TiDB as expected?
> 
> ```sql
> create table t (c enum('a', 'A') binary);
> insert into t values ('a');
> insert into t values ('A');
> insert into t values ('b');
> select * from t;
> ```

No, and the I'm on my way to add the fit for TiDB. Thanks.

wangggong:
> Is it enough to support the syntax? Is the behavior in TiDB as expected?
> 
> ```sql
> create table t (c enum('a', 'A') binary);
> insert into t values ('a');
> insert into t values ('A');
> insert into t values ('b');
> select * from t;
> ```

For this case, the result for this fix is same as in MySQL. No addition fix in TiDB. Sorry for the mis-test 🙈.

djshow832 (commented):
LGTM

ti-srebot (approved):

PR #977, opened 2020-08-19
Populate offset for func calls

kyleconroy:

I use this package to power the MySQL support in https://github.com/kyleconry/sqlc. One of the features in sqlc is named parameter support via functions. For example, sqlc will take this query

```sql
/* name: BooksByTitleYear :many */
SELECT * FROM books
WHERE title = sqlc.arg('title');
```

And rewrite it with parameter placeholders

```sql
/* name: BooksByTitleYear :many */
SELECT * FROM books
WHERE title = ?;
```

To support this functionality, I've added an Offset field to `FuncCallExpr`. It behaves exactly like the Offset field on SelectField. I've only updated the call site that I need, but happy to do it every place a FuncCallExpr is initialized.

### Check List

Tests

 - Integration test (this lives in the sqlc repository)

Code changes

### DIFF
--- a/ast/functions.go
+++ b/ast/functions.go
@@ -356,6 +356,8 @@ type FuncCallExpr struct {
 	FnName model.CIStr
 	// Args is the function args.
 	Args []ExprNode
+	// Offset is used to get original text.
+	Offset int
 }
 
 // Restore implements Node interface.
--- a/parser.y
+++ b/parser.y
@@ -6631,6 +6631,7 @@ FunctionCallGeneric:
 	}
 |	Identifier '.' Identifier '(' ExpressionListOpt ')'
 	{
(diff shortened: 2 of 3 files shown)


tangenta (commented):
Could you add a related integration test in `parser_test.go`?

kyleconroy:
@tangenta Add a test for added Offset field

tangenta:
I think it is inefficient to add an `offset` field for every struct manually, so I try to solve it in a "meta" level.

Please review the PR #984 and check if it solves the problem.

kyleconroy:
> Please review the PR #984 and check if it solves the problem.

Yep, #984 is a much better solution. I'll close this and follow the other pull request.

PR #978, opened 2020-08-20
model: add StateReplicaOnly type to string

zimulala:
In `StateReplicaOnly` schema state, we see the log is `SchemaState:none`.

Add replica only type to `String`.

### Check List 

Tests 


kennytm (commented):
lgtm

PR #979, opened 2020-08-20
Add 2 Job SchemaStatus for drop column usage

blacktear23:

https://github.com/pingcap/tidb/pull/19303

Add two status for Job.SchemaState used by drop column with composite indices on create temp composite indices step.

### Check List 

Tests 

 - No code

Code changes

- No Change

Side effects

- No

Related changes

- No


blacktear23:
@bb7133 PTAL

PR #980, opened 2020-08-21
parser: return parser warning of ErrWarnDeprecatedIntegerDisplayWidth directly

AilinKid:

parser: return parser warning of `ErrWarnDeprecatedIntegerDisplayWidth` directly rather than `yylex.Errorf(...)`.

because the former has the `cause()` member function, and this error can be traced by `parser.ErrWarnDeprecatedIntegerDisplayWidth.Equal(err)` in TiDB

return parser warning of ErrWarnDeprecatedIntegerDisplayWidth directly rather than yylex.Errorf(...).

### Check List 

Tests 

 - Unit test
 - Integration test


parser.y
```diff
@@ -9850,7 +9850,7 @@ NumericType:
 		x := types.NewFieldType($1.(byte))
 		x.Flen = $2.(int)
 		if $2.(int) != types.UnspecifiedLength && types.TiDBStrictIntegerDisplayWidth {
-			yylex.AppendError(yylex.Errorf("Integer display width is deprecated and will be removed in a future release."))
+			yylex.AppendError(ErrWarnDeprecatedIntegerDisplayWidth.GenWithStackByArgs())
```
zimulala, line 9853:
```suggestion
			yylex.AppendError(ErrWarnDeprecatedIntegerDisplayWidth)
```

PR #983, opened 2020-08-26
model: fix null origin default value for bit column

tangenta:
In PR pingcap/tidb#18036,
```sql
drop table if exists t;
create table t (a bit);
insert into t values (null);
select count(*) from t where a is null;
```

Expected: 
```
+----------+
| count(*) |
+----------+
| 1        |
+----------+
```
But got:
```
+----------+
| count(*) |
+----------+
| 0        |
+----------+
```
If the origin default value in the BIT column is null, it should not be converted to a string. 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N/A

Side effects

N/A

Related changes

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -151,7 +151,7 @@ func (c *ColumnInfo) SetOriginDefaultValue(value interface{}) error {
 
 // GetOriginalDefaultValue gets the origin default value.
 func (c *ColumnInfo) GetOriginDefaultValue() interface{} {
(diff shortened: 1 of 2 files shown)


kennytm (commented):
LGTM

crazycs520 (approved):
LGTM

ti-srebot (approved):

PR #984, opened 2020-08-26
goyacc: record the original text position in AST

### LINKED ISSUE #910: How do I get the position of an AST node?
For [sqlc](https://github.com/kyleconroy/sqlc), I need to know the position of individual AST nodes in the original source text. The PostgreSQL parser I'm using ([pg_query_go](https://github.com/lfittl/pg_query_go/)) attaches a `Location` to many of the AST nodes.

This package has a `Pos` struct, but I don't see a way to get a Pos for a given AST node. Am I missing something?

tangenta:
Close #910.

- Add two methods for the interface `Node`, `SetOriginTextPosition` and `OriginTextPosition` respectively.
- Add a function call test case.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Possible performance regression?

Related changes

N/A


kyleconroy (dismissed):
This is fantastic! Much better than my one-off fixes. Thank you so much.

StevenACoffman:
@tangenta This could use a rebase, if you get a chance. Thanks for the excellent work!

tangenta:
@kennytm Since we record the text position into AST, the ASTs parsed from the **original** SQL and the **restored** SQL may be different. Is it acceptable?

kyleconroy:
I only care about the original SQL, so this solves my issue. Thanks again!

kennytm:
@tangenta we don't care about syntactic equivalence in `Restore()`, only semantic equivalence (e.g. `FETCH FIRST n ROWS` and `LIMIT n` currently has the same `Restore()` representation).

tangenta:
@kennytm PTAL

crazycs520 (commented):
LGTM

ast/base.go
```diff
@@ -20,7 +20,16 @@ import (
 // node is the struct implements node interface except for Accept method.
 // Node implementations should embed it in.
 type node struct {
-	text string
+	text   string
+	offset int
+}
+
```
AilinKid, line 26: may need a comment for the Exported function.

ast/expressions.go
```diff
@@ -1438,3 +1439,41 @@ func (n *SetCollationExpr) Accept(v Visitor) (Node, bool) {
...
+	n.SetOriginTextPosition(0)
+	return n, false
+}
+
+func (e *exprTextPositionCleaner) Leave(n Node) (node Node, ok bool) {
+	return n, true
+}
+
+// ExpressionDeepEqual compares the equivalence of two expressions.
+func ExpressionDeepEqual(a ExprNode, b ExprNode) bool {
```
AilinKid, line 1468: what's the need for this function, test only?

AilinKid (commented):
rest LGTM

ast/expressions.go
tangenta: This is used to compare two expression nodes, regardless of the position of the original text.

ast/base.go
tangenta: Updated.

AilinKid (approved):
LGTM

ti-srebot (approved):

ast/expressions.go
AilinKid: ok

Issue #973, opened 2020-08-14
Support the SQL:2008 Standard `OFFSET m ROWS FETCH FIRST n ROWS` clause


kennytm:
## Feature Request

**Is your feature request related to a problem? Please describe:**

SQL:2008 introduced the [FETCH FIRST clause](https://en.wikipedia.org/wiki/Select_(SQL)#FETCH_FIRST_clause) to unify the LIMIT feature with varying syntax among all RDBMS (CD 9075-2:201X §7.16/6–7). This feature is supported in [PostgreSQL](https://www.postgresql.org/docs/current/sql-select.html#SQL-LIMIT), DB2 and Oracle. Users migrating from outside MySQL may find it more natural to use the standard syntax than LIMIT. (see also https://bugs.mysql.com/bug.php?id=78929.)

**Describe the feature you'd like:**

Parse the clause:

```sql
[ OFFSET m { ROW | ROWS } ]
[ FETCH { FIRST | NEXT } [ n ] { ROW | ROWS } ONLY ]
```

as equivalent to

```sql
LIMIT n OFFSET m
```

(if `n` is missing it default to 1)

(FIRST and NEXT are synonyms)

**Describe alternatives you've considered:**

Don't support it.

**Teachability, Documentation, Adoption, Migration Strategy:**

bb7133:
Sounds an interesting feature :)

time-and-fate:
After discussion with kennytm, I'll only support  `FETCH { FIRST | NEXT } [ n ] { ROW | ROWS } ONLY` part in #985 for now because it seems very complicated to support the `[ OFFSET m { ROW | ROWS } ]` part.

PR #985, opened 2020-08-27
parser: add `FETCH FIRST` syntax support

time-and-fate:

#973
not close it because only a part of the syntax is supported.

Support `FETCH { FIRST | NEXT } [ n ] { ROW | ROWS } ONLY` syntax.

Note1: 

MS SQL Server/Oracle/PostgreSQL support `[ OFFSET m { ROW | ROWS } ] FETCH { FIRST | NEXT } [ n ] { ROW | ROWS } ONLY` syntax. We didn't implement the `[ OFFSET m { ROW | ROWS } ]` part.

That's because `offset` is not a reserved word in MySQL(which is we need to be compatible with) so SQLs like `select * from t offset` and `select 1 offset` are valid. Supporting this syntax means we need to handle those SQLs and SQLs like `select a from t offset offset 3 rows fetch first 2 rows only` correctly. 

I'm not sure why yacc can't resolve this automatically but simply adding a rule for `[ OFFSET m { ROW | ROWS } ]` would cause syntax conflicts. Supporting that and supporting `offset` as an unreserved word can't be easily achieved at the same time based on our current syntax rules.

FYI, other Databases' behavior provided below.
MySQL: `offset` is not a reserved word. The "FETCH FIRST" syntax is not supported.
MS SQL Server: `offset` is not a reserved word. The syntax is supported but must be used after the `ORDER BY` clause.
DB2 & Oracle: `offset` is not a reserved word. The syntax is supported.
PostgreSQL: `offset` is a reserved word. The syntax is supported.

Note2: 

Currently, window function related tokens are treated differently. When window function enabled, these tokens are treated as reserved words. When disabled, they are treated as not keywords.
There is a test for this: https://github.com/pingcap/parser/blob/cc152d5441ed1e684593afb0e25d8389fb69f29d/parser_test.go#L4748-L4761
`ROWS` are one of these window function tokens, but to support this new syntax, we need to treat `ROWS` as a reserved word.
So I modified the `tokenMap` to always treat `ROWS` as a reserved word. This behavior is consistent with MySQL 8.0 but not consistent with MySQL 5.x since `ROWS` became reserved in MySQL 8.0.2.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects
 - Breaking backward compatibility (`ROWS` becomes reserved)

Related changes

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -567,6 +567,7 @@ var tokenMap = map[string]int{
 	"ROW_COUNT":                rowCount,
 	"ROW_FORMAT":               rowFormat,
 	"ROW":                      row,
+	"ROWS":                     rows,
 	"RTREE":                    rtree,
 	"SAMPLES":                  samples,
 	"SAN":                      san,
@@ -794,7 +795,6 @@ var windowFuncTokenMap = map[string]int{
 	"OVER":         over,
 	"PERCENT_RANK": percentRank,
 	"RANK":         rank,
-	"ROWS":         rows,
 	"ROW_NUMBER":   rowNumber,
 	"WINDOW":       window,
 }
--- a/misc.go
+++ b/misc.go
@@ -322,6 +322,7 @@ var tokenMap = map[string]int{
 	"EXTRACT":                  extract,
 	"FALSE":                    falseKwd,
 	"FAULTS":                   faultsSym,
+	"FETCH":                    fetch,
 	"FIELDS":                   fields,
 	"FILE":                     file,
 	"FIRST":                    first,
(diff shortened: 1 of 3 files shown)


kennytm (commented):
lgtm

kennytm:
> This behavior is consistent with MySQL 8.0 but not consistent with MySQL 5.x since `ROWS` became reserved in MySQL 8.0.2.

Just a reminder: when the behavior of MySQL 5.x and 8.0 differ, always prefer to be compatible with 8.0 first.

zz-jason (approved):
LGTM

ti-srebot (approved):

PR #994, opened 2020-08-31
expression: disable constrant flod for control expressions

fzhedu:
solve the constrant fold case in https://github.com/pingcap/tidb/pull/19367

disable the constrant fold in control expressions, such as if(), ifnull() and case... when....

### Check List 

Tests 

 - Unit test in tidb

Related changes


fzhedu:
unnecessary change parser to achieve it.

PR #1000, opened 2020-09-01
model: support columar table

youjiali1995:
Signed-off-by: youjiali1995 <zlwgx1023@gmail.com>

Support columnar table.

Add a `IsColumnar` field to `TableInfo`. It's true when the engine is `TiFlash`.

### Check List 

Tests 

 - No code

Code changes


model/model.go
```diff
@@ -312,6 +312,10 @@ type TableInfo struct {
 
 	// TiFlashReplica means the TiFlash replica info.
 	TiFlashReplica *TiFlashReplicaInfo `json:"tiflash_replica"`
+
+	// IsColumnar means the table is column-oriented.
+	// It's true when the engine of the table is TiFlash.
```
wjhuang2016, line 317: Is TiFlash Only?

youjiali1995: Yes

wjhuang2016: If so, I think it's better to say "It's true when the engine of the table is TiFlash only."

cfzjywxk:
LGTM

youjiali1995:
🤨Why the test always fails? I just added a field.

kennytm:
no need to care about build-integration, it should be fixed from the tidb side.

cfzjywxk:
Seems some planner tests have failed, still unstable by now...

wjhuang2016 (commented):
LGTM

youjiali1995:
@wjhuang2016 Can you merge it?

wjhuang2016 (approved):

PR #1003, opened 2020-09-01
*: fix string(int) issues for Go 1.15 compatibility

LENSHOOD:
Since Go 1.15, expression "string(int)" to make type casting from int to string are not allowed. 
Check more info at [golang/go#3939](https://github.com/golang/go/issues/3939).

Run `make test` at Go 1.15 env would lead to below error:
```shell
./lexer_test.go:204:12: conversion from untyped int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```

Replace such casting expressions to other legally expressions.


kennytm (commented):
LGTM

wjhuang2016 (approved):
LGTM

ti-srebot (approved):

PR #1004, opened 2020-09-01
Fix syntax error caused by unexpected hint

### LINKED ISSUE #987: Unexpected optimizer hint causes syntax error
## Bug Report

1. What did you do?

```
SELECT id FROM tbl WHERE id = 0 FOR UPDATE /*+ xyz */;
```

In MySQL, an unexpected optimizer hint causes a warning, not an error:

```
mysql> SELECT id FROM tbl WHERE id = 0 FOR UPDATE /*+ xyz */;
Empty set, 1 warning (0.01 sec)

mysql> show warnings;
+---------+------+-----------------------------------------------------+
| Level   | Code | Message                                             |
+---------+------+-----------------------------------------------------+
| Warning | 1064 | Optimizer hint syntax error near 'xyz */' at line 1 |
+---------+------+-----------------------------------------------------+
1 row in set (0.00 sec)
```

3. What did you see instead?

This causes a syntax error in TiDB.

```
mysql> SELECT id FROM tbl WHERE id = 0 FOR UPDATE /*+ xyz */;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 53 near "/*+ xyz */"
```

```
mysql> select tidb_version()\G
*************************** 1. row ***************************
tidb_version(): Release Version: v4.0.4
Edition: Community
Git Commit Hash: c61fc7247e9f6bc773761946d5b5294d3f2699a5
Git Branch: heads/v4.0.4
UTC Build Time: 2020-07-31 03:49:54
GoVersion: go1.13.4
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
Check Table Before Drop: false
1 row in set (0.01 sec)
```

ichn-hu:

close #987 

Added a look back token to filter out special case `FOR UPDATE` which might mistakenly take the comment after as a optimization hint.

### Check List 

Tests 

 - Unit test

Code changes

N/A

Side effects

 - Increased code complexity
 
Related changes

N/A

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -59,7 +59,8 @@ type Scanner struct {
 
 	// lastKeyword records the previous keyword returned by scan().
 	// determine whether an optimizer hint should be parsed or ignored.
-	lastKeyword int
+	lastKeyword               int
+	keywoardBeforeLastKeyword int
 
 	// hintPos records the start position of the previous optimizer hint.
 	lastHintPos Pos
@@ -128,6 +129,7 @@ func (s *Scanner) AppendError(err error) {
 func (s *Scanner) Lex(v *yySymType) int {
 	tok, pos, lit := s.scan()
 	s.lastScanOffset = pos.Offset
+	s.keywoardBeforeLastKeyword = s.lastKeyword
 	s.lastKeyword = 0
 	v.offset = pos.Offset
 	v.ident = lit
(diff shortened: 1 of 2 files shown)


lexer.go
```diff
@@ -376,8 +378,13 @@ func startWithSlash(s *Scanner) (tok int, pos Pos, lit string) {
 		// See https://dev.mysql.com/doc/refman/5.7/en/optimizer-hints.html
 		if _, ok := hintedTokens[s.lastKeyword]; ok {
 			// only recognize optimizers hints directly followed by certain
-			// keywords like SELECT, INSERT, etc.
-			isOptimizerHint = true
+			// keywords like SELECT, INSERT, etc., only a special case "FOR UPDATE" needs to be handled
+			// we will report a warning in order to match MySQL's behavior, but the hint content will be ignored
```
kennytm, line 382: note the MySQL doesn't necessarily warn, e.g. the following produces no warnings (and the hint is ignored too).

```sql
select 1 for update /*+ hash_join() */;
```

but we don't need to replicate this MySQL behavior. the current solution is good enough.

kennytm (commented):
LGTM

lysu (approved):
LGTM

ti-srebot (approved):

PR #982, opened 2020-08-25
using standard error code to replace terror

imtbkcat:
### What problem does this PR solve? 
According to [RFC](https://github.com/pingcap/tidb/blob/master/docs/design/2020-05-08-standardize-error-codes-and-messages.md), TiDB need to adjust error format and using standard error to replace terror. The standard error has been implemented in https://github.com/pingcap/errors/tree/standard-error

### What is changed and how it works?

Since standard error is very similar with `terror`. We can change very few api to replace the `terror`. Just rename `terror.Error` with `errors.Error`

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change

Side effects

 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - None

PR #1014, opened 2020-09-08
using standard error code to replace terror (#982)

imtbkcat:
cherry-pick #982  to release-4.0

According to [RFC](https://github.com/pingcap/tidb/blob/master/docs/design/2020-05-08-standardize-error-codes-and-messages.md), TiDB need to adjust error format and using standard error to replace terror. The standard error has been implemented in https://github.com/pingcap/errors/tree/standard-error

Since standard error is very similar with `terror`. We can change very few api to replace the `terror`. Just rename `terror.Error` with `errors.Error`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

 - None


bobotu (approved):
LGTM

lonng:
LGTM

PR #1026, opened 2020-09-14
parser : support table statement and value statement

YangKian:
Issue Number: pingcap/tidb#19934
Support the TABLE and VALUES statements introduced in MySQL 8.0.19.

- [x]  `TABLE tbl_name [ORDER BY col_name] [LIMIT limit [OFFSET offset]];`

- [x] `VALUES ROW(val1, val2, …), … [ORDER BY column_N [LIMIT limit [OFFSET offset]]];`

### Check List 

Tests 

 - Unit test
 - Integration test


parser.y
```diff
@@ -11932,4 +11934,33 @@ EncryptionOpt:
...
+		$$ = nil
+	}
+|	"LIMIT" LimitOption
+	{
+		$$ = &ast.Limit{Count: $2.(ast.ExprNode)}
+	}
+|	"LIMIT" LimitOption "OFFSET" LimitOption
+	{
+		$$ = &ast.Limit{Offset: $4.(ast.ExprNode), Count: $2.(ast.ExprNode)}
+	}
```
kennytm, line 11965: please just use `SelectStmtLimit`. while the documentation in https://dev.mysql.com/doc/refman/8.0/en/table.html only mentions `LIMIT limit [OFFSET offset]`, in fact the legacy form `LIMIT offset, limit` does work in TABLE statements.

parser_test.go
```diff
@@ -501,6 +501,15 @@ func (s *testParserSuite) TestDMLStmt(c *C) {
...
 		ROLLBACK;`, true, "START TRANSACTION; INSERT INTO `tmp` SELECT * FROM `bar`; SELECT * FROM `tmp`; ROLLBACK"},
 
+		// table statement
+		{"TABLE t", true, "TABLE `t`"},
+		{"TABLE t ORDER BY b", true, "TABLE `t` ORDER BY `b`"},
+		{"TABLE t LIMIT 3", true, "TABLE `t` LIMIT 3"},
+		{"TABLE t ORDER BY b LIMIT 3", true, "TABLE `t` ORDER BY `b` LIMIT 3"},
+		{"TABLE t ORDER BY b LIMIT 3 OFFSET 2", true, "TABLE `t` ORDER BY `b` LIMIT 3 OFFSET 2"},
+		{"TABLE t1 UNION TABLE t2", true, "TABLE `t1` UNION TABLE `t2`"},
+		{"TABLE t1 UNION SELECT * FROM t2", true, "TABLE `t1` UNION SELECT * FROM `t2`"},
```
kennytm, line 511: please also include these example test cases:

```sql
SELECT * FROM t1 WHERE a IN (TABLE t2);
TABLE t1 INTERSECT TABLE t2;
TABLE t1 EXCEPT TABLE t2;
```

and optionally:

```sql
INSERT INTO `test`.`b` TABLE `test`.`a`;
REPLACE INTO `test`.`b` TABLE `test`.`a`;
CREATE TABLE `test`.`b` TABLE `test`.`a`;
CREATE VIEW `test`.`b` AS TABLE `test`.`a`;
TABLE t1 INTO OUTFILE 'a.txt';
```

parser.y
YangKian: Yeah, I found that too. I've not checked `"FETCH" FirstOrNext FetchFirstOpt RowOrRows "ONLY"` branch in the `SelectStmtLimit`, if this branch is not allowed, should I create the new `TableStmtLimit` or just use the `SelectStmtLimit`?

parser_test.go
YangKian: Get~

parser.y
kennytm: still just use `SelectStmtLimit`. 

`TABLE t1 FETCH FIRST ROW ONLY` is a standard ISO syntax and also supported in PostgreSQL, so there's no reason to reject it.

YangKian: Ok, I'll fix it

YangKian:
Hi~ @kennytm @tangenta Now I finished the basic function of table statement and value statement, I'm stuck in support such SQL:

` SELECT c FROM t1 UNION TABLE t2`
` TABLE t1 UNION SELECT c FROM t2`
` SELECT * FROM t1 WHERE a IN (TABLE t2)`

The failed reason I found is: in `type SetOprSelectList struct`, I add a new field `TableList  *SetOprTableList` to support `TABLE UNION` statement, but when an union statement contains both `select` and `table` statement, the filed `TableList` and `SelectList` will conflict. I can't find a good idea to deal with this situation, can you give me some advice?

Thanks!

(ps: still leaves some comments undeleted, I'll clean them when all function completed)

kennytm:
perhaps change `SetOprSelectList` to hold `[]DMLNode` instead of `[]*SelectStmt` (or maybe even introduce a new interface covering `*SelectStmt`, `*TableStmt` and `*ValueStmt`)

parser.y
```diff
@@ -5564,6 +5580,12 @@ InsertValues:
 	{
 		$$ = &ast.InsertStmt{Setlist: $2.([]*ast.Assignment)}
 	}
+|	"ROW" '(' ValuesList ')'
+	{
+		$$ = &ast.InsertStmt{
+			Lists: $3.([][]ast.ExprNode),
+		}
+	}
```
kennytm, line 5588: `INSERT INTO x ROW (...)` is not valid. should be changing `ValuesList` instead.

YangKian:
> perhaps change `SetOprSelectList` to hold `[]DMLNode` instead of `[]*SelectStmt` (or maybe even introduce a new interface covering `*SelectStmt`, `*TableStmt` and `*ValueStmt`)

I tried this method, but I'm failed _(:з」∠)_ maybe I do something wrong. Anyway, I'll try it again, thanks!

YangKian:
@kennytm Hi~ I've done the work, PTAL

parser.y
YangKian: fix

YangKian:
Emm... Seems so many error caused by my modifying of the field name related with `SetOprStmt`, how can I fix this =-=

kennytm:
@YangKian if you mean the integration test, we don't need to fix them _now_. Instead, fix them in pingcap/tidb#19934, as explained in https://github.com/pingcap/parser/blob/master/docs/update-parser-for-tidb.md.

YangKian:
Yeah, all the unit tests passed in the local develop environment, I'll follow your comment to fix the integration test.

YangKian:
@kennytm I follow the doc and create a pr pingcap/tidb#20176, seems the replacement of `go.mod` doesn't take effect, just don't know why, it works in my local development environment _(:з」∠)_ By the way, the code is ready for review, please take a look when you have time, thanks!

kennytm:
The `check_dev` CI errors are intentional (to avoid accidentally merging a PR with `replace` directives). Don't worry about them.

parser_test.go
```diff
@@ -502,6 +502,35 @@ func (s *testParserSuite) TestDMLStmt(c *C) {
...
+		{"CREATE VIEW v AS TABLE t", true, "CREATE ALGORITHM = UNDEFINED DEFINER = CURRENT_USER SQL SECURITY DEFINER VIEW `v` AS TABLE `t`"},
+		{"CREATE VIEW v AS (TABLE t)", true, "CREATE ALGORITHM = UNDEFINED DEFINER = CURRENT_USER SQL SECURITY DEFINER VIEW `v` AS (TABLE `t`)"},
+		{"SELECT * FROM t1 WHERE a IN (TABLE t2)", true, "SELECT * FROM `t1` WHERE `a` IN (TABLE `t2`)"},
+
+		// values statement
+		{"VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8)", true, "VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8)"},
+		{"VALUES ROW(1,s,3.1), ROW(5,y,9.9)", true, "VALUES ROW(1,`s`,3.1), ROW(5,`y`,9.9)"},
+		{"VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8) LIMIT 3", true, "VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8) LIMIT 3"},
+		{"VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8) ORDER BY a", true, "VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8) ORDER BY `a`"},
+		{"VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8) ORDER BY a LIMIT 2", true, "VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8) ORDER BY `a` LIMIT 2"},
```
kennytm, line 532: please add a positive test case for:

```sql
values row(1);
```

the following should also be *valid syntax* with semantic error handled by TiDB itself:

```sql
values row();            -- ERROR 3942 (HY000): Each row of a VALUES clause must have at least one column, unless when used as source in an INSERT statement
values row(1, default);  -- ERROR 3943 (HY000): A VALUES clause cannot use DEFAULT values, unless used as a source in an INSERT statement.
values row(1), row(2, 3);-- ERROR 1136 (21S01): Column count doesn't match value count at row 2
```

also add a negative test case:

```sql
values (1, 2);   -- syntax error
```

parser_test.go
```diff
@@ -502,6 +502,35 @@ func (s *testParserSuite) TestDMLStmt(c *C) {
...
+		{"INSERT INTO t.a TABLE t.b", true, "INSERT INTO `t`.`a` TABLE `t`.`b`"},
+		{"REPLACE INTO ta TABLE tb", true, "REPLACE INTO `ta` TABLE `tb`"},
+		{"REPLACE INTO t.a TABLE t.b", true, "REPLACE INTO `t`.`a` TABLE `t`.`b`"},
+		{"TABLE t1 INTO OUTFILE 'a.txt'", true, "TABLE `t1` INTO OUTFILE 'a.txt'"},
+		{"TABLE t ORDER BY a INTO OUTFILE '/tmp/abc'", true, "TABLE `t` ORDER BY `a` INTO OUTFILE '/tmp/abc'"},
+		{"CREATE TABLE t.a TABLE t.b", true, "CREATE TABLE `t`.`a` AS TABLE `t`.`b`"},
+		{"CREATE TABLE ta TABLE tb", true, "CREATE TABLE `ta` AS TABLE `tb`"},
+		{"CREATE TABLE ta (x INT) TABLE tb", true, "CREATE TABLE `ta` (`x` INT) AS TABLE `tb`"},
+		{"CREATE VIEW v AS TABLE t", true, "CREATE ALGORITHM = UNDEFINED DEFINER = CURRENT_USER SQL SECURITY DEFINER VIEW `v` AS TABLE `t`"},
+		{"CREATE VIEW v AS (TABLE t)", true, "CREATE ALGORITHM = UNDEFINED DEFINER = CURRENT_USER SQL SECURITY DEFINER VIEW `v` AS (TABLE `t`)"},
```
kennytm, line 524: also include:

```sql
create table a values row(1);
create view a as values row(1);
```

kennytm:
I'm thinking of an alternative formulation, please see if this could reduce the amount of changes needed in TiDB?

Instead of having separate `SelectStmt`, `TableStmt` and `ValuesStmt` types, we use a single `SelectStmt` to represent all three statements:

```diff
+type SelectStmtKind uint8
+const (
+       SelectStmtKindSelect SelectStmtKind = iota
+       SelectStmtKindTable
+       SelectStmtKindValues
+)
+
 // SelectStmt represents the select query node.
 // See https://dev.mysql.com/doc/refman/5.7/en/select.html
 type SelectStmt struct {
        dmlNode
        resultSetNode

+       Kind  SelectStmtKind
+       Lists []*RowExpr  // fill this in only when Kind == SelectStmtKindValues
+    
```

for instance a `TABLE a;` statement should generate a `&SelectStmt{Kind: SelectStmtKindTable, From: &TableRefsClause{TableRefs: &Join{Left: &TableName{Name: NewCIStr("a")}}}}`

YangKian:
Thanks for all the advice, I'll fix them as soon as possible

YangKian:
@kennytm Hi~ I've done the work, PTAL

ast/dml.go
```diff
@@ -775,7 +775,27 @@ func (n *OrderByClause) Accept(v Visitor) (Node, bool) {
...
+)
+
+func (s *SelectStmtKind) String() string {
+	switch *s {
+	case SelectStmtKindSelect:
+		return "SELECT "
+	case SelectStmtKindTable:
+		return "TABLE "
+	case SelectStmtKindValues:
+		return "VALUES "
```
kennytm, line 793: for consistency don't include the trailing space. (as a result add `ctx.WritePlain(" ")` after line 850.)

```suggestion
	case SelectStmtKindSelect:
		return "SELECT"
	case SelectStmtKindTable:
		return "TABLE"
	case SelectStmtKindValues:
		return "VALUES"
```

ast/dml.go
```diff
@@ -823,95 +847,110 @@ func (n *SelectStmt) Restore(ctx *format.RestoreCtx) error {
...
+			}
+		}
+	case SelectStmtKindTable:
+		if err := n.From.Restore(ctx); err != nil {
+			return errors.Annotate(err, "An error occurred while restore SelectStmt.From")
+		}
+	case SelectStmtKindValues:
+		for i, v := range n.Lists {
+			if err := v.Restore(ctx); err != nil {
+				return errors.Annotate(err, "An error occurred while restore ValuesStmt.OrderBy")
```
kennytm, line 950:
```suggestion
				return errors.Annotatef(err, "An error occurred while restore SelectStmt.Lists[%d]", i)
```

parser.y
```diff
@@ -1027,6 +1027,8 @@ import (
 	RoleSpecList                           "Rolename and auth option list"
 	RowFormat                              "Row format option"
 	RowValue                               "Row value"
+	RowStmt                                "Row constructor statement"
```
kennytm, line 1030: it is just called "row constructor", not a statement :)

```suggestion
	RowStmt                                "Row constructor"
```

kennytm (commented):
Rest LGTM

ast/dml.go
YangKian: fixed

ast/dml.go
YangKian: So sorry, I forgot to change this, fix now

parser.y
YangKian: fixed

YangKian:
I'm done 😄

kennytm:
LGTM, thanks!

PTAL @tangenta

parser.y
```diff
@@ -12015,4 +12060,31 @@ EncryptionOpt:
...
+RowStmtField:
+	RowValue
+	{
+		$$ = &ast.RowExpr{Values: $1.([]ast.ExprNode)}
+	}
+|	RowStmtField ',' RowValue
+	{
+		values := append($1.([]ast.ExprNode), $3.(ast.ExprNode))
+		$$ = &ast.RowExpr{Values: values}
+	}
```
tangenta, line 12089: This rule seems to be unused. We can remove `RowStmtField` and use `RowValue` instead.

tangenta (commented):
Rest LGTM

tangenta (approved):
LGTM

PR #1031, opened 2020-09-15
model: add SQL mode and warnings field for ddl job

AilinKid:
Add SQL mode and warnings field for DDL job

When we do background DDL jobs, passing the SQL Mode into the job, so the background worker can identify whether interpret some error as warnings under this kind of SQL mode.

### Check List 

Tests 

 - Unit test
 - Integration test (in TiDB side)

Code changes


model/ddl.go
```diff
@@ -222,6 +223,11 @@ type Job struct {
 
 	// Priority is only used to set the operation priority of adding indices.
 	Priority int `json:"priority"`
+
+	// SQL Mode
+	SqlMode       mysql.SQLMode
+	Warnings      []*terror.Error
+	WarningsCount []int64
```
kennytm, line 230: why we need the `WarningsCount` array? like, why represent it as `[w1, w2, w3], [1, 3, 1]` instead of a single Warnings array `[w1, w2, w2, w2, w3]`?

AilinKid: Here are the things:
Since the reorg job will fetch all data of the table, it is possible to generate warnings as much as the rows. To store all these warnings in DDL job as one raw KV is not that suitable.
So we try to keep the first one warning for each warning type and add the warning counts for the next one of the same type.

kennytm: ok.

djshow832 (commented):
LGTM

model/ddl.go
zimulala: Could we put these fields to `ReorgMeta`? It seems we will use sql_mode in reorg.

model/ddl.go
```diff
@@ -222,6 +223,11 @@ type Job struct {
 
 	// Priority is only used to set the operation priority of adding indices.
 	Priority int `json:"priority"`
+
+	// SQL Mode
+	SqlMode       mysql.SQLMode
```
zimulala, line 228:
```suggestion
	SQLMode       mysql.SQLMode
```

model/ddl.go
```diff
@@ -222,6 +223,11 @@ type Job struct {
 
 	// Priority is only used to set the operation priority of adding indices.
 	Priority int `json:"priority"`
+
+	// SQL Mode
```
zimulala, line 227: This comment is meaningless.

AilinKid: Addressed

model/ddl.go
zimulala: PTAL @AilinKid

AilinKid: JOB model needs these fields for front-end to fetch ddl results.
In my first edition, I did store the warnings in the reorg meta but found these warnings was still finnally stored in job. so I abandon this way.

zimulala: I don't understand. I mean to put these fields to `job. ReorgMeta `.

zimulala (approved):
LGTM

ti-srebot (approved):

PR #1032, opened 2020-09-16
fix CreateTablespacePriv order in const

xiaodong-ji:
Tidb does not start properly with error `[error="[privilege:8049]mysql.user"]`.

Adjust the order of CreateTablespacePriv.

### Check List

Tests

### DIFF
--- a/mysql/const.go
+++ b/mysql/const.go
@@ -211,6 +211,8 @@ const (
 	SuperPriv
 	// CreateUserPriv is the privilege to create user.
 	CreateUserPriv
+	// CreateTablespacePriv is the privilege to create tablespace.
+	CreateTablespacePriv
 	// TriggerPriv is not checked yet.
 	TriggerPriv
 	// DropPriv is the privilege to drop schema/table.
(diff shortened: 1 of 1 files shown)


tangenta (commented):
LGTM

mysql/const.go
```diff
@@ -251,11 +251,11 @@ const (
 	// ConfigPriv is the privilege to enable the use SET CONFIG statements.
 	ConfigPriv
 
-	// AllPriv is the privilege for all actions.
-	AllPriv
-
 	// CreateTablespacePriv is the privilege to create tablespace.
 	CreateTablespacePriv
+
+	// AllPriv is the privilege for all actions.
```
kennytm, line 257: perhaps add a warning comment like #952 to ensure `AllPriv` is always the last value.

xiaodong-ji: OK

qw4990:
PTAL @kennytm

kennytm (approved):
LGTM

ti-srebot (approved):

PR #1033, opened 2020-09-18
using standard error code to replace terror (#982)

imtbkcat:
cherry-pick #982  to release-4.0

According to [RFC](https://github.com/pingcap/tidb/blob/master/docs/design/2020-05-08-standardize-error-codes-and-messages.md), TiDB need to adjust error format and using standard error to replace terror. The standard error has been implemented in https://github.com/pingcap/errors/tree/standard-error

Since standard error is very similar with `terror`. We can change very few api to replace the `terror`. Just rename `terror.Error` with `errors.Error`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

 - None


terror/terror.go
```diff
@@ -151,154 +175,39 @@ func (ec ErrClass) NewStd(code ErrCode) *Error {
...
-func (e *Error) getMySQLErrorCode() uint16 {
-	codeMap, ok := ErrClassToMySQLCodes[e.class]
+func getMySQLErrorCode(e *Error) uint16 {
+	rfcCode := e.RFCCode()
+	var class ErrClass
+	if index := strings.Index(string(rfcCode), ":"); index > 0 {
+		if ec, has := rfcCode2errClass[string(rfcCode)[:index]]; has {
+			class = ec
+		} else {
+			log.Warn("Unknown error class", zap.String("class", string(rfcCode)[:index]))
```
lonng, line 196: Do we need to warn this?

terror/terror.go
```diff
@@ -119,23 +128,38 @@ func (ec ErrClass) NotEqualClass(err error) bool {
...
+
+// New defines an *Error with an error code and an error message.
+// Usually used to create base *Error.
+// Attention:
+// this method is not goroutine-safe and
+// usually be used in global variable initializer
+func (ec ErrClass) New(code ErrCode, message string) *Error {
+	rfcCode := ec.initError(code)
+	err := errors.Normalize(message, errors.MySQLErrorCode(int(code)), errors.RFCCodeText(rfcCode))
+	errCodeMap[code] = err
```
lonng, line 152: Is there any concurrent problem?

terror/terror.go
```diff
@@ -119,23 +128,38 @@ func (ec ErrClass) NotEqualClass(err error) bool {
...
+	return err
+}
+
+// NewStdErr defines an *Error with an error code, an error
+// message and workaround to create standard error.
+func (ec ErrClass) NewStdErr(code ErrCode, message string, desc string, workaround string) *Error {
+	rfcCode := ec.initError(code)
+	err := errors.Normalize(message, errors.MySQLErrorCode(int(code)), errors.RFCCodeText(rfcCode), errors.Description(desc), errors.Workaround(workaround))
+	errCodeMap[code] = err
```
lonng, line 161: ditto

terror/terror.go
```diff
@@ -151,154 +175,39 @@ func (ec ErrClass) NewStd(code ErrCode) *Error {
...
-		return true
-	}
-	inErr, ok := originErr.(*Error)
-	return ok && e.class == inErr.class && e.code == inErr.code
-}
-
-// NotEqual checks if err is not equal to e.
-func (e *Error) NotEqual(err error) bool {
-	return !e.Equal(err)
+	return errors.Normalize(message, errors.MySQLErrorCode(int(code)), errors.RFCCodeText(fmt.Sprintf("%s:%d", errClass2Desc[ec], code)))
```
lonng, line 178: What if `errClass2Desc[ec]` is empty?

terror/terror.go
imtbkcat: yes, this method is not groutine-safe, but always used for global init.

terror/terror.go
imtbkcat: errClass2Desc[ec] will not be empty, because the caller has been init at error class register.

terror/terror.go
imtbkcat: fixed, PTAL

lonng (commented):
LGTM

bobotu (approved):
LGTM

PR #1036, opened 2020-09-21
ast: reset JoinLevel at the beginning of TableRefsClause.Restore

Hexilee:
Signed-off-by: hexilee <i@hexilee.me>

Currently, `format.RestoreCtx.JoinLevel` does not work well in the multi-level subquery. 
If you parse the following SQL and restore it, you will get unnecessary pairs of parentheses in the inner subquery.

```SQL
-- origin
SELECT * FROM (SELECT * FROM (SELECT * FROM `t1`) AS `tmp`) AS `tmp`
```

```SQL
-- restored
SELECT * FROM (SELECT * FROM ((SELECT * FROM (`t1`)) AS `tmp`)) AS `tmp`
```

Unnecessary parentheses can be ignored by tidb. However, in mysql, it will cause a syntax error.

```log
mysql 8.0.21

mysql root@127.0.0.1:test> select * from (SELECT * FROM ((SELECT * FROM (`t1`)) AS `tmp`)) AS `tmp`;
(1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')) AS `tmp`' at line 1")
```

This PR reset `JoinLevel` at the beginning of `TableRefsClause.Restore`, then `JoinLevel` will never affect subquery.

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -615,6 +615,11 @@ type TableRefsClause struct {
 
 // Restore implements Node interface.
 func (n *TableRefsClause) Restore(ctx *format.RestoreCtx) error {
+	level := ctx.JoinLevel
+	ctx.JoinLevel = 0
+	defer func() {
+		ctx.JoinLevel = level
+	}()
 	if err := n.TableRefs.Restore(ctx); err != nil {
 		return errors.Annotate(err, "An error occurred while restore TableRefsClause.TableRefs")
 	}
--- a/ast/dml_test.go
+++ b/ast/dml_test.go
@@ -234,6 +234,7 @@ func (ts *testDMLSuite) TestTableRefsClauseRestore(c *C) {
 		{"t", "`t`"},
 		{"t1 join t2", "`t1` JOIN `t2`"},
 		{"t1, t2", "(`t1`) JOIN `t2`"},
+		{"(select * from (select * from t1) as tmp) as tmp", "(SELECT * FROM (SELECT * FROM `t1`) AS `tmp`) AS `tmp`"},
 	}
 	extractNodeFunc := func(node Node) Node {
 		return node.(*SelectStmt).From


kennytm:
The culprit is actually the `Join` structure being used to represent non-JOINs. When the `Join` does not represent a real JOIN the JoinLevel should not be changed. Please change the patch to 

```diff
diff --git a/ast/dml.go b/ast/dml.go
index f6e4a1d..ac33cf8 100644
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -91,12 +91,15 @@ func (n *Join) Restore(ctx *format.RestoreCtx) error {
                ctx.WritePlain("(")
                defer ctx.WritePlain(")")
        }
-       ctx.JoinLevel++
+       if n.Right != nil {
+               ctx.JoinLevel++
+       }
        if err := n.Left.Restore(ctx); err != nil {
                return errors.Annotate(err, "An error occurred while restore Join.Left")
        }
-       ctx.JoinLevel--
-       if n.Right == nil {
+       if n.Right != nil {
+               ctx.JoinLevel--
+       } else {
                return nil
        }
        if n.NaturalJoin {
```

also move the test case to `parser_test.go` or somewhere that tests a complete statement. `TestTableRefsClauseRestore` is not the correct function to insert this test.

Hexilee:
I don't think so. The following example also causes a syntax error in mysql.

```SQL
-- origin
SELECT * FROM (SELECT * FROM (SELECT * FROM `t1`) AS `tmp`) AS `tmp`, t
```

```SQL
-- restored
SELECT * FROM (SELECT * FROM ((SELECT * FROM (`t1`)) AS `tmp`)) AS `tmp`, t
```

The culprit is the `JoinLevel` affects subqueries.

PR #1040, opened 2020-09-27
digester: consider paramMarker as literal when normalizing SQL

time-and-fate:

https://github.com/pingcap/tidb/issues/19836

Currently, `normalize()` doesn't handle `paramMarker`('?'), this makes some bindings can't work as expected like the issue above.

Prepared statement `select * from t limit ?, ?` is normalized to `select * from t limit ?, ?`. However, a binding for `select * from t limit 1, 2` is normalized to and can only be matched by `select * from t limit ...`.

When normalizing SQL, consider `paramMarker` as literal, this will make `paramMarker` be normalized to `genericSymbol`. This makes sense also because both `paramMarker` and `genericSymbol` correspond to '?'.

After this change, `select * from t limit ?, ?` will be normalized to `select * from t limit ...` instead of `select * from t limit ?, ?`.

### Check List 

Tests 

 - Unit test


eurekaka (commented):
LGTM

time-and-fate:
@SunRunAway PTAL~

eurekaka:
@SunRunAway PTAL

kennytm:
LGTM

eurekaka (approved):

PR #1043, opened 2020-09-29
correct varchar() binary 

wjhuang2016:

Refer to https://github.com/pingcap/tidb/pull/20269

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


ast/ddl_test.go
```diff
@@ -260,7 +260,7 @@ func (ts *testDDLSuite) TestDDLColumnDefRestore(c *C) {
 		{"id DATETIME", "`id` DATETIME"},
 		{"id DECIMAL(4,2)", "`id` DECIMAL(4,2)"},
 		{"id char(1)", "`id` CHAR(1)"},
-		{"id varchar(10) BINARY", "`id` VARCHAR(10) BINARY"},
```
kennytm, line 263: this looks wrong to me.

wjhuang2016: Yeah, this PR is not complete.
I'm thinking that how to deal with BINARY.
Ignore it or set collation like MySQL?
I prefer to ignore it.

tangenta: @nullnotnil PTAL

nullnotnil: I prefer set collate over ignoring. Ignoring could create problems if collations are enabled and there is an index on the column.

As @kennytm points out, `BINARY` is just an alias for collate as the current charsets binary collation. It even does this in MySQL 5.7, but without the deprecation warning:

```sql
mysql [localhost:5731] {msandbox} (test) > CREATE TABLE t (a varchar(10) binary);

mysql [localhost:5731] {msandbox} (test) > SHOW CREATE TABLE t\G
*************************** 1. row ***************************
       Table: t
Create Table: CREATE TABLE `t` (
  `a` varchar(10) CHARACTER SET latin1 COLLATE latin1_bin DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1
1 row in set (0.00 sec)
```

wjhuang2016:
After the discussion, this problem should be fixed in pingcap/tidb#20269.

PR #1044, opened 2020-10-05
add constant

pengdaqian2020:

Issue Number: close pingcap/tidb#19907

add two constant for function **SUM** and **AVG**

### Check List

Tests

 - No code

### DIFF
--- a/mysql/const.go
+++ b/mysql/const.go
@@ -289,6 +289,11 @@ const (
 	MaxDoublePrecisionLength = 53
 )
 
+const (
+	SumPrecisionIncrement = 22
+	AVGPrecisionIncrement = 4
+)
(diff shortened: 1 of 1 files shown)


mysql/const.go
```diff
@@ -289,6 +289,11 @@ const (
 	MaxDoublePrecisionLength = 53
 )
 
+const (
+	SumPrecisionIncrement = 22
+	AVGPrecisionIncrement = 4
+)
+
```
djshow832, line 296: I don't think they should be added in the parser. They are relevant to the executor and it's better to be added in the tidb repo.

pengdaqian2020: ok

PR #1058, opened 2020-10-16
collation: add new pinyin collation utf8mb4_zh_pinyin_tidb_as_cs

xiongjiwei:

add a new collation named `utf8mb4_zh_pinyin_tidb_as_cs` which has collation id `2048`
proposal https://github.com/pingcap/tidb/pull/19984
tidb PR https://github.com/pingcap/tidb/pull/20504

a new method to get collation by given id.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


wjhuang2016 (commented):
How about rename `utf8mb4_general_zh_ci` to `utf8mb4_zh_ci`

kennytm:
if we want to follow MySQL's naming convention it should be `utf8mb4_chinese_ci` (similar to `big5_chinese_ci`, `gb18030_chinese_ci`, `gb2312_chinese_ci`, `gb2312_chinese_ci`).

(note that CLDR also defined the "stroke" and "zhuyin" order, and we could have `utf8mb4_chinese_stroke_ci` etc.)

wjhuang2016:
I don't want to follow MySQL's old naming convention, `zh` is more standard than `chinese`.
As you can see in the new collation introduced in 8.0, `utf8mb4_zh_0900_as_cs` uses `zh`.
Consider that CLDR has defined "pinyin", "stroke" and "zhuyin", we can use "utf8mb4_zh_pinyin", "utf8mb4_zh_stroke", "utf8mb4_zh_zhuyin". If we want to distinguish case-sensitive, we can add extra '_cs' or '_ci'.

xiongjiwei:
I prefer contains `tidb` in collation name, it indicate this collation is `tidb` unique.   I vote`utf8mb4_zh_tidb0500_ci`

wjhuang2016 (commented):
LGTM

charset/charset.go
```diff
@@ -421,6 +431,7 @@ var collations = []*Collation{
 	{246, "utf8mb4", "utf8mb4_unicode_520_ci", false},
 	{247, "utf8mb4", "utf8mb4_vietnamese_ci", false},
 	{255, "utf8mb4", "utf8mb4_0900_ai_ci", false},
+	{2048, "utf8mb4", "utf8mb4_tidb_zh_pinyin_cs", false},
```
kennytm, line 434:
```suggestion
	{2048, "utf8mb4", "utf8mb4_zh_tidb_pinyin_cs", false},
```

kennytm (commented):
assuming it is a typo, rest LGTM.

(btw, are we sure we want case-sensitive ('a' ≠ 'A') not case-insensitive ('a' = 'A') here?)

charset/charset.go
xiongjiwei: > are we sure we want case-sensitive

yes, cs implemention will be eaiser, and the newest chinese language collation `utf8mb4_zh_0900_cs_as` is cs too

kennytm (approved):

ti-srebot (dismissed):

charset/charset.go
```diff
@@ -421,6 +431,7 @@ var collations = []*Collation{
 	{246, "utf8mb4", "utf8mb4_unicode_520_ci", false},
 	{247, "utf8mb4", "utf8mb4_vietnamese_ci", false},
 	{255, "utf8mb4", "utf8mb4_0900_ai_ci", false},
+	{2048, "utf8mb4", "utf8mb4_zh_tidb_pinyin_cs", false},
```
bb7133, line 434:
```suggestion
	{2048, "utf8mb4", "utf8mb4_tidb_zh_pinyin_cs", false},
```
How about this?

bb7133: Please update the PR description accordingly.

xiongjiwei: the naming convention in mysql,  `zh` follows `utf8mb4` and `cs` at last, `tidb_pinyin` can be seems as a special version.

kennytm: tidb is the version, and `pinyin` is the variant. this is similar to `utf8mb4_de_pb_0900_ai_ci` (pb = phonebook) and `utf8mb4_es_trad_0900_ai_ci` (trad = traditional).

xiongjiwei: updated name to `utf8mb4_zh_pinyin_tidb_as_cs`

bb7133:
LGTM

Thanks for your suggesion @kennytm

bb7133 (approved):

ti-srebot (approved):

PR #1061, opened 2020-10-19
ast: Don't write comma when restore TableOption of ALTER TABLE

lance6716:
it's optional to write a comma between `table_option`

https://dev.mysql.com/doc/refman/5.7/en/alter-table.html

if we write it default, a single schema change DDL will become multi-schema change DDL as in https://github.com/pingcap/dm/issues/1175#issuecomment-709900540

don't write comma

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Breaking backward compatibility (?)

Related changes

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -2353,7 +2353,7 @@ func (n *AlterTableSpec) Restore(ctx *format.RestoreCtx) error {
 		default:
 			for i, opt := range n.Options {
 				if i != 0 {
-					ctx.WritePlain(", ")
+					ctx.WritePlain(" ")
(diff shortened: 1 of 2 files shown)


lance6716:
PTAL @kennytm @AilinKid 
cc @csuzhangxc

kennytm:
(circle ci is stuck)

kennytm (commented):
lgtm

(but tidb really should allow these "multi-schema changes" which can be collapsed)

tangenta (approved):
LGTM

PR #1067, opened 2020-10-26
parser: support SQL bind syntax for update / delete

eurekaka:

Parser changes for https://github.com/pingcap/tidb/issues/15827

Support `Create Binding` / `Drop Binding` for `Update` / `Delete`. Note that, `Delete` has a special syntax `Delete From Using`, which would conflict with SQL bind syntaxes, so we don't support SQL bind syntax for this special kind of `Delete` queries.

### Check List 

Tests 

 - Unit test

Code changes

N/A

Side effects

N/A

Related changes


kennytm:
After this PR it is syntactically possible to write

```sql
CREATE SESSION BINDING FOR 
  SELECT * FROM a
USING
  UPDATE a SET b = 1;
```

will this be rejected in TiDB?

eurekaka:
> After this PR it is syntactically possible to write
> 
> ```sql
> CREATE SESSION BINDING FOR 
>   SELECT * FROM a
> USING
>   UPDATE a SET b = 1;
> ```
> 
> will this be rejected in TiDB?

Yes, we can report error in TiDB later.

kennytm (commented):
LGTM

lzmhhh123 (commented):
LGTM

parser.y
```diff
@@ -11142,6 +11151,11 @@ RoleSpecList:
 		$$ = append($1.([]*ast.UserSpec), $3.(*ast.UserSpec))
 	}
 
+BindableStmt:
+	SetOprStmt1
+|	UpdateStmt
+|	DeleteWithoutUsingStmt
```
lzmhhh123, line 11157: Just for curious. What's the impact if we ignore `DeleteWithUsingStmt`?

eurekaka: We can't create SQL bind for `delete using` style statements.

lzmhhh123: Like this SQL `DELETE FROM t1, t2 USING t1 INNER JOIN t2 INNER JOIN t3
WHERE t1.id=t2.id AND t2.id=t3.id;`?

eurekaka: Yes.

lzmhhh123: OK.

lzmhhh123 (approved):

PR #1071, opened 2020-10-28
support SELECT FROM TABLESAMPLE syntax

tangenta:
See https://github.com/pingcap/tidb/issues/20567.

The syntax for "TABLESAMPLE" is as follows:
```
sample_clause = "TABLESAMPLE" [ "BERNOULLI" | "SYSTEM" | "REGIONS" ] '(' expression [ "PERCENT" | "ROWS" ] ')'
                [ "REPEATABLE" '(' number ')' ]
```

~I put `SampleClause` as a special part of rule in `SelectStmt` instead of `TableRef`, because `TableRef` is also used in `UPDATE/DELETE` statements and we don't want to support `TABLESAMPLE` for them.~

~However, this also prevent the `JOIN` clause to be syntax-checked:~

<del>

```sql
select * from tbl1 tablesample (10), tbl2 tablesample (20); -- syntax error
select * from tbl1 a tablesample (10) join tbl2 b tablesample (20) on a.id <> b.id; -- syntax error
```
 
</del>

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N/A

Side effects

N/A

Related changes

N/A


parser.y
```diff
@@ -7280,6 +7291,74 @@ SelectStmtFromTable:
 		$$ = st
 	}
 
+SelectStmtFromTableSample:
+	SelectStmtBasic "FROM" TableName PartitionNameListOpt TableAsNameOpt "TABLESAMPLE" TableSampleMethodOpt '(' Expression TableSampleUnitOpt ')' RepeatableOpt WhereClauseOptional
```
kennytm, line 7295: it's better attach this to this line to support JOIN:

```bison
TableFactor:
	TableName PartitionNameListOpt TableAsNameOpt IndexHintListOpt TableSampleOpt # <--
```

this means the TableSample is part of the `ast.TableSource` or `ast.TableName`.

their use in UPDATE and DELETE can be rejected after parsing.

ast/dml.go
```diff
@@ -775,6 +797,90 @@ func (n *OrderByClause) Accept(v Visitor) (Node, bool) {
...
+type TableSample struct {
+	node
+	SampleMethod     SampleMethodType
+	Expr             ExprNode
+	SampleClauseUnit SampleClauseUnitType
+	RepeatableSeed   ExprNode
+}
+
+func (s *TableSample) Restore(ctx *format.RestoreCtx) error {
+	ctx.WriteKeyWord(" TABLESAMPLE ")
```
kennytm, line 826: as a rule we have laid down many months ago, the Restore() result should not have any leading or trailing spaces...

```suggestion
	ctx.WriteKeyWord("TABLESAMPLE ")
```

kennytm (commented):
rest LGTM

kennytm:
LGTM

parser.y
```diff
@@ -7805,11 +7888,14 @@ TableRef:
 |	JoinTable
 
 TableFactor:
-	TableName PartitionNameListOpt TableAsNameOpt IndexHintListOpt
+	TableName PartitionNameListOpt TableAsNameOpt IndexHintListOpt TableSampleOpt
```
AilinKid, line 7891: should append a warning? since we don't support them all in TiDB actually...

AilinKid (commented):
Rest LGTM

parser.y
tangenta: Related work is done in TiDB.

AilinKid (approved):
LGTM

ti-srebot (approved):

PR #1080, opened 2020-11-09
model: add functionality to check if a base table

nullnotnil:

This will be used in https://github.com/pingcap/tidb/issues/20880

I added a new function instead of checking for not a sequence or a view, because I think there are other cases that need this that may currently only test for not a view, but should also test for not a sequence.

New function to safely test for is a table.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


kennytm (commented):
Rest LGTM

AilinKid (commented):
LGTM

ti-srebot (approved):

PR #1095, opened 2020-11-13
*: add alter table read only/write support

jackysp:

This PR intends to support Oracle like `alter table read only/write` syntax. It is used to do not release locks automatically when the session exits. Which will make the memory cache easy to effect. Related to https://github.com/pingcap/tidb/pull/20396

### Check List 

Tests 

 - Unit test

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -386,6 +386,9 @@ const (
 	TableLockRead
 	// TableLockReadLocal is not supported.
 	TableLockReadLocal
+	// TableLockReadOnly is used to set a table into read-only status,
+	// when the session exits, it will not release its lock automatically.
+	TableLockReadOnly
 	// TableLockWrite means only the session with this lock has write/read permission.
 	// Only the session that holds the lock can access the table. No other session can access it until the lock is released.
 	TableLockWrite
@@ -401,6 +404,8 @@ func (t TableLockType) String() string {
 		return "READ"
 	case TableLockReadLocal:
 		return "READ LOCAL"
+	case TableLockReadOnly:
+		return "READ ONLY"
 	case TableLockWriteLocal:
 		return "WRITE LOCAL"
 	case TableLockWrite:
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -2160,6 +2160,7 @@ const (
 	AlterTableRenameTable
 	AlterTableAlterColumn
 	AlterTableLock
+	AlterTableWriteable
 	AlterTableAlgorithm
 	AlterTableRenameIndex
 	AlterTableForce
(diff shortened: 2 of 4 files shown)


jackysp:
I would like to extend the `lock tables read only` syntax to ensure that the lock is not cleared when the session exits. Please leave your comments! @crazycs520 @coocood

coocood:
LGTM

crazycs520 (approved):
LGTM

model/model.go
```diff
@@ -386,6 +386,9 @@ const (
 	TableLockRead
 	// TableLockReadLocal is not supported.
 	TableLockReadLocal
+	// TableLockReadOnly is used to set a table into read-only status,
+	// when the session exits, it will not release its lock automatically.
+	TableLockReadOnly
```
kennytm, line 391: How to reset "READ ONLY" status if it won't be released even when the session exits? Quoting [MySQL](https://dev.mysql.com/doc/refman/8.0/en/lock-tables.html) and [MariaDB](https://mariadb.com/kb/en/lock-tables/),

> One session cannot acquire locks for another session or release locks held by another session. 

Meaning we can't use `UNLOCK TABLES` if the current connection holding the READ ONLY lock is dead.


BTW in MySQL, `READ ONLY` is applied via the [`ALTER DATABASE`](https://dev.mysql.com/doc/refman/8.0/en/alter-database.html) statement (no table-level read only switch):

```sql
ALTER DATABASE t READ ONLY = 1;
```

nullnotnil: > I would like to extend the `lock tables read only` syntax to ensure that the lock is not cleared when the session exits. Please leave your comments! @crazycs520 @coocood

As well as the locks being cleared when the session exists, MySQL will also clear all locks on any subsequent `LOCK TABLES` requests. I believe the original intention may have been to prevent having to do deadlock detection? It's hard to tell, but the semantics get messy if some usage has permanance.

I like the syntax @kennytm suggested:
```sql
ALTER DATABASE `db` READ ONLY = 1;
ALTER TABLE t1 READ ONLY = 1;
```

jackysp: We could use `admin cleanup table lock [table name]`, but it is not a good choice. I'll design a unlock SQL syntax for this with @crazycs520 and @coocood

jackysp: Ah, yes, `ALTER TABLE t READ ONLY` and `ALTER TABLE t READ WRITE` looked nice. But I don't know if its semantics match up with what we want.

jackysp:
PTAL @kennytm @nullnotnil @coocood @crazycs520

kennytm (commented):
LGTM

ti-srebot (approved):

Issue #1097, opened 2020-11-16
Add more check on parser part


blacktear23:
## Feature Request

In TiDB there has a lot of separate validation code for syntax check, and some of them can be move to parser component. If we move it to parser component we will got some benefit:

- Less code
- Fully coverage
- Easily find out the missing part

Currently some TiDB bug is related to syntax check which is implements in planner component and it is not easy to find out how many corner case we are missing.

**Describe alternatives you've considered:**

We can move some simple syntax check to `parser.y`, such as `Float`, `Double` and `Decimal` type precision check. And we also can build a mechanism that to determine which syntax validation should implements in parser component.

PR #914, opened 2020-07-05
parser: remove partitioning

wangggong:
### What problem does this PR solve? 

Close #603 (cc #598)

### Bad SQL example

```SQL
ALTER TABLE x REMOVE PARTITIONING;
```

### What is changed and how it works?

Update judgement in `parser.y`.

Tests 

 - Unit test

PR #1099, opened 2020-11-16
parser: add cast check

pengdaqian2020:

[#21064](https://github.com/pingcap/tidb/issues/21063)

add decimal scale and precision check

### Check List

Tests

 - Unit test
 - Integration test
 
Code changes

No
No
No

Side effects

No
 - Increased code complexity
No
No

Related changes

yes
No
No

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -7030,7 +7030,22 @@ CastType:
 	}
 |	"DECIMAL" FloatOpt
 	{
+		startOffset := parser.startOffset(&yyS[yypt-3])
+		endOffset := parser.endOffset(&yyS[yypt-2])
+		colString := parser.src[startOffset:endOffset]
 		fopt := $2.(*ast.FloatOpt)
+		if fopt.Flen > mysql.MaxDecimalWidth {
+			yylex.AppendError(ErrTooBigPrecision.GenWithStackByArgs(fopt.Flen, colString, mysql.MaxDecimalWidth))
+			return 1
+		}
+		if fopt.Decimal > mysql.MaxDecimalScale {
+			yylex.AppendError(ErrTooBigScale.GenWithStackByArgs(fopt.Decimal, colString, mysql.MaxDecimalScale))
+			return 1
+		}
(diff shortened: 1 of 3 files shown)


pengdaqian2020:
@blacktear23 
PLAT

kennytm:
Hi, thanks for the contribution.

Unfortunately I need to push back on this direction of adding semantic checks directly into `parser.y`. As we know `pingcap/parser` is not _just_ used for TiDB. Officially it is also used in DM for parsing MySQL and MariaDB DDLs, and this package is also [used by third-party](https://pkg.go.dev/github.com/pingcap/parser?tab=importedby) for generating AST from SQL. 

Therefore, semantic checks like this PR _only_ benefits TiDB, but limits legit third-party usage (like performing [AST transformation](https://github.com/XiaoMi/soar) which shouldn't be tripped by `DECIMAL(1000,20)`, or maybe building a MyBigNumDB which legit supports `DECIMAL(1000,20)`).

IMO the proper way is to check the AST after parsing, like:

```go
statements, _warns, _err := p.Parse("sql", "", "")
verifier := NewTiDBVerifier() // bundle all TiDB-specific semantic checks into this Visitor
for _, stmt := range statements {
    if _, ok := stmt.Accept(verifier); !ok {
        return verifier.Err()
    }
}
```

The idea of placing the semantic checks directly in `parser.y` to generate errors and warnings has been tried several times and also moved back to TiDB later (#965, #888, #795, #920, #914). Let's not repeat this again.

cc #1097 @blacktear23

blacktear23:
@kennytm Agreed

blacktear23:
@pengdaqian2020 sorry for my misleading.

pengdaqian2020:
> @pengdaqian2020 sorry for my misleading.

:joy: so it seems I should check it in `preprocessor` ?

blacktear23:
> > @pengdaqian2020 sorry for my misleading.
> 
> 😂 so it seems I should check it in `preprocessor` ?

Yes. 😂

PR #1104, opened 2020-11-23
model: Add StateGlobalTxnWriteOnly

Yisaer:

ref https://github.com/pingcap/tidb/issues/21119 https://github.com/pingcap/tidb/pull/21356

Add new Schema state `StateGlobalTxnWriteOnly`

### Check List 

Tests 

 - Unit test


Yisaer:
@djshow832 PTAL

model/model.go
```diff
@@ -764,13 +768,39 @@ func (pi *PartitionInfo) GetNameByID(id int64) string {
...
-	ID       int64      `json:"id"`
-	Name     CIStr      `json:"name"`
-	LessThan []string   `json:"less_than"`
-	InValues [][]string `json:"in_values"`
-	Comment  string     `json:"comment,omitempty"`
+	ID       int64       `json:"id"`
+	Name     CIStr       `json:"name"`
+	LessThan []string    `json:"less_than"`
+	InValues [][]string  `json:"in_values"`
+	State    SchemaState `json:"state"`
```
djshow832, line 802: There is a compatibility problem here. After upgrading TiDB from an older version, since there was no field `State` on TiKV, so the `State` field is filled with 0 value, which corresponds to `StateNone`. However, `StateNone` means neither readable nor writable.

AilinKid: yes @tiancaiamao @bb7133  PTAL

Yisaer: use `PartitonState` instead.

model/model.go
```diff
@@ -764,6 +769,31 @@ func (pi *PartitionInfo) GetNameByID(id int64) string {
...
+func (pi *PartitionInfo) SetStateByID(id int64, state SchemaState) bool {
+	for i, pstate := range pi.States {
+		if pstate.ID == id {
+			pstate.State = state
+			pi.States[i] = pstate
+			return true
+		}
+	}
+	return false
+}
```
djshow832, line 790: `States` is empty by default. If you set one of them to `StateGlobalTxnWriteOnly`, it always fails.

Yisaer: I think we can init `states` in `tidb` repo while  `PartitionInfo` can provide some useful methods here.

djshow832: Alternatively, you can save the states that are not `StatePublic` in `pi.States`.
To this end, just append a new `PartitionState` if it's not found in `pi.States`.

model/model.go
```diff
@@ -764,6 +769,31 @@ func (pi *PartitionInfo) GetNameByID(id int64) string {
...
+		}
+	}
+	return StateNone, false
+}
+
+func (pi *PartitionInfo) SetStateByID(id int64, state SchemaState) bool {
+	for i, pstate := range pi.States {
+		if pstate.ID == id {
+			pstate.State = state
+			pi.States[i] = pstate
```
djshow832, line 785: Why set it back?

model/model.go
djshow832: Is there any practical method to init `States`? Sometimes `TableInfo` is initialized by `json.Unmarshal`, and it's not elegant to fill `States` after unmarshalling.

Yisaer: `To this end, just append a new PartitionState if it's not found in pi.States` I think this is a good idea.

model/model.go
```diff
@@ -764,6 +769,54 @@ func (pi *PartitionInfo) GetNameByID(id int64) string {
...
+			return
+		}
+	}
+	if pi.States == nil {
+		pi.States = make([]PartitionState, 0, 1)
+	}
+	pi.States = append(pi.States, newState)
+}
+
+func (pi *PartitionInfo) GCPartitionStates() {
```
djshow832, line 795: When will you call it?

Yisaer: Haven't decided yet. Maybe when a partition is being dropped, we can call this function to gc the unmaintained parition states.

model/model.go
```diff
@@ -748,6 +752,7 @@ type PartitionInfo struct {
 	AddingDefinitions []PartitionDefinition `json:"adding_definitions"`
 	// DroppingDefinitions is filled when dropping a partition that is in the mid state.
 	DroppingDefinitions []PartitionDefinition `json:"dropping_definitions"`
+	States              []PartitionState      `json:"states"`
```
AilinKid, line 755: should keep the states element offset be consistent with the definition?

Yisaer: I think this is unnecessary.

djshow832 (commented):
LGTM

AilinKid (approved):
LGTM

ti-srebot (approved):

PR #1105, opened 2020-11-25
*: add a flag `RestoreWithDefaultDB` for Restore

rebelice:
It's a part of [tidb-pr](https://github.com/pingcap/tidb/pull/21275)

- add a flag `RestoreWithDefaultDB` for Restore
- modify some format case

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lzmhhh123 (commented):
LGTM

ast/dml.go
```diff
@@ -983,6 +986,9 @@ func (n *SelectStmt) Restore(ctx *format.RestoreCtx) error {
 		if n.TableHints != nil && len(n.TableHints) != 0 {
 			ctx.WritePlain("/*+ ")
 			for i, tableHint := range n.TableHints {
+				if i != 0 {
+					ctx.WritePlain(", ")
+				}
```
kennytm, line 991: No, table hints *are* supposed to be separated by spaces, not commas. Demo on MySQL 8:

```
mysql> select /*+ set_var(sql_mode='x') set_var(timestamp=1) */ @@timestamp;
+-------------+
| @@timestamp |
+-------------+
|    1.000000 |
+-------------+
1 row in set, 1 warning (0.00 sec)

mysql> select /*+ set_var(sql_mode='x'), set_var(timestamp=1) */ @@timestamp;
+-------------------+
| @@timestamp       |
+-------------------+
| 1606280193.327941 |
+-------------------+
1 row in set, 2 warnings (0.00 sec)
```

kennytm (changes requested):

ast/dml.go
rebelice: But we use commas in our [user docs](https://docs.pingcap.com/zh/tidb/stable/optimizer-hints), so we need to modify the user docs?

rebelice:
I have modified， PTAL @kennytm

ast/dml.go
kennytm: better modify the docs to align with MySQL. the optional comma can be thought as TiDB's syntax extension.

format/format.go
```diff
@@ -278,16 +282,25 @@ func (rf RestoreFlags) HasSpacesAroundBinaryOperationFlag() bool {
 	return rf.has(RestoreSpacesAroundBinaryOperation)
 }
 
+func (rf RestoreFlags) HasWithDefaultDB() bool {
+	return rf.has(RestoreWithDefaultDB)
```
kennytm, line 286: IMO returning `rf.DefaultDB != ""` is enough, no need to introduce a flag

format/format.go
```diff
@@ -278,16 +282,25 @@ func (rf RestoreFlags) HasSpacesAroundBinaryOperationFlag() bool {
...
+}
+
+func (rf RestoreFlags) HasStringWithoutDefaultCharset() bool {
+	return rf.has(RestoreStringWithoutDefaultCharset)
+}
```
kennytm, line 291: this doesn't seem to have any effect

kennytm (commented):
Rest LGTM

format/format.go
rebelice: This is used in TiDB, related pr is [here](https://github.com/pingcap/tidb/pull/21275)

format/format.go
rebelice: Done.

lzmhhh123 (approved):

lzmhhh123:
The test failed due to an unstable test in the planner.

PR #1122, opened 2020-12-08
*: support multiple select options

wjhuang2016:
Close issue https://github.com/pingcap/tidb/issues/20133

Add SelectStmtOptsList, and merge options if there is more than one option.

Note: After this PR merge.
`select SQL_BUFFER_RESULT SQL_BUFFER_RESULT  from t;` would cause parser error. This is the same as MySQL.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -1315,11 +1315,11 @@ import (
 	TextString                      "text string item"
 
 %precedence empty
+%precedence lowerThanSelectOpt
 %precedence sqlBufferResult
 %precedence sqlBigResult
 %precedence sqlSmallResult
 %precedence sqlCache sqlNoCache
-%precedence higherThanSelectOpt
 %precedence lowerThanIntervalKeyword
 %precedence interval
 %precedence next
@@ -8269,7 +8269,12 @@ SelectStmtOpt:
 	{
 		opt := &ast.SelectStmtOpts{}
 		opt.SQLCache = true
-		opt.Distinct = $1.(bool)
+		if $1.(bool) {
+			opt.Distinct = true
+		} else {
+			opt.Distinct = false
+			opt.ExplicitAll = true
+		}
 		$$ = opt
 	}
 |	Priority
@@ -8328,7 +8333,7 @@ SelectStmtOpts:
 		opt.SQLCache = true
 		$$ = opt
 	}
-|	SelectStmtOptsList %prec higherThanSelectOpt
+|	SelectStmtOptsList %prec lowerThanSelectOpt
 
 SelectStmtOptsList:
 	SelectStmtOptsList SelectStmtOpt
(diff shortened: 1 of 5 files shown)


parser.y
```diff
@@ -8249,47 +8257,133 @@ SelectStmtLimitOpt:
...
+	{
+		opt := &ast.SelectStmtOpts{}
+		opt.SQLCache = true
+		opt.StraightJoin = $1.(bool)
+		$$ = opt
+	}
+
 SelectStmtOpts:
-	TableOptimizerHints DefaultFalseDistinctOpt PriorityOpt SelectStmtSQLSmallResult SelectStmtSQLBigResult SelectStmtSQLBufferResult SelectStmtSQLCache SelectStmtCalcFoundRows SelectStmtStraightJoin
+	%prec empty
```
tangenta, line 8330: If we change to `SelectStmtOptsList`, the order of hints & selection options can be arbitrary. Is it expected?

wjhuang2016: Yeah. I test it in MySQL.

parser.y
```diff
@@ -8251,47 +8259,134 @@ SelectStmtLimitOpt:
...
+		opt.Priority = $1.(mysql.PriorityEnum)
+		$$ = opt
+	}
+|	SelectStmtSQLSmallResult
+	{
+		opt := &ast.SelectStmtOpts{}
+		opt.SQLCache = true
+		opt.SQLSmallResult = $1.(bool)
+		$$ = opt
+	}
```
kennytm, line 8295: just use the keyword directly? it's not like the non-terminal can produce anything other than `true`.

```suggestion
|	"SQL_SMALL_RESULT"
	{
		opt := &ast.SelectStmtOpts{}
		opt.SQLCache = true
		opt.SQLSmallResult = true
		$$ = opt
	}
```

(ditto for the rest of this list)

kennytm (commented):
Rest LGTM

parser_test.go
```diff
@@ -100,7 +100,7 @@ func (s *testParserSuite) TestSimple(c *C) {
 		"max_rows", "min_rows", "national", "quarter", "escape", "grants", "status", "fields", "triggers", "language",
 		"delay_key_write", "isolation", "partitions", "repeatable", "committed", "uncommitted", "only", "serializable", "level",
 		"curtime", "variables", "dayname", "version", "btree", "hash", "row_format", "dynamic", "fixed", "compressed",
-		"compact", "redundant", "sql_no_cache sql_no_cache", "sql_cache sql_cache", "action", "round",
+		"compact", "redundant", "`sql_no_cache`", "`sql_cache`", "action", "round",
```
kennytm, line 103:
```suggestion
		"compact", "redundant", "1 sql_no_cache`", "1 sql_cache", "action", "round",
```

quoting defeats the whole purpose of this test.

kennytm (commented):
rest LGTM

tangenta (commented):
LGTM

kennytm:
LGTM

ti-srebot (approved):

PR #1123, opened 2020-12-09
parser: fix aggregate functions compatibility to MySQL8.0

### LINKED ISSUE #1116: var_pop|stddev_pop|stddev_samp with `DISTINCT` should raise an syntax error
## Bug Report

More details PTAL https://github.com/pingcap/tidb/issues/19499

hidehalo:

close #1116 

change yacc rules of `SumExpr` below :

- VAR_POP
- VAR_SAMP
- VARIANCE
- STDDEV_POP
- STDDEV_SAMP

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


go.mod
```diff
@@ -1,19 +1,26 @@
 module github.com/pingcap/parser
 
 require (
+	github.com/cznic/fileutil v0.0.0-20181122101858-4d67cfea8c87 // indirect
```
kennytm, line 4: i don't think you need to modify `go.mod` and `go.sum`.

parser_test.go
```diff
@@ -3243,6 +3243,48 @@ func (s *testParserSuite) TestErrorMsg(c *C) {
 
 	_, _, err = parser.Parse("select 1 collate some_unknown_collation", "", "")
 	c.Assert(err.Error(), Equals, "[ddl:1273]Unknown collation: 'some_unknown_collation'")
+
+	_, _, err = parser.Parse("select std(all c1) from t", "", "")
```
kennytm, line 3247: if you're truly aiming for MySQL 8.0 compatibility then `all` *must* still be supported.

```
mysql> select version();
+-----------+
| version() |
+-----------+
| 8.0.20    |
+-----------+
1 row in set (0.00 sec)

mysql> select std(all a) from a;
+------------+
| std(all a) |
+------------+
|        0.5 |
+------------+
1 row in set (0.00 sec)

mysql> select std(distinct a) from a;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'distinct a) from a' at line 1
```

go.mod
hidehalo: 👌

hidehalo:
I think it should close for now, if it is necessary for related issue, we just let it re-open.

PR #1125, opened 2020-12-11
parser: disallow into-outfile clause in some place

tangenta:
Part of pingcap/tidb#19097.

This PR validate the 'into outfile' option. This clause is only allowed in the last select statement.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


kennytm:
I'd prefer we do this kind of check outside of `parser.y` as a separation of concern. Hmm perhaps the semantic checker can really be moved from pingcap/tidb into pingcap/parser.

MySQL 8 and MariaDB 10.4 have different error messages for these errors. And I think the MySQL 8 is clearer.

In MySQL 8, using SELECT INTO in the wrong place generates the 3954 error instead of the standard 1064 "syntax error" (this PR):

```
mysql> (select 1 into outfile '1.txt') union select 1;
ERROR 3954 (HY000): Misplaced INTO clause, INTO is not allowed inside subqueries, and must be placed at end of UNION clauses.
```

also CREATE VIEW AS SELECT INTO generates the 1350 error:

```
mysql> create view c as select 1 into outfile '1.txt';
ERROR 1350 (HY000): View's SELECT contains a 'INTO' clause
```

Thus I suspect MySQL is checking after the AST is built, not during. 

Perhaps as a side effect, MySQL allows the following code which has been parse errors in MariaDB:

```sql
-- NO ERROR, the 'into outfile' part is just ignored (looks wrong to me)
insert into a select 1, 2 into outfile '1.txt';
create table c as select 1 into outfile '1.txt';

-- same as removing the parenthesis
(((((select 1 into outfile '1.txt')))));
```

BTW the following should work:

```sql
create binding for 
    select 1 into outfile '1.txt'
using 
    select 1 into outfile '1.txt';
```

wjhuang2016:
I agree with kenny,   semantic check should be done in the TiDB part.

tangenta:
Will be fixed by https://github.com/pingcap/tidb/pull/21956.

PR #1129, opened 2020-12-17
ast, parser: fix parse join 

wjhuang2016:

Fix https://github.com/pingcap/tidb/issues/20563
We can't parse `t1 join t2 join t3 on xxx on xxx`.
Because the order is: `(t1 join t2 join t3 on xxx) on xxx`. What we need is `t1 join (t2 join t3 on xxx) on xxx`

1. Make `join straightJoin inner cross left right full natural` below `tableRefPriority`, which is the same as MySQL.
2. Implement NewCrossJoin(), here is its logic:
```
// NewCrossJoin builds a cross join without `on` or `using` clause.
// If the right child is a join tree, we need to handle it differently to make the precedence get right.
// Here is the example: t1 join t2 join t3
//                 JOIN ON t2.a = t3.a
//  t1    join    /    \
//              t2      t3
// (left)         (right)
//
// We can not build it directly to:
//         JOIN
//        /    \
//       t1	   JOIN ON t2.a = t3.a
//             /   \
//            t2    t3
// The precedence would be t1 join (t2 join t3 on t2.a=t3.a), not (t1 join t2) join t3 on t2.a=t3.a
// We need to find the left-most child of the right child, and build a cross join of the left-hand side
// of the left child(t1), and the right hand side with the original left-most child of the right child(t2).
//          JOIN t2.a = t3.a
//         /    \
//       JOIN    t3
//       /  \
//      t1  t2
// Besides, if the right handle side join tree's join type is right join, we need to rewrite it to left join.
// So t1 join t2 right join t3 would be rewrite to t1 join t3 left join t2.
// If not, t1 join (t2 right join t3) would be (t1 join t2) right join t3. After rewrite the right join to left join.
// We get (t1 join t3) left join t2, the semantics is correct.
```

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


winoros (commented):
lgtm

parser.y
```diff
@@ -1338,10 +1338,10 @@ import (
 %precedence remove
 %precedence lowerThenOrder
 %precedence order
-%left join straightJoin inner cross left right full natural
 
 /* A dummy token to force the priority of TableRef production in a join. */
 %left tableRefPriority
+%left join straightJoin inner cross left right full natural
```
kennytm, line 1344: would it be easier to make the JOIN operator right-associative? 🤔

wjhuang2016: NewCrossJoin() will do the reorganization, make it left-associative.

kennytm: i mean if you use `%right` here perhaps you don't need that `NewCrossJoin()` function

wjhuang2016: Then  `t1 right join t2 right join t3` would be `t1 right join (t2 right join t3)`, it's not correct.

kennytm: why is that incorrect? `a RIGHT JOIN b RIGHT JOIN c ON w1 ON w2` is indeed equivalent to `a RIGHT JOIN (b RIGHT JOIN c ON w1) ON w2`, 
and the same for LEFT JOIN (`a LEFT JOIN b LEFT JOIN c ON w1 ON w2` == `a LEFT JOIN (b LEFT JOIN c ON w1) ON w2`)

wjhuang2016: Sorry, the example is wrong.
Consider `select * from t1 join t2 right join t3 on t2.a=t3.a;`, if using right-associative, we got 
`select * from t1 join (t2 right join t3 on t2.a=t3.a)`. They are not the same.
```
select * from t1 join t2 join t3;
mysql> select * from t1 join t2 right join t3 on t2.a=t3.a;
+------+------+------+
| a    | a    | a    |
+------+------+------+
| NULL | NULL |    3 |
+------+------+------+
1 row in set (0.00 sec)

mysql> select * from t1 join (t2 right join t3 on t2.a=t3.a);
+------+------+------+
| a    | a    | a    |
+------+------+------+
|    1 | NULL |    3 |
+------+------+------+
1 row in set (0.00 sec)
```

tangenta (commented):
LGTM

ti-srebot (approved):

PR #1133, opened 2020-12-19
Add CannotInsertFlag to indicate that this table cannot be inserted

sev7ndayyoo:

Include `no_zero_date` in `sqlmode`, these scenarios should fail to `add column`:
```
create table t (id int);
insert into t values(1);
alter table t add column a date not null;
alter table t add column a date not null;
```
Considering that data may be written during the DDL state change phase, need a flag to prevent data insert during this period. So add this flag.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes


sev7ndayyoo:
@bb7133  PTAL

PR #1136, opened 2020-12-22
ast, parser: add FUNCTION and PROCEDURE for GRANT object_type

lance6716:
parser didn't support `GRANT ... ON FUNCTION ...`, where DM plans to parse and check privileges in https://github.com/pingcap/dm/issues/1320

ref https://dev.mysql.com/doc/refman/5.7/en/grant.html

add `ObjectTypeFunction` and `ObjectTypeProcedure`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -1338,6 +1338,8 @@ import (
 %precedence remove
 %precedence lowerThenOrder
 %precedence order
+%precedence lowerThanFunction
+%precedence function
 
 /* A dummy token to force the priority of TableRef production in a join. */
 %left tableRefPriority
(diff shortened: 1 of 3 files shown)


parser.y
```diff
@@ -11774,13 +11776,22 @@ PrivType:
 	}
 
 ObjectType:
+	%prec lowerThanFunction
```
lance6716, line 11779: function is not a reserved keywork in 5.7 (but become reserved after 8.0.1). see https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-F

here I didn't change it to reserved to keep compatibility

lance6716:
PTAL @wjhuang2016 @leoppro @xhebox

xhebox (approved):
LGTM

kennytm (approved):
LGTM

ti-srebot (approved):

PR #986, opened 2020-08-27
yy_parser: enable supportWindowFunc by default

tangenta:
### What problem does this PR solve? 
Close https://github.com/pingcap/parser/issues/959.

As suggested by https://github.com/pingcap/parser/issues/959#issuecomment-668383316.

### What is changed and how it works?
Enable parse window function by default.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

N/A

Side effects

N/A

Related changes

N/A

PR #1137, opened 2020-12-23
misc: move keyword 'ROW' from 'tokenMap' to 'windowFuncTokenMap'

bb7133:
Make it possible to use 'ROW' as unreserved keyword, which is useful for users to make migration from MySQL 5.7

move keyword 'ROW' from 'tokenMap' to 'windowFuncTokenMap'.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -577,7 +577,6 @@ var tokenMap = map[string]int{
 	"ROUTINE":                  routine,
 	"ROW_COUNT":                rowCount,
 	"ROW_FORMAT":               rowFormat,
-	"ROW":                      row,
 	"ROWS":                     rows,
 	"RTREE":                    rtree,
 	"RESUME":                   resume,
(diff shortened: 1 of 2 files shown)


kennytm:
Note that EnableWindowFunc is true by default since #986. So unit test passing doesn't validate this PR.

When it is disabled, as your changes to the test cases revealed, this affects not just `ROW` when using with window function, but also:

```sql
-- ROW constructor (supported in MySQL 5.x)
select row(1,2) = row(3,4);

-- VALUES statement (supported in MySQL 8.0)
values row(1);

-- FETCH FIRST clause (TiDB extension for standard SQL compatibility)
select 1 fetch first row only;
```

We need to keep "ROW" in the tokenMap. But moving the token to UnReservedKeyword should be sufficient.

PR #1139, opened 2020-12-25
util: Add PrepareStmt into readOnly

Yisaer:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/ast/util.go
+++ b/ast/util.go
@@ -55,6 +55,8 @@ func IsReadOnly(node Node) bool {
 			}
 		}
 		return true
+	case *PrepareStmt:
+		return true
 	default:
 		return false
 	}
--- a/ast/util_test.go
+++ b/ast/util_test.go
@@ -76,6 +76,9 @@ func (s *testCacheableSuite) TestCacheable(c *C) {
 
 	stmt = &ShowStmt{}
 	c.Assert(IsReadOnly(stmt), IsTrue)
+
+	stmt = &PrepareStmt{}
+	c.Assert(IsReadOnly(stmt), IsTrue)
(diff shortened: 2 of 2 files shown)


kennytm (commented):
lgtm

tangenta (approved):
LGTM

ti-srebot (approved):

tangenta:
@Yisaer I just noticed that the integration test failure is related to this PR.

```
FAIL: session_test.go:1153: testSessionSuite.TestAutoIncrementWithRetry
```

Perhaps `PrepareStmt` should not be read-only because it modifies the session variable [here](https://github.com/pingcap/tidb/blob/5825c56764b14b4372a2466b2365c146a9ce2a8c/executor/prepared.go#L212). And the function `IsReadOnly` is also used for [statement retry](https://github.com/pingcap/tidb/blob/5825c56764b14b4372a2466b2365c146a9ce2a8c/session/tidb.go#L201): only the statements that are not `IsReadOnly` will be recorded in the history.

Since `PrepareStmt` is read-only now, it will not be recorded in the history. As a result, the case below may fail:
```sql
begin;
prepare stmt from 'insert into t (c2) values (?)';
...
execute stmt using @v1;
...
commit;
```

When a write conflict occurs, it will retry and throw an error like
```
ERROR 8111 (HY000): Prepared statement not found
```

Yisaer:
@tangenta  You are right. It seems `PrepareStmt` shouldn't be thought as `readOnly`

PR #1162, opened 2021-01-28
Add "START TRANSACTION WITH CAUSAL CONSISTENCY ONLY"

ekexium:
Signed-off-by: ekexium <ekexium@gmail.com>

See pingcap/tidb#22597

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


kennytm:
if external consistency (> strict serializability) is the default and is here considered orthogonal to RU/RC/RR stuff, what happens if i do

```sql
START TRANSACTION /* WITH EXTERNAL CONSISTENCY */;
SET TRANSACTION ISOLATION LEVEL READ COMMITTED;
...
```

?

ekexium:
@kennytm External consistency doesn't mean strict serializability. It is orthogonal to isolation levels. 

So I think it should not interfere with Read Committed.

TiDB used to guarantee external consistency. The new feature _async commit_ can break it (for better performance), so we need a new grammar to control the behavior.

sticnarf:
@kennytm Right... External consistency can be an ambiguous word.
Do you think we can use `START TRANSACTION WITH CAUSAL CONSISTENCY`?

kennytm:
So to clarify, is "external consistency" here the same concept as the one mentioned in https://cloud.google.com/spanner/docs/true-time-external-consistency and https://stackoverflow.com/questions/60365103/differences-between-strict-serializable-and-external-consistency (which both claimed it is stronger than linearizability + serializability)

sticnarf:
> So to clarify, is "external consistency" here the same concept as the one mentioned in https://cloud.google.com/spanner/docs/true-time-external-consistency and https://stackoverflow.com/questions/60365103/differences-between-strict-serializable-and-external-consistency (which both claimed it is stronger than linearizability + serializability)

No, the "external consistency" here is roughly the same with linearizability. It has nothing to do with serializability.

ast/misc.go
```diff
@@ -384,9 +384,10 @@ func (n *ExecuteStmt) Accept(v Visitor) (Node, bool) {
 // See https://dev.mysql.com/doc/refman/5.7/en/commit.html
 type BeginStmt struct {
 	stmtNode
-	Mode     string
-	ReadOnly bool
-	Bound    *TimestampBound
+	Mode                       string
+	ReadOnly                   bool
+	Bound                      *TimestampBound
+	WithoutExternalConsistency bool
```
sticnarf, line 390: Change the name to be consistent with the statement?

ekexium: The word "external consistency" is used in tidb code. Do you think we should change them as well?

sticnarf: Yes, I think so. Let's change to use the more precise word "causal consistency" in TiDB.

parser.y
```diff
@@ -2620,7 +2620,7 @@ BeginTransactionStmt:
 |	"START" "TRANSACTION" "WITH" "CAUSAL" "CONSISTENCY"
```
kennytm, line 2620: maybe make the statement <code>START TRANSACTION WITH CAUSAL CONSISTENCY **ONLY**</code> to emphasize it is a downgrade from linearizability.

youjiali1995 (approved):

ti-srebot (approved):

PR #1164, opened 2021-02-02
ast: support `explain format = 'test'` syntax

eurekaka:

Parser PR for https://github.com/pingcap/tidb/issues/22663

Add a new format for EXPLAIN.

### Check List 

Tests 

 - Unit test

Code changes

N/A

Side effects

N/A

Related changes

### DIFF
--- a/ast/misc.go
+++ b/ast/misc.go
@@ -69,6 +69,7 @@ const (
 	ExplainFormatJSON    = "json"
 	ExplainFormatHint    = "hint"
 	ExplainFormatVerbose = "verbose"
+	ExplainFormatBrief   = "brief"
 	PumpType             = "PUMP"
 	DrainerType          = "DRAINER"
 )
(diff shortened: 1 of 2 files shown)


ast/misc.go
```diff
@@ -69,6 +69,7 @@ const (
 	ExplainFormatJSON    = "json"
 	ExplainFormatHint    = "hint"
 	ExplainFormatVerbose = "verbose"
+	ExplainFormatTest    = "test"
```
qw4990, line 72: Does `test` seems informal and too casual? How about `brief` or some names more specified like `without-id`?

qw4990 (commented):
LGTM

qw4990:
I think it's an internal feature so we don't need to update the doc.

time-and-fate (commented):
LGTM

ti-srebot (approved):

PR #1151, opened 2021-01-12
lexer: quote identifier when get sql digest

xiongjiwei:
### What problem does this PR solve? 

- quote identifier with backquote. 
- digest of `underscoreCS` should be (_charset)
https://github.com/pingcap/tidb/issues/14241

### Check List 

Tests 

 - Unit test

PR #1166, opened 2021-02-14
parser: regard NULL as literal when normalizing query

nayuta-yanagisawa:

Currently, `Normalize()` converts `SELECT NULL` to `select null`, while NULL is regarded as a literal in MySQL, as described in the [documentation](https://dev.mysql.com/doc/refman/8.0/en/literals.html). It seems natural that `SELECT NULL` would be converted to `select ?` from the consistency viewpoint.

A famous query normalization tool, [pt-fingerprint](https://www.percona.com/doc/percona-toolkit/LATEST/pt-fingerprint.html), for MySQL also treats NULL in that way.

```
$ pt-fingerprint --query "SELECT NULL"
select ?
```

Let `isLit()` regards `NULL` and `\N` as a literal.

Note that checking if `tok == null` in `isLit()` is not enough because `scan()` recognize `\N` as `null` but `NULL` as `identifier`. We should also check if `strings.ToLower(t.lit) == "null"`. This seems not so strange, in my view, since `scan()` just roughly classifies tokens.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/digester.go
+++ b/digester.go
@@ -338,6 +338,8 @@ func (d *sqlDigester) isLit(t token) (beLit bool) {
 		beLit = true
 	} else if t.lit == "*" {
 		beLit = true
+	} else if tok == null || strings.ToLower(t.lit) == "null" {
+		beLit = true
 	}
 	return
 }
--- a/digester.go
+++ b/digester.go
@@ -338,7 +338,7 @@ func (d *sqlDigester) isLit(t token) (beLit bool) {
 		beLit = true
 	} else if t.lit == "*" {
 		beLit = true
-	} else if tok == null || strings.ToLower(t.lit) == "null" {
+	} else if tok == null || (tok == identifier && strings.ToLower(t.lit) == "null") {
 		beLit = true
 	}
 	return
--- a/digester_test.go
+++ b/digester_test.go
@@ -31,6 +31,7 @@ func (s *testSQLDigestSuite) TestNormalize(c *C) {
 		{"SELECT 1", "select ?"},
 		{"select null", "select ?"},
 		{"select \\N", "select ?"},
+		{"SELECT `null`", "select null"},
 		{"select * from b where id = 1", "select * from b where id = ?"},
 		{"select 1 from b where id in (1, 3, '3', 1, 2, 3, 4)", "select ? from b where id in ( ... )"},
 		{"select 1 from b where id in (1, a, 4)", "select ? from b where id in ( ? , a , ? )"},
--- a/digester_test.go
+++ b/digester_test.go
@@ -29,6 +29,8 @@ func (s *testSQLDigestSuite) TestNormalize(c *C) {
 		expect string
 	}{
 		{"SELECT 1", "select ?"},
+		{"select null", "select ?"},
+		{"select \\N", "select ?"},
 		{"select * from b where id = 1", "select * from b where id = ?"},
 		{"select 1 from b where id in (1, 3, '3', 1, 2, 3, 4)", "select ? from b where id in ( ... )"},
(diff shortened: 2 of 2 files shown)


digester.go
```diff
@@ -338,6 +338,8 @@ func (d *sqlDigester) isLit(t token) (beLit bool) {
 		beLit = true
 	} else if t.lit == "*" {
 		beLit = true
+	} else if tok == null || strings.ToLower(t.lit) == "null" {
```
kennytm, line 341:
```suggestion
	} else if tok == null {
```

Both `\N` and `NULL` result in the same token.

digester_test.go
```diff
@@ -29,6 +29,8 @@ func (s *testSQLDigestSuite) TestNormalize(c *C) {
 		expect string
 	}{
 		{"SELECT 1", "select ?"},
+		{"select null", "select ?"},
+		{"select \\N", "select ?"},
```
kennytm, line 33: please add a test case for:

```sql
SELECT `null`
```

since `` `null` `` is a quoted identifier this should *not* be normalized to `select ?`.

nayuta-yanagisawa:
@kennytm  Thank you very much for your comments. I agree with them and fixed the code according to them.

kennytm:
LGTM. 

I think we should merge #1151 first so the normalization of `` select `null` `` won't produce the wrong SQL.

digester.go
```diff
@@ -338,6 +338,8 @@ func (d *sqlDigester) isLit(t token) (beLit bool) {
 		beLit = true
 	} else if t.lit == "*" {
 		beLit = true
+	} else if tok == null || (tok == identifier && strings.ToLower(t.lit) == "null") {
```
djshow832, line 341: ``` `null` ``` should be normalized as ``` `null` ```, not `null`.

What it looks like in MySQL:
```sql
mysql> create table `null`(id int);
Query OK, 0 rows affected (0.04 sec)

mysql> select digest_text from performance_schema.events_statements_summary_by_digest where query_sample_text like 'create table `null`%';
+--------------------------------------+
| digest_text                          |
+--------------------------------------+
| CREATE TABLE `null` ( `id` INTEGER ) |
+--------------------------------------+
1 row in set (0.00 sec)
```

digester_test.go
```diff
@@ -29,6 +29,9 @@ func (s *testSQLDigestSuite) TestNormalize(c *C) {
...
+		{"SELECT `null`", "select null"},
```
djshow832, line 34: Should be ```select `null` ```.

digester.go
nayuta-yanagisawa: Thank you for your comment. I agree with you. This would be fixed once #1151 is merged. So, it would be better to merge the present PR after #1151.

digester_test.go
nayuta-yanagisawa: I will fix the test case once #1151 is merged.

nayuta-yanagisawa:
I merged the master branch, but the integration tests fail until tidb's tests are fixed according to #1151.

kennytm:
PTAL @AilinKid @djshow832

nayuta-yanagisawa:
All tests now pass. PTAL @AilinKid @djshow832

nayuta-yanagisawa:
~All tests now pass.~ Sorry, it was not true. I will check it.

kennytm:
you could ignore the integration test for now... to fix it you need to update TiDB itself

nayuta-yanagisawa:
I see. Thank you for your advice.

xiongjiwei (approved):

ti-srebot (approved):

PR #1169, opened 2021-02-19
Changed ActionType def from hardcode to iota

2014BDuck:
N/A. I assume it's better to slightly change those definitions to match the way we define other enum constants.

Changed the way to define DDL action const from hardcoding to iota.

```Go
// From
const (
	ActionNone         ActionType = 0
	ActionCreateSchema ActionType = 1
	ActionDropSchema   ActionType = 2
)

//To
const (
	ActionNone ActionType = iota
	ActionCreateSchema
	ActionDropSchema
)
```

### Check List 

Tests 

- N/A

Side effects

- N/A

Related changes

- N/A

### DIFF
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -30,54 +30,54 @@ type ActionType byte
 
 // List DDL actions.
 const (
-	ActionNone                          ActionType = 0
-	ActionCreateSchema                  ActionType = 1
-	ActionDropSchema                    ActionType = 2
-	ActionCreateTable                   ActionType = 3
-	ActionDropTable                     ActionType = 4
-	ActionAddColumn                     ActionType = 5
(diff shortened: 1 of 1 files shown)


tangenta:
@2014BDuck The explicit way prevents you from adding a new type [in the middle](https://github.com/pingcap/parser/pull/952/files), which breaks the compatibility. 

What's more, it is more convenient to search the meaning of a number produced by log.

2014BDuck:
> @2014BDuck The explicit way prevents you from adding a new type [in the middle](https://github.com/pingcap/parser/pull/952/files), which breaks the compatibility.
> 
> What's more, it is more convenient to search the meaning of a number produced by log.

OK. That's convincing. I will close this PR then.

PR #1173, opened 2021-02-22
restore: fix the bug about restore join table generate syntax error SQL

xiongjiwei:

```sql
ts = TableSource

select * from (t1) join t2;
select * from t1, t2;              select * from t1 join t2;
    join                                    join
    /  \                                    /  \
  join  ts                                 ts  ts
   |
   ts

select * from t1 join (t2);      select * from (t1) join (t2);
      join                                  join
      /  \                                  /  \
     ts  join                             join join
           |                               |     |
          ts                               ts    ts
```

there are 4 kinds of `join` AST, and SQL `select * from t1, t2; ` will restore to `select * from (t1) join t2;`. 
it is fine if `t1` is `TableName` or `Join`, but if `t1` is a `select` stmt, it will generate SQL like 
```sql
select * from (select * from t) t1, t2; ==> select * from ((select * from t) t1) join t2;
```
the result SQL will be reported syntax error in both MySQL8.0 and tidb. the root cause is that if `TableSource` is `SelectStmt` or `SetOprStmt`, we cannot use parentheses enclosed. 

### Check List 

Tests 

 - Unit test

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -145,15 +145,27 @@ func NewCrossJoin(left, right ResultSetNode) (n *Join) {
 
 // Restore implements Node interface.
 func (n *Join) Restore(ctx *format.RestoreCtx) error {
-	if ctx.JoinLevel != 0 {
-		ctx.WritePlain("(")
-		defer ctx.WritePlain(")")
+	useCommaJoin := false
+	_, leftIsJoin := n.Left.(*Join)
+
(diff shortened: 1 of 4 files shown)


leoppro (commented):
LGTM

tangenta (commented):
LGTM

ti-srebot (approved):

PR #1181, opened 2021-03-03
statistics: introduce a new kind of syntax to drop global-stats

qw4990:
Before this PR, we already have the syntax `drop stats {table-name} partition {partition-name}` to drop partition-stats.
And if we want to drop global-stats, we now use the keyword `global` to replace `{partition-name}` like `drop stats t partition global`.
But this solution may cause a conflict if a user names a partition as `global`.
To solve this conflict, this PR introduces a new kind of syntax `drop stats {table-name} global`.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/ast/stats.go
+++ b/ast/stats.go
@@ -176,6 +176,7 @@ type DropStatsStmt struct {
 
 	Table          *TableName
 	PartitionNames []model.CIStr
+	IsGlobalStats  bool
 }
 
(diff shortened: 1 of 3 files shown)


Reminiscent (commented):
LGTM

rebelice (approved):

ti-srebot (approved):

PR #1182, opened 2021-03-05
privileges: show create view privilege should in `AllTablePrivs`

xiongjiwei:

please see https://github.com/pingcap/tidb/issues/23127


tangenta:
Do we need to update the related tests in TiDB?

xiongjiwei:
> Do we need to update the related tests in TiDB?

there is no related test in `TiDB`. I will add a test for it when updating the parser

tangenta (commented):
LGTM

ti-srebot (approved):

PR #1189, opened 2021-03-10
set right not null flag when newing extra PK

fzhedu:
fix https://github.com/pingcap/tics/issues/1509, where the extra PK 's not null flag is not right set.

set the not null flag for extran PK

### Check List 

Tests 

 - Unit test
https://github.com/pingcap/tidb/pull/23237

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -140,6 +140,9 @@ func (*testModelSuite) TestModelBasic(c *C) {
 	}
 	no := anIndex.HasPrefixIndex()
 	c.Assert(no, Equals, false)
+
+	extraPK := NewExtraHandleColInfo()
+	c.Assert(extraPK.Flag, Equals, uint(mysql.NotNullFlag | mysql.PriKeyFlag))
 }
 
 func (*testModelSuite) TestJobStartTime(c *C) {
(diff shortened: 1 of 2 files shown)


hanfei1991:
We'd better add a case (even manually) to prove that this change will work and be necessary.

fzhedu:
> We'd better add a case (even manually) to prove that this change will work and be necessary.

a unit test is added in tidb

model/model_test.go
```diff
@@ -140,6 +140,9 @@ func (*testModelSuite) TestModelBasic(c *C) {
 	}
 	no := anIndex.HasPrefixIndex()
 	c.Assert(no, Equals, false)
+
+	extraPK := NewExtraHandleColInfo()
+	c.Assert(extraPK.Flag, Equals, uint(mysql.NotNullFlag | mysql.PriKeyFlag))
```
kennytm, line 145: please fix `gofmt`.

```
gofmt (simplify)
model/model_test.go
make: *** [Makefile:25: fmt] Error 1
```

fzhedu: done

ti-srebot (approved):

solotzg:
@fzhedu Shall we cherry-pick it to release-5.0?

PR #1192, opened 2021-03-12
Added better node position logging

TeemoKill:

Maybe #910 -> How do I get the position of an AST node.
The previous solution logged the starting offset of the original text, 
this pull request added more detailed position information to ast.Node

Moved the struct `Pos` from `parser/lexer.go` into `parser/ast/ast.go`

Added 2 methods  

```go
func (ast.node) SetOriginTextPos(ast.Pos) 
func (ast.node) OriginTextPos() ast.Pos
```  

to the interface `ast.Node`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

N/A


tangenta:
Hi @TeemoKill, would you like to explain why we need this feature? Did you encounter any problems?

PR #851, opened 2020-05-10
Unreserved 8 keywords

kennytm:
### What problem does this PR solve? 

Fix #846.

### What is changed and how it works?

Unreserved the following 8 keywords:

* chain
* error
* general
* nvarchar
* pack_keys
* parser
* shard_row_id_bits
* pre_split_regions

And added a manually executed test case to ensure the list of reserved keywords mostly match that of MySQL. The only discrepancy currently are:

* **current_role** — reserved in TiDB, not reserved in MySQL (introduced by #228, caused significant reduce/reduce conflict when unreserved, cc @imtbkcat)
* **function** — reserved since MySQL 8.0.1, not reserved in TiDB
* **separator** — reserved in MySQL (why?), not reserved in TiDB

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch
    * Affects all TiDB versions.

### Release note 

* `CHAIN`, `ERROR`, `GENERAL`, `NVARCHAR`, `PACK_KEYS`, `PARSER`, `SHARD_ROW_ID_BITS` and `PRE_SPLIT_REGIONS` are no longer treated as reserved words.

PR #1193, opened 2021-03-17
downgrade mysql driver version

lysu:

https://github.com/pingcap/parser/pull/851 introduce go-sql-driver/mysql dependence with high version for test case

but https://github.com/pingcap/mysql-tester will use pingcap/parser

and https://github.com/go-sql-driver/mysql/pull/676 remove Warn API and strict mode which heavy used in pingcap/mysql-tester and pingcap/tidb-test.

(warn can workaround but strict need modify large number for tidb-test testcase)

so this PR choose downgrade mysql/driver version

downgrade veresion

### Check List 

Tests 

 - Manual test 

```
upgrade mysql-tester and run tidb-test success
```

Code changes

 - n/a
 
Side effects

 - n/a
 
Related changes

 - n/a

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/parser/1193)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/cznic/sortutil v0.0.0-20181122101858-f5f958428db8
 	github.com/cznic/strutil v0.0.0-20171016134553-529a34b1c186
 	github.com/cznic/y v0.0.0-20170802143616-045f81c6662a
-	github.com/go-sql-driver/mysql v0.0.0-20170715192408-3955978caca4
+	github.com/go-sql-driver/mysql v1.4.1
 	github.com/pingcap/check v0.0.0-20190102082844-67f458068fc8
 	github.com/pingcap/errors v0.11.5-0.20201029093017-5a7df2af2ac7
 	github.com/pingcap/log v0.0.0-20200511115504-543df19646ad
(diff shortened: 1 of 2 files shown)


kennytm:
is there any plan to fix the tests?

lysu:
> is there any plan to fix the tests?

@bb7133 @zz-jason maybe can schedule a task to upgrade tidb-test's go-sql-driver/mysql to latest version?

kennytm:
@lysu  ok. but could you set the version to v1.4.1 instead of a git commit?

lysu:
> @lysu ok. but could you set the version to v1.4.1 instead of a git commit?

```
git name-rev --tags --name-only a8b7ed4454a6a4f98f85d3ad558cd6d97cec6959
v1.4.0~35
```

v1.4.0 also include https://github.com/go-sql-driver/mysql/pull/676, we can use v1.3.0

lysu:
@kennytm  PTAL, we have tried some way without modify parser but still have no good workaround for this..

kennytm:
@lysu let's use 1.3.0. the dependency is only imported for an optional test.

lysu:
yes, it's using 1.3.0 in the latest commit :)

ti-srebot (approved):

PR #1197, opened 2021-03-24
auth: fix CheckScrambledPassword() panic for invalid input

tiancaiamao:

image

The input data may be invalid, but calling `CheckScrambledPassword()` on malformed data should not make TiDB server panic.

### Check List 

Tests 

 - Unit test

Related changes

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -103,6 +103,9 @@ func CheckScrambledPassword(salt, hpwd, auth []byte) bool {
 	terror.Log(errors.Trace(err))
 	hash := crypt.Sum(nil)
 	// token = scrambleHash XOR stage1Hash
+	if len(auth) != len(hash) {
+		return false
+	}
 	for i := range hash {
 		hash[i] ^= auth[i]
 	}
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -45,6 +45,6 @@ func (s *testAuthSuite) TestCheckScramble(c *C) {
 	c.Assert(res, IsTrue)
 
 	// Do not panic for invalid input.
-	res := CheckScrambledPassword(salt, hpwd, []byte("xxyyzz"))
-	c.Assert(res, False)
+	res = CheckScrambledPassword(salt, hpwd, []byte("xxyyzz"))
+	c.Assert(res, IsFalse)
 }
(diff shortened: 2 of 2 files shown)


tiancaiamao:
The malformed data maybe related to https://github.com/pingcap/tidb/pull/19603, maybe we are not resolving the protocol correctly, but I have no clue.

PTAL @djshow832 @morgo

djshow832 (commented):
Add some test cases?

tiancaiamao:
> Add some test cases?

https://github.com/pingcap/parser/pull/1197/files#diff-b06746870e62f95017714121be6134cffdc4bd1d46dd21f8cae8c484e3e2829dR47-R49   That's all for it.

auth/auth.go
```diff
@@ -103,6 +103,9 @@ func CheckScrambledPassword(salt, hpwd, auth []byte) bool {
 	terror.Log(errors.Trace(err))
 	hash := crypt.Sum(nil)
 	// token = scrambleHash XOR stage1Hash
+	if len(auth) != len(hash) {
```
djshow832, line 106: Is it possible that `len(auth) > len(hash)`?

tiancaiamao: It doesn't matter, as long as `len(auth) != len(hash)` the input is invalid.

ti-srebot (approved):

morgo:
> The malformed data maybe related to [pingcap/tidb#19603](https://github.com/pingcap/tidb/pull/19603), maybe we are not resolving the protocol correctly, but I have no clue.
> 
> PTAL @djshow832 @morgo

I took a look. The `authSwitchRequest` is relatively self contained unless the `resp.AuthPlugin` is `caching_sha2_password` (it doesn't attempt to authSwitch for any other plugins, even though it could). Do you know if it is a client attempting to use MySQL 8.0 auth?

There are other authentication plugins besides  `caching_sha2_password`  OR `mysql_native_password` that clients could attempt to use too, and it's also entirely possible it is a connector that doesn't speak the protocol correctly.

kennytm:
I think we should just get rid of the integration test. The TiDB is always lagging behind making it almost always failing and impossible to use "/merge".

tiancaiamao:
> I took a look. The `authSwitchRequest` is relatively self contained unless the `resp.AuthPlugin` is `caching_sha2_password` (it doesn't attempt to authSwitch for any other plugins, even though it could). Do you know if it is a client attempting to use MySQL 8.0 auth?
> 
> There are other authentication plugins besides `caching_sha2_password` OR `mysql_native_password` that clients could attempt to use too, and it's also entirely possible it is a connector that doesn't speak the protocol correctly.

Yup, I've said I have no clue. I'm not sure how that happen and what trigger the malformed input.
We need more feedback from the user.

PR #1198, opened 2021-03-24
mysql: small privileges refactor

xhebox:

While fixing https://github.com/pingcap/tidb/issues/22703, I noticed two problems:

1. I need to check if a set of privileges has a specific privilege for several times. But we only have raw slices `[]PrivilegeType`, which means I either write a loop or a function on my own - too verbose. The privileges verification package, however, is using bitmask(AllPrivMask) to do that - not consistent.
2. It is very annoying that TiDB `str, ok := Priv2UserCol[v]` followed by three lines of `if !ok { return "invalid priv" }`. We should use the global set, e.g.`AllGlobalPrivs`, to check if it is a valid privilege, then just use `String()` to get a string.

After checking code in `parser/mysql`, I decided to do some small refactors to:

1. Hide direct usage of all maps like `AllXXXPrivs`, or `Priv2UserCol` by invoking methods. Possibly replace maps with switch.
2. Make one bitset type for collections of privileges. Then executor and privileges packages in TiDB could share one type.

The logic will be more self-contained. And TiDB will get a consistent API and a less verbose solution for my fix on tidb#22703. However, the refactor will break the compilation of TiDB.

So this PR will only try to prepare for removing all direct accesses to maps. The refactor should be possible to be done in several PRs without breaking others' PR merging.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes


mysql/privs.go
```diff
@@ -0,0 +1,293 @@
+package mysql
```
kennytm, line 1:
```suggestion
// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.

package mysql
```

kennytm (commented):
Rest LGTM

tangenta:
LGTM

ti-srebot (approved):

PR #1206, opened 2021-04-13
*: add as of clause for `SET` and `SELECT`

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

Ref https://github.com/pingcap/tidb/issues/22765#issuecomment-810906473
### Check List 

Tests 

 - Unit test


parser_test.go
```diff
@@ -5900,3 +5900,26 @@ func (s *testParserSuite) TestCTE(c *C) {
...
+	var sb strings.Builder
+	for _, tt := range table {
+		stmt, _, err := parser.Parse(tt.src, "", "")
+		c.Assert(err, IsNil)
+
+		sel := stmt[0].(*ast.SelectStmt)
+		mode := ast.TimestampReadBoundTimestamp
+		if !tt.ok {
+			mode = ast.TimestampReadExactTimestamp
+		}
```
tangenta, line 5919: The `tt.ok` usually means parse success... You can define your own anonymous struct for the `mode`.

nolouch:
Some details need to discuss https://github.com/pingcap/tidb/issues/22765#issuecomment-820904195

parser_test.go
```diff
@@ -5900,3 +5900,14 @@ func (s *testParserSuite) TestCTE(c *C) {
 
 	s.RunTest(c, table)
 }
+
+func (s *testParserSuite) TestAsOfClause(c *C) {
+	table := []testCase{
+		{"SELECT * FROM `t` AS OF TIMESTAMP READ_TS_IN(DATE_SUB(NOW(), INTERVAL 3 SECOND), NOW());", true, "SELECT * FROM `t` AS OF TIMESTAMP READ_TS_IN(DATE_SUB(NOW(), INTERVAL 3 SECOND), NOW())"},
```
kennytm, line 5906: please add a test case for 

```sql
select * from t as /*comment*/ of timestamp now();
```

kennytm (commented):
is it possible to use a TSO as the input to the AS OF clause?

```sql
select * from a as of timestamp 424430616285806594;
-- or?
select * from a as of tso 424430616285806594;
```

tangenta (dismissed):

nolouch:
@kennytm  Our TSO has some logic bit, not really Unix timestamp. if really have a scene TSO as input we can add supports it in the future.

kennytm:
/lgtm

>  Our TSO has some logic bits, not really Unix timestamp. if really have a scene TSO as input we can add it in the future.

I'd like to note that `@@tidb_snapshot` can accept both the timestamp string and TSO. Also if `READ_TS_IN` calculates the output using resolvedTS it's natural for the function to return a TSO than a DATETIME(6).

nolouch:
If want tso as input, may we also support it in `select * from a as of timestamp 424430616285806594`?

ti-srebot (approved):

parser_test.go
```diff
@@ -5604,6 +5604,7 @@ func (checker *nodeTextCleaner) Enter(in ast.Node) (out ast.Node, skipChildren b
 		}
 	case *ast.FuncCallExpr:
 		node.FnName.O = strings.ToLower(node.FnName.O)
+		node.SetOriginTextPosition(0)
```
JmPotato, line 5607: What does this line work for?

nolouch: It's clear the text in the node, then we can use reflect.DeepEquals to compare two statements in :
https://github.com/pingcap/parser/blob/f416dc9329d0e340a892173f9628f5074d7077ca/parser_test.go#L388-L391

PR #1208, opened 2021-04-16
Issue 23494

Howie59:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -215,8 +215,11 @@ type Job struct {
 	SchemaState SchemaState     `json:"schema_state"`
 	// SnapshotVer means snapshot version for this job.
 	SnapshotVer uint64 `json:"snapshot_ver"`
-	// StartTS uses timestamp allocated by TSO.
+	// CreateTS uses timestamp allocated by TSO.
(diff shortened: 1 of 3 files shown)


model/ddl.go
```diff
@@ -215,8 +215,11 @@ type Job struct {
 	SchemaState SchemaState     `json:"schema_state"`
 	// SnapshotVer means snapshot version for this job.
 	SnapshotVer uint64 `json:"snapshot_ver"`
-	// StartTS uses timestamp allocated by TSO.
+	// CreateTS uses timestamp allocated by TSO.
 	// Now it's the TS when we put the job to TiKV queue.
+	CreateTS uint64 `json:"create_ts"`
```
zimulala, line 220: I think we'd better don't change the `StartTS` in the `Job` structure that may cause incompatibility between the old and new versions. We can change it when executing showing DDL jobs.

Howie59: this means we needn't add the variable`CreateTS`?

PR #1211, opened 2021-04-21
parser: support the 'create global temporary table' syntax

tiancaiamao:

Support `create global temporary table ... on commit delete rows`
See https://github.com/pingcap/tidb/pull/24168

This syntax is compatible with MySQL temporary table, a `GLOBAL` keyword is used to distinguish them.
This syntax is much the same with Oracle (and also  the ANSI SQL standard).

A compatibility issue is that when a user use this syntax and use a downstream MySQL as TiDB backup, they will encounter DDL error when the use this statement, because this syntax is a superset of the MySQL syntax.

In the real use cases, it shouldn't be a problem, because if a user use this syntax, they probably migrating from Oracle, so they do not use a MySQL backup.

### Check List 

Tests 

 - Unit test

Code changes

Related changes

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -335,6 +335,27 @@ type TableInfo struct {
 	// IsColumnar means the table is column-oriented.
 	// It's true when the engine of the table is TiFlash only.
 	IsColumnar bool `json:"is_columnar"`
+
+	TempTableType `json:"temp_table_type"`
+}
+
+type TempTableType byte
+
+const (
+	TempTableNone TempTableType = iota
+	TempTableGlobal
+	TempTableLocal
+)
+
+func (t TempTableType) String() string {
+	switch t {
+	case TempTableGlobal:
+		return "global"
+	case TempTableLocal:
+		return "local"
+	default:
+		return ""
+	}
 }
 
 // TableLockInfo provides meta data describing a table lock.
--- a/parser.y
+++ b/parser.y
@@ -495,6 +495,7 @@ import (
 	preSplitRegions       "PRE_SPLIT_REGIONS"
 	preceding             "PRECEDING"
 	prepare               "PREPARE"
+	preserve              "PRESERVE"
 	privileges            "PRIVILEGES"
 	process               "PROCESS"
 	processlist           "PROCESSLIST"
@@ -1036,6 +1037,7 @@ import (
 	NoWriteToBinLogAliasOpt                "NO_WRITE_TO_BINLOG alias LOCAL or empty"
 	ObjectType                             "Grant statement object type"
 	OnDuplicateKeyUpdate                   "ON DUPLICATE KEY UPDATE value list"
+	OnCommitOpt                            "ON COMMIT DELETE |PRESERVE ROWS"
 	DuplicateOpt                           "[IGNORE|REPLACE] in CREATE TABLE ... SELECT statement or LOAD DATA statement"
 	OptErrors                              "ERRORS or empty"
 	OptFull                                "Full or empty"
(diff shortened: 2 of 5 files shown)


parser.y
```diff
@@ -3634,12 +3634,20 @@ CreateTableStmt:
...
-			ReferTable:  $6.(*ast.TableName),
-			IfNotExists: $4.(bool),
-			IsTemporary: $2.(bool),
+			Table:            $5.(*ast.TableName),
+			ReferTable:       $6.(*ast.TableName),
+			IfNotExists:      $4.(bool),
+			TemporaryKeyword: $2.(ast.TemporaryKeyword),
 		}
 	}
+|	"CREATE" "GLOBAL" "TEMPORARY" "TABLE" IfNotExists TableName TableElementListOpt "ON" "COMMIT" "DELETE" "ROWS"
```
djshow832, line 3643: Are you going to support `create global temporary table ... like / as` and `TableOption` in another PR?
I think `TableOption` is always a must.

tiancaiamao:
Integration test is fixed in https://github.com/pingcap/tidb/pull/24196

parser.y
kennytm: yeah why not just replace `OptTemporary` by a rule accepting both `TEMPORARY` and `GLOBAL TEMPORARY` (and &epsilon;).

tiancaiamao: If we make the `OptTemporary` rule accept `GLOBAL TEMPORARY` and reuse the old rules above, we have to disable a lot of things on the TiDB part, it's risky to introduce bugs if we miss something.

So instead of supporting the grammar fully and disable features, I prefer to make a small grammar first and add then features.
@kennytm @djshow832

model/model.go
```diff
@@ -335,6 +335,9 @@ type TableInfo struct {
 	// IsColumnar means the table is column-oriented.
 	// It's true when the engine of the table is TiFlash only.
 	IsColumnar bool `json:"is_columnar"`
+
+	// IsTemporary is true when this table is a global temporary table.
+	IsTemporary bool `json:"is_temporary"`
```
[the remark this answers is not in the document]
tiancaiamao: Done

parser.y
```diff
@@ -4256,17 +4264,17 @@ DropIndexStmt:
 DropTableStmt:
 	"DROP" OptTemporary TableOrTables IfExists TableNameList RestrictOrCascadeOpt
 	{
-		$$ = &ast.DropTableStmt{IfExists: $4.(bool), Tables: $5.([]*ast.TableName), IsView: false, IsTemporary: $2.(bool)}
+		$$ = &ast.DropTableStmt{IfExists: $4.(bool), Tables: $5.([]*ast.TableName), IsView: false, TemporaryKeyword: $2.(ast.TemporaryKeyword)}
```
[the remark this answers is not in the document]
tiancaiamao: Global temporary table use the normal `drop` statement without  the `temporary` keyword.

parser.y
djshow832: We need to at least support charset, collate, and storage engine before GA. So we need to write duplicate parser code here in the near future.

djshow832: Actually, most of the table options can be ignored ( some of them are already ignored by normal tables ). I think it's fine to handle the other options ( by someone else ).

tiancaiamao:
I've change `OptTemporary` to include `GLOBAL TEMPORARY` now, PTAL @djshow832 @kennytm

parser.y
```diff
@@ -3621,28 +3622,44 @@ DatabaseOptionList:
...
-		stmt.IsTemporary = $2.(bool)
+		stmt.TemporaryKeyword = $2.(ast.TemporaryKeyword)
 		stmt.Options = $7.([]*ast.TableOption)
 		if $8 != nil {
 			stmt.Partition = $8.(*ast.PartitionOptions)
 		}
 		stmt.OnDuplicate = $9.(ast.OnDuplicateKeyHandlingType)
 		stmt.Select = $11.(*ast.CreateTableStmt).Select
+		if ($12 != nil && stmt.TemporaryKeyword != ast.TemporaryGlobal) || (stmt.TemporaryKeyword == ast.TemporaryGlobal && $12 == nil) {
+			yylex.AppendError(yylex.Errorf("GLOBAL TEMPORARY and ON COMMIT DELETE ROWS must appear together"))
```
djshow832, line 3638: Add a test case for this?

model/model.go
```diff
@@ -335,6 +335,27 @@ type TableInfo struct {
...
+)
+
+func (t TempTableType) String() string {
+	switch t {
+	case TempTableGlobal:
+		return "global"
+	case TempTableLocal:
+		return "local"
+	default:
+		return "none"
```
coocood, line 357: Why not empty string?

tiancaiamao: Done

ti-srebot (approved):

tiancaiamao:
The license/cal check fails and I try to modify the git history of my commit email, `git push -f` is used

PTAL @coocood @djshow832

coocood (approved):

PR #1214, opened 2021-04-22
ast: visit Join.Using in Join.Accept

PragmaTwice:
`ast.Join.Accept` should visit its field `Using`

A for loop is added to `ast.Join.Accept` for visiting all nodes in the slice `Using`

### Check List 

Tests 

`make test` passes in repo `pingcap/tidb` while replace `pingcap/parser` to `pragmatwice/parser@patch-ast-dml-join-accept`

```
...
Great, all tests passed.
```

Code changes

Side effects

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -246,6 +246,13 @@ func (n *Join) Accept(v Visitor) (Node, bool) {
 		}
 		n.On = node.(*OnCondition)
 	}
+	for i, col := range n.Using {
+		node, ok = col.Accept(v)
+		if !ok {
+			return n, false
+		}
+		n.Using[i] = node.(*ColumnName)
+	}
 	return v.Leave(n)
 }
 


tisonkun:
@PragmaTwice Is there any issue related to this change or could you add some tests for how it fixes?

PragmaTwice:
> @PragmaTwice Is there any issue related to this change or could you add some tests for how it fixes?

No issue now, but I want to iterate all nodes in the AST via `Node.Accept(Visitor)`, and `Join.Using` is missing.

I think the original purpose of `Node.Accept(Visitor)` is to visit its all children node.

tangenta (approved):

ti-srebot (approved):

tisonkun:
> > @PragmaTwice Is there any issue related to this change or could you add some tests for how it fixes?
> 
> @tisonkun No issue now, but I want to iterate all nodes in the AST via `Node.Accept(Visitor)`, and `Join.Using` is missing.
> 
> I think the original purpose of `Node.Accept(Visitor)` is to visit its all children nodes.

@PragmaTwice thanks for your reply. I'm curious it doesn't cause some issues before..

PR #1217, opened 2021-04-28
fix CTE parser

### LINKED ISSUE #1218: Some CTE statement is parse wrongly
## Bug Report

1. What did you do?
``` sql
with cte as (select 1) select 1 union with cte as (select 1) select * from cte;
with cte as (select 1) (select 1);
with cte as (select 1) (select 1 union select 1);
select * from (with cte as (select 1) select 1 union select 2) qn;
select * from t where 1 > (with cte as (select 2) select * from cte);
( with cte(n) as ( select 1 ) select n+1 from cte union select n+2 from cte) union select 1;
( with cte(n) as ( select 1 ) select n+1 from cte) union select 1;
( with cte(n) as ( select 1 ) (select n+1 from cte)) union select 1;
```

The same as MySQL

3. What did you see instead?
Inconsistent with MySQL

master

wjhuang2016:

Fix #1218
The with clause should be set before setOprStmt, not selectStmt. 
For example:
with cte as (select 1) select 1 union select 2;

The with clause should be added on the SetOprStmt, not the first selectStmt.
See the added tests.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -1061,6 +1061,8 @@ type SelectStmt struct {
 	TableHints []*TableOptimizerHint
 	// IsInBraces indicates whether it's a stmt in brace.
 	IsInBraces       bool
+	// WithBeforeBraces indicates whether stmt's with clause is before the brace.
+	// It's used to distinguish (with xxx select xxx) and with xxx (select xxx)
 	WithBeforeBraces bool
 	// QueryBlockOffset indicates the order of this SelectStmt if counted from left to right in the sql text.
 	QueryBlockOffset int
--- a/ast/dml.go
+++ b/ast/dml.go
@@ -1404,12 +1404,18 @@ func (n *SelectStmt) Accept(v Visitor) (Node, bool) {
 type SetOprSelectList struct {
 	node
 
+	With             *WithClause
 	AfterSetOperator *SetOprType
 	Selects          []Node
 }
 
 // Restore implements Node interface.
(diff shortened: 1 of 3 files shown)


ast/dml.go
```diff
@@ -1060,7 +1060,8 @@ type SelectStmt struct {
 	// TableHints represents the table level Optimizer Hint for join type
 	TableHints []*TableOptimizerHint
 	// IsInBraces indicates whether it's a stmt in brace.
-	IsInBraces bool
+	IsInBraces       bool
+	WithBeforeBraces bool
```
xiongjiwei, line 1064: it will be useful if add a comment for `WithBeforeBraces`

xiongjiwei (approved):

ast/dml.go
tangenta: Do you mean parentheses `()`? I remember brace means `{}`.

kennytm: IsInBraces here refer to the `()` too (why do we need to care if `()` exists anyway). Legacy typo 🤷

wjhuang2016: It's used to distinguish (with xxx select xxx) and with xxx (select xxx)

xiongjiwei (approved):

tangenta (approved):

ti-srebot (approved):

PR #1221, opened 2021-05-08
parser, ast: make parser happy for PARTITION BY [LINEAR] KEY ALGORITHM syntax

spongedu:
 
For Now, in parser.y, the following rule: 
https://github.com/pingcap/parser/blob/a8654fa287fd2bf0f28530737d62b6fb26488078/parser.y#L3725
doesn't actually work, because  lex translate `'='` into token `eq`, which is not equal to `int('=')`. See:
https://github.com/pingcap/parser/blob/a8654fa287fd2bf0f28530737d62b6fb26488078/misc.go#L103

1. use `"="` or `eq` instead
2. tiny modification to make `Restore` happy

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

   N/A

Related changes

### DIFF
--- a/ast/ddl.go
+++ b/ast/ddl.go
@@ -933,29 +933,44 @@ func (n *ColumnDef) Validate() bool {
 	return !(generatedCol && illegalOpt4gc)
 }
 
+type TemporaryKeyword int
+
+const (
+	TemporaryNone TemporaryKeyword = iota
+	TemporaryGlobal
+	TemporaryLocal
+)
+
 // CreateTableStmt is a statement to create a table.
 // See https://dev.mysql.com/doc/refman/5.7/en/create-table.html
 type CreateTableStmt struct {
 	ddlNode
 
 	IfNotExists bool
-	IsTemporary bool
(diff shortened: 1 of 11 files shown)


ast/ddl.go
```diff
@@ -3299,6 +3308,10 @@ func (n *PartitionMethod) Restore(ctx *format.RestoreCtx) error {
 	}
 	ctx.WriteKeyWord(n.Tp.String())
 
+	if n.KeyAlgorithm != nil {
+		ctx.WritePlain(fmt.Sprintf(" ALGORITHM = %d", n.KeyAlgorithm.Type))
```
kennytm, line 3312:
```suggestion
		ctx.WriteKeyWord(" ALGORITHM")
		ctx.WritePlainf(" = %d", n.KeyAlgorithm.Type)
```

xiongjiwei (dismissed):
/lgtm

ti-srebot (dismissed):

kennytm (approved):

PR #1225, opened 2021-05-16
Fix Format syntax err message when value out of range

chenquan:
related PR: pingcap/tidb#17489 pingcap/tidb#17745

Ignore error that the decimal  out  of range the value 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/yy_parser.go
+++ b/yy_parser.go
@@ -158,7 +158,24 @@ func (parser *Parser) Parse(sql, charset, collation string) (stmt []ast.StmtNode
 		warns = nil
 	}
 	if len(errs) != 0 {
-		return nil, warns, errors.Trace(errs[0])
+		err := errs[0]
+		var isErr = true
+		if len(parser.result) != 0 {
+			node := parser.result[0]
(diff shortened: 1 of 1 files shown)


chenquan:
Why are all integration tests timed out?
https://app.circleci.com/pipelines/github/pingcap/parser
image

chenquan:
PTAL @tangenta @kennytm @JmPotato

yy_parser.go
```diff
@@ -158,7 +158,24 @@ func (parser *Parser) Parse(sql, charset, collation string) (stmt []ast.StmtNode
 		warns = nil
 	}
 	if len(errs) != 0 {
-		return nil, warns, errors.Trace(errs[0])
+		err := errs[0]
```
xiongjiwei, line 161: It changes the logistic, can you please explain why do not return the error in some situations?

xiongjiwei:
BTW, can you please add some tests for it?

yy_parser.go
chenquan: Maybe it's not the best solution, I'll close it.

PR #1227, opened 2021-05-18
Fix running of auth tests

dveeden:
### What problem does this PR solve? 

The tests in `auth/auth_test.go` were not run.

To see the issue modify `auth/auth_test.go` to make a test case that should fail. Then run `make test`. The expected outcome is a failed test run, but in fact it doesn't show any errors.

### What is changed and how it works?

- This adds `check.TestingT()` to `auth/auth_test.go`
- This adds a simple check in `test.sh` to test for similar cases.

### Check List 

Tests 

 - Unit test

PR #1232, opened 2021-05-24
Caching sha2

dveeden:

For improved compatibility with MySQL 8.0 TiDB needs to support the `caching_sha2` authentication method.

This implements the validation of passwords for this authentication format.

Related:
- https://github.com/pingcap/tidb/issues/9411
- https://github.com/pingcap/parser/pull/1227 (this PR contains a commit for that PR)

- Split out `mysql_native_password` related code from `auth/auth.go`
- Added `caching_sha2` code.

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity


auth/caching_sha2.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// Checks if a MySQL style caching_sha2 authentication string matches a password
+func CheckShaPassword(pwhash []byte, password string) (bool, error) {
+	pwhash_parts := bytes.Split(pwhash, []byte("$"))
+	if len(pwhash_parts) != 4 {
+		return false, errors.New("failed to decode hash parts")
+	}
+
+	hash_type := string(pwhash_parts[1])
```
tiancaiamao, line 186: Check `pwhash_parts` size first?
If my memory is correct, TiDB does not do `Auth()` in a separate goroutine, that means a panic in `pwhash_parts[2]` make the process exit. See https://github.com/pingcap/parser/pull/1197

tiancaiamao: Or `defer recover` here if verify the input in advance is hard? https://blog.golang.org/defer-panic-and-recover

auth/caching_sha2_test.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2015 PingCAP, Inc.
```
tiancaiamao, line 1: ditto

auth/caching_sha2_test.go
```diff
@@ -0,0 +1,69 @@
...
+	_, err := CheckShaPassword(pwhash, pwd)
+	c.Assert(err, NotNil)
+}
+
+// The output from NewSha2Password is not stable as the hash is based on the genrated salt.
+// This is why CheckShaPassword is used here.
+func (s *testAuthSuite) TestNewSha2Password(c *C) {
+	pwd := "testpwd"
+	pwhash := NewSha2Password(pwd)
+	fmt.Printf("pwhash = %-v", pwhash)
```
tiancaiamao, line 65: Remove the debug log here?

auth/caching_sha2.go
```diff
@@ -0,0 +1,208 @@
...
+		n--
+		ret = append(ret, b64t[w&0x3f])
+		w >>= 6
+	}
+
+	return ret
+}
+
+func sha256crypt(plaintext string, salt []byte, iterations int) string {
+	// 1, 2, 3
```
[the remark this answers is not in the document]
dveeden: These refer to https://www.akkadia.org/drepper/SHA-crypt.txt 
I'll add a note about this on the top of the function

auth/caching_sha2.go
```diff
@@ -0,0 +1,208 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
```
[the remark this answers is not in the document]
dveeden: Moving this to pingcap/tidb might make sense, but that would increase the scope of this quite a bit. And doing this while adding `caching_sha2_password` is probably not smart. @morgo what is your take on this?

auth/caching_sha2.go
dveeden: Isn't that done on line 184 ?

dveeden:
I think we should clone this PR and continue with #1236 which includes the work from this PR

auth/caching_sha2.go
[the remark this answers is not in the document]
tiancaiamao: Okay

ti-srebot (approved):

tiancaiamao:
I'd rather merge this on first, 400 lines changes is already quite large.
Generally, a single PR more than 300 lines makes a hard time for the reviewer...
You can rebase master and update https://github.com/pingcap/parser/pull/1236 then?

PR #1237, opened 2021-06-02
testing the integration test

tiancaiamao:
For testing ...

PR #1238, opened 2021-06-02
CI: Use Go 1.16 instead of Go 1.13

dveeden:

The Integration Test output currently has this:
```
note: module requires Go 1.16
```

This changes the container image used by CI

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build-ut:
     docker:
-      - image: golang:1.13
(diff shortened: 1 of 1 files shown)


dveeden:
Looks like @tiancaiamao is already working on this in #1237

tiancaiamao:
Nope, I'm not updating the Go version, I just testing the integration test.

You can reopen this one.

tiancaiamao:
But I have not yet figure out why the integration test does not work and was interrupted then~

PR #1240, opened 2021-06-02
fix restore for count(*)

XuHuaiyu:
Before this commit, `select count(*) from t ` will be restored to `select count(1) from t`.
After this commit, `select count(*) from t ` can be restored to `select count(*) from t` correctly

### Check List 

Tests 

 - Unit test

Code changes

Side effects

N/A

Related changes


ast/functions_test.go
```diff
@@ -128,7 +128,7 @@ func (ts *testFunctionsSuite) TestAggregateFuncExprRestore(c *C) {
 		{"BIT_OR(test_score)", "BIT_OR(`test_score`)"},
 		{"BIT_XOR(test_score)", "BIT_XOR(`test_score`)"},
 		{"COUNT(test_score)", "COUNT(`test_score`)"},
-		{"COUNT(*)", "COUNT(1)"},
+		{"COUNT(*)", "COUNT(*)"},
```
xiongjiwei, line 131: what if 
```sql
count(a)
count(`a*`)
```

PR #1242, opened 2021-06-08
auth: Fix salt generation of caching_sha2_password

dveeden:

This code intended to set the first bit of every byte of the salt
to 0 as the salt should not contain multi-byte UTF-8.

However it failed to do this. This changes the `^` with a `^&` to fix
that.

In addition to that this also avoids the '$' inside salts as that is
already used as delimiter in the authentication string.

### Check List 

Tests 

 - Unit test

This is related to https://github.com/pingcap/tidb/pull/24991

### DIFF
--- a/auth/caching_sha2.go
+++ b/auth/caching_sha2.go
@@ -205,8 +205,15 @@ func CheckShaPassword(pwhash []byte, password string) (bool, error) {
 func NewSha2Password(pwd string) string {
 	salt := make([]byte, SALT_LENGTH)
 	rand.Read(salt)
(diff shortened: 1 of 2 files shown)


dveeden:
https://github.com/mysql/mysql-server/blob/98b2ccb470de120d36bc4a623c814cdfded958ec/mysys/crypt_genhash_impl.cc#L425-L436 is how MySQL does this.

ti-srebot (approved):

tiancaiamao (approved):

PR #1262, opened 2021-07-04
parser: fix the compatiblity of explain_format

hawkingrei:

fix the compatiblity of explain_format [tidb/issues/25913](https://github.com/pingcap/tidb/issues/25913)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes
add three NotKeywordToken: VERBOSE BRIEF DOT

Related changes


mysql/util.go
```diff
@@ -93,3 +97,11 @@ func GetDefaultFieldLengthAndDecimalForCast(tp byte) (flen int, decimal int) {
 	}
 	return -1, -1
 }
+
+// GetDefaultExplainFormat return the processed values for format of explain
+func GetDefaultExplainFormat(f string) string {
```
xhebox, line 102: Isn't `row` is returned for `TRADITIONAL`?

https://github.com/pingcap/parser/blob/f88b9d7642986fb6f5c4f327bb346b6b2fb1aa63/parser.y#L4463-L4466

I don't get it. Maybe we could replace `stringLit` into `ExplainFormatType`? May you investigate why it is `stringLit` rather than `ExplainFormatType`?

xhebox (commented):
Rest LGTM

mysql/util.go
hawkingrei: In fact, the type of parts is different. the first is Identifier. The socend is a string. so we can support diffrent types in expalain_format. In goyacc, I don't find a function that it can get value from different type. but in MySQL, ``` ident_or_text``` can do it.

hawkingrei: it is possible for it to improve, But I think it is a appropriate way to create a new PR for it to improve.

xhebox (dismissed):

parser.y
```diff
@@ -4469,6 +4474,22 @@ ExplainFormatType:
 	{
 		$$ = "json"
 	}
+|	"ROW"
```
tiancaiamao, line 4477: `ROW` is reserved keywords since 8.0.2, shall we follow it?

mysql/util.go
tiancaiamao: What's the purpose exporting this function, is it used anywhere else?

mysql/util.go
```diff
@@ -93,3 +97,11 @@ func GetDefaultFieldLengthAndDecimalForCast(tp byte) (flen int, decimal int) {
...
+	if strings.Compare(strings.ToUpper(f), "TRADITIONAL") == 0 {
+		return "row"
+	}
+	return f
```
tangenta, line 106: I think this conversion should be done in TiDB.

tiancaiamao: Agree...

How about the `Restore()` of that statement?

hawkingrei (commented):
removed this function

mysql/util.go
hawkingrei: it has removed.

parser.y
hawkingrei: it has been a reserved keyword. i add a test for it.

parser.y
```diff
@@ -4463,12 +4466,28 @@ ExplainStmt:
 ExplainFormatType:
 	"TRADITIONAL"
 	{
-		$$ = "row"
+		$$ = "traditional"
```
tangenta, line 4469: Does it need to be lowercase? If no, we can simply remove `$$ = "xxxx"`.

hawkingrei: Great, I remote it

tangenta (approved):

ti-srebot (approved):

parser.y
tiancaiamao: In MySQL8:

```
mysql> create table row (ts varchar(32));
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'row (ts varchar(32))' at line 1

mysql> create table `row` (ts varchar(32));
Query OK, 0 rows affected (0.02 sec)
```

In 5.7 it is not a reserved keyword, so the behaviour is different.

tiancaiamao:
Is there a corresponding TiDB PR for this? @hawkingrei

hawkingrei:
> Is there a corresponding TiDB PR for this? @hawkingrei

No, i will create a new pr to fix this problem in tidb after merge this pr.

tiancaiamao (approved):

PR #1271, opened 2021-07-08
 Support add/drop attributes for table or partition

rleungx:

It's used for the region label feature. Closes https://github.com/pingcap/parser/issues/1288.

This PR is going to support
```
ALTER TABLE t ATTRIBUTES="attr1,attr2,...,attrn"
ALTER TABLE t PARTITION p ATTRIBUTES="attr1,attr2,...,attrn"
```
to change the attributes of a table or a partition.

And use the following syntax to reset all attributes for a table or a partition:
```
ALTER TABLE t ATTRIBUTES=default
ALTER TABLE t PARTITION p ATTRIBUTES=default
```

### Check List 

Tests 

 - Unit test

Related changes


kennytm:
what is the difference between this and PLACEMENT POLICY?

ast/ddl.go
```diff
@@ -2648,6 +2651,19 @@ func (n *AlterTableSpec) Restore(ctx *format.RestoreCtx) error {
 				return errors.Annotatef(err, "An error occurred while restore AlterTableSpec.PlacementSpecs[%d]", i)
 			}
 		}
+	case AlterTableAlterPartitionAttributes:
+		if len(n.PartitionNames) != 1 {
+			return errors.Errorf("Maybe partition options are combined.")
```
JmPotato, line 2656: This error message indicates the number of partion names is over 1, what about zero?

rleungx:
> what is the difference between this and PLACEMENT POLICY?

We are going to add some attributes to a table/partition to solve some specified problems. e.g., `notmerge` attribute means don't merge a table. It's different from the placement rule which mainly focuses on data placement.

tangenta:
@rleungx We may also need to consider the removal syntax for attributes. Anyway, it would be better if there is a detailed design document.

rleungx:
PTAL @djshow832 @xhebox, thanks.

rleungx:
> @rleungx We may also need to consider the removal syntax for attributes. Anyway, it would be better if there is a detailed design document.

How about using `alter table t attributes=default` for removing all attributes?

parser.y
```diff
@@ -1586,9 +1587,13 @@ AlterTablePartitionOpt:
 	PartitionOpt
 	{
 		if $1 != nil {
-			$$ = &ast.AlterTableSpec{
-				Tp:        ast.AlterTablePartition,
-				Partition: $1.(*ast.PartitionOptions),
+			if alterTableSpec, ok := $1.(*ast.AlterTableSpec); ok {
```
xhebox, line 1590: I suggest adding a new term `ATTRIBUTES : "ATTRIBUTES" "=" stringLit | "DEFAULT"`, which will return the attributes spec. And the change here is not elegant for me. It is strange if `PartitionOpt` did not return `PartitionOptions`.

xhebox (approved):

nolouch (approved):
lgtm

parser_test.go
```diff
@@ -2783,6 +2783,20 @@ func (s *testParserSuite) TestDDL(c *C) {
 		{"ALTER TABLE t ALTER PLACEMENT POLICY CONSTRAINTS='str' ROLE=leader REPLICAS=1", true, "ALTER TABLE `t` ALTER PLACEMENT POLICY CONSTRAINTS='str' ROLE=LEADER REPLICAS=1"},
 		{"ALTER TABLE t ADD PLACEMENT POLICY CONSTRAINTS='str1' ROLE=leader REPLICAS=1, ADD PLACEMENT POLICY CONSTRAINTS='str2' ROLE=leader REPLICAS=1", true, "ALTER TABLE `t` ADD PLACEMENT POLICY CONSTRAINTS='str1' ROLE=LEADER REPLICAS=1, ADD PLACEMENT POLICY CONSTRAINTS='str2' ROLE=LEADER REPLICAS=1"},
 
+		// alter attributes
+		{"ALTER TABLE t ATTRIBUTES='str'", true, "ALTER TABLE `t` ATTRIBUTES='str'"},
+		{"ALTER TABLE t ATTRIBUTES='str1,str2'", true, "ALTER TABLE `t` ATTRIBUTES='str1,str2'"},
+		{"ALTER TABLE t ATTRIBUTES=DEFAULT", true, "ALTER TABLE `t` ATTRIBUTES=DEFAULT"},
+		{"ALTER TABLE t ATTRIBUTES=default", true, "ALTER TABLE `t` ATTRIBUTES=DEFAULT"},
+		{"ALTER TABLE t ATTRIBUTES=DeFaUlT", true, "ALTER TABLE `t` ATTRIBUTES=DEFAULT"},
```
nolouch, line 2791: BTW, Default meaning empty or may have some default attributes？

rleungx: It means empty.

parser.y
```diff
@@ -1586,6 +1588,16 @@ PlacementSpecList:
...
+AttributesOpt:
+	"ATTRIBUTES" "=" "DEFAULT"
+	{
+		$$ = &ast.AttributesSpec{Default: true}
+	}
+|	"ATTRIBUTES" "=" stringLit
+	{
+		$$ = &ast.AttributesSpec{Default: false, Attributes: $3}
+	}
+
```
djshow832, line 1600: As I recall, `=` is optional in all table options in MySQL.

parser_test.go
```diff
@@ -2783,6 +2783,20 @@ func (s *testParserSuite) TestDDL(c *C) {
 		{"ALTER TABLE t ALTER PLACEMENT POLICY CONSTRAINTS='str' ROLE=leader REPLICAS=1", true, "ALTER TABLE `t` ALTER PLACEMENT POLICY CONSTRAINTS='str' ROLE=LEADER REPLICAS=1"},
 		{"ALTER TABLE t ADD PLACEMENT POLICY CONSTRAINTS='str1' ROLE=leader REPLICAS=1, ADD PLACEMENT POLICY CONSTRAINTS='str2' ROLE=leader REPLICAS=1", true, "ALTER TABLE `t` ADD PLACEMENT POLICY CONSTRAINTS='str1' ROLE=LEADER REPLICAS=1, ADD PLACEMENT POLICY CONSTRAINTS='str2' ROLE=LEADER REPLICAS=1"},
 
+		// alter attributes
+		{"ALTER TABLE t ATTRIBUTES='str'", true, "ALTER TABLE `t` ATTRIBUTES='str'"},
+		{"ALTER TABLE t ATTRIBUTES='str1,str2'", true, "ALTER TABLE `t` ATTRIBUTES='str1,str2'"},
```
djshow832, line 2788: How about double quotation marks, like "str1, str2"?

djshow832 (commented):
Rest LGTM

djshow832 (approved):

PR #1272, opened 2021-07-11
Model: add the new field to the Job Struct.

ou-bing:
Support to delete multiple indexes in one statement.

Add the new field to the Job Struct, I want to save some warnings when performing the change multi-schema operation.


ou-bing:
@zimulala hello,  I want to save some warnings when performing the change multi-schema operation. Do you think that's ok?

PR #1274, opened 2021-07-14
parser: support SQL bind capture filter syntax

eurekaka:

Parser changes for https://github.com/pingcap/tidb/issues/15655

Add new syntax for baseline capture filter.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

N/A

Related changes

### DIFF
--- a/ast/misc.go
+++ b/ast/misc.go
@@ -1737,6 +1737,9 @@ const (
 	AdminShowTelemetry
 	AdminResetTelemetryID
 	AdminReloadStatistics
+	AdminCaptureFilterTable
+	AdminCaptureFilterDB
+	AdminShowCaptureFilter
 )
 
 // HandleRange represents a range where handle value >= Begin and < End.
@@ -1806,9 +1809,11 @@ type AdminStmt struct {
 
 	Tp        AdminStmtType
 	Index     string
+	DBName    string
 	Tables    []*TableName
 	JobIDs    []int64
 	JobNumber int64
+	Ignore    bool
 
 	HandleRanges []HandleRange
 	ShowSlow     *ShowSlow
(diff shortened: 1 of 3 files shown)


tangenta:
`NOT IGNORE` sounds counter-intuitive... Can we use `ADD`/`REMOVE` instead?

```
admin capture bindings remove table t1;
admin capture bindings remove table test.t1;
admin capture bindings add table t1;
admin capture bindings add table test.t1;
admin capture bindings remove database db1;
admin capture bindings add database db1;
admin capture bindings show;
```

eurekaka:
> `NOT IGNORE` sounds counter-intuitive... Can we use `ADD`/`REMOVE` instead?
> 
> ```
> admin capture bindings remove table t1;
> admin capture bindings remove table test.t1;
> admin capture bindings add table t1;
> admin capture bindings add table test.t1;
> admin capture bindings remove database db1;
> admin capture bindings add database db1;
> admin capture bindings show;
> ```

The capture filter is blacklist, so we need a negative keyword. The `not ignore` is to remove the item from the blacklist. If we use `add` / `remove`, users may think that only the added items are considered for baseline capture.

tangenta:
Ok. Another question: is it possible to use a system table to represent the blacklist(like `mysql.expr_pushdown_blacklist`)? One of the advantages is that we can use `INSERT`/`DELETE`/`SELECT` to show the list.

eurekaka:
> Ok. Another question: is it possible to use a system table to represent the blacklist(like `mysql.expr_pushdown_blacklist`)? One of the advantages is that we can use `INSERT`/`DELETE`/`SELECT` to show the list, which is easier to learn.

Alright.

PR #1278, opened 2021-07-19
parser: add plan recreator statement

rebelice:

This is a step in [tidb/issues/26325](https://github.com/pingcap/tidb/issues/26325)

Add `PLAN RECREATOR` statement

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


parser.y
```diff
@@ -13158,4 +13166,102 @@ RowStmt:
...
+		if $8 != nil {
+			x.OrderBy = $8.(*ast.OrderByClause)
+		}
+		if $9 != nil {
+			x.Limit = $9.(*ast.Limit)
+		}
+
+		$$ = x
+	}
+|	"PLAN" "RECREATOR" "DUMP" "EXPLAIN" "ANALYZE" "SLOW" "QUERY" WhereClauseOptional OrderByOptional SelectStmtLimitOpt
```
xhebox, line 13233: Do you think `AnalyzeOptional` will better? Not a necessary change request.

xhebox (commented):
Rest LGTM. BTW, the document is internal, and there is no corresponding RFC on github.

parser.y
rebelice: Agree with you. I'll modify it later.

rebelice:
> Rest LGTM. BTW, the document is internal, and there is no corresponding RFC on github.

Here is a description of this feature in https://github.com/pingcap/tidb/issues/26325. This is the new feature for Sprint 4, I'll add documents visible to everyone before this feature GA. Is this OK for you?

xhebox:
@rebelice Sure.

parser.y
rebelice: When I added `AnalyzeOptional`, I found that this abstraction seemed not necessary but added more work. I currently prefer not to modify this.

ast/misc.go
```diff
@@ -343,6 +343,11 @@ func (n *PlanRecreatorStmt) Accept(v Visitor) (Node, bool) {
 		return n, false
 	}
 	n.Stmt = node.(StmtNode)
+	stmt, ok := n.Stmt.Accept(v)
```
xhebox, line 346: Accept twice?

rebelice: Sorry about that. Fixed.

xhebox (approved):

parser.y
```diff
@@ -5582,6 +5586,7 @@ UnReservedKeyword:
 |	"DAY"
 |	"DEALLOCATE"
 |	"DO"
+|	"DUMP"
```
tiancaiamao, line 5589: It's not in the UnReservedKeyword list:

It should be NotKeyword?

parser.y
```diff
@@ -5624,6 +5629,7 @@ UnReservedKeyword:
 |	"PROXY"
 |	"QUICK"
 |	"REBUILD"
+|	"RECREATOR"
```
tiancaiamao, line 5632: ditto

parser.y
```diff
@@ -5719,6 +5725,7 @@ UnReservedKeyword:
 |	"PROFILES"
 |	"MICROSECOND"
 |	"MINUTE"
+|	"PLAN"
```
tiancaiamao, line 5728: ditto ... plan is also not a keyword

parser.y
rebelice: updated all

tiancaiamao (approved):

PR #1283, opened 2021-07-21
parser: support Oracle translate function

wzru:
https://github.com/pingcap/tidb/pull/26408

parser implementation

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

### DIFF
--- a/ast/functions.go
+++ b/ast/functions.go
@@ -223,6 +223,7 @@ const (
 	SubstringIndex  = "substring_index"
 	ToBase64        = "to_base64"
 	Trim            = "trim"
+	Translate       = "translate"
 	Upper           = "upper"
 	Ucase           = "ucase"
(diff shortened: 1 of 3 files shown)


kennytm (approved):

ast/functions.go
```diff
@@ -223,6 +223,7 @@ const (
 	SubstringIndex  = "substring_index"
 	ToBase64        = "to_base64"
 	Trim            = "trim"
+	Translate       = "translate"
```
XuHuaiyu, line 226: Should we add "translate" in UnReservedKeyWord?

wzru: Yes. `translate` is unreserved keyword.

wjhuang2016 (approved):

PR #1286, opened 2021-07-24
ddl: add drop column with index flag

AilinKid:

1: add drop column with index flag
2:  since the being dropped column and  being dropped indexes are schema-change combined, we should set a flag with the column for some convenient indentification
used in https://github.com/pingcap/tidb/pull/26527

### Check List 

Tests 

 - No code


zimulala (approved):
LGTM

kennytm (approved):

PR #1287, opened 2021-07-27
restore: support restoring SQL wrapping with TiDB special comments

tangenta:

TiDB Binlog and Lightning use regex to substitute the TiDB-specific keyword with special comments:

```go
FeatureIDAutoRandom:   regexp.MustCompile(`(?P<REPLACE>(?i)AUTO_RANDOM\b\s*(\s*\(\s*\d+\s*\)\s*)?)`),
...
FeatureIDForceAutoInc: regexp.MustCompile(`(?P<REPLACE>(?i)FORCE)\b\s*AUTO_INCREMENT\s*`),
```

The regex is fragile because it is easy to get things wrong when the table names or column names contain these keywords, for example:

```sql
CREATE TABLE AUTO_RANDOM (id int);
```

is incorrectly rewritten to
```sql
CREATE TABLE `/*T![auto_rand] AUTO_RANDOM AUTO_RANDOM */ ` (id int);
```

Comparing with pattern matching, using `ast.Restore` to wrap special comments is not prone to error.

- Collect TiDB-specific features into `tidb/features.go` as the single source of truth.
- Add a restore flag: `RestoreTiDBSpecialComment`, support wrapping with TiDB special comment. 
- Remove `SpecialCommentsController`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

NA

Related changes


kennytm:
what is the reason removing `SpecialCommentsController` and exposing the `SpecialComments` map directly 🤔

tidb/features.go
```diff
@@ -0,0 +1,46 @@
...
+
+var featureIDs = map[string]struct{}{
+	FeatureIDAutoRandom:     {},
+	FeatureIDAutoIDCache:    {},
+	FeatureIDAutoRandomBase: {},
+	FeatureIDClusteredIndex: {},
+	FeatureIDForceAutoInc:   {},
+}
+
+func CanParseFeature(fs ...string) bool {
```
tiancaiamao, line 39: The old code is more general, the parser is not specific to TiDB and you can register the feature to control it's behavior ...

tiancaiamao: OK, since you are the original author of this code, if you thinks it's OK, I have no problem with it.

tiancaiamao (approved):

lance6716:
Thanks.
TODO: DM needs to revert https://github.com/pingcap/dm/pull/1823 after update parser dependency

tidb/features.go
tangenta: > the parser is not specific to TiDB

But the keywords(`AUTO_RANDOM`, `CLUSTERED_INDEX`...) are TiDB-specific. It is reasonable to attach these keywords to fixed feature IDs.

> you can register the feature to control it's behavior

The register mechanism is not that convenient. If the user registers a new feature ID(`/*T![new_feature_id] */`), he must also modify the code in the parser to add a new keyword to wrap. If the user doesn't need the new keyword, he should use `/*T xxx */` instead.

kennytm (approved):

PR #1297, opened 2021-08-05
error code: add a new error code

xiongjiwei:

add a new error: `ErrCannotConvertString`, used for convert string through different charset failed.


xhebox (approved):
Reference https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_cannot_convert_string

PR #1299, opened 2021-08-06
*: support new placement rules syntax

xhebox:

The syntax part for pingcap/tidb#26581. One extra change: `constraints` is moved from `unreserved` to `notkeywords`, since mysql does not have such keyword.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes


parser.y
```diff
@@ -1562,13 +1622,13 @@ PlacementOptions:
...
+	"ADD" "PLACEMENT" "POLICY" OldPlacementOptions
 	{
 		spec := $4.(*ast.PlacementSpec)
 		spec.Tp = ast.PlacementAdd
 		$$ = spec
 	}
-|	"ALTER" "PLACEMENT" "POLICY" PlacementOptions
+|	"ALTER" "PLACEMENT" "POLICY" OldPlacementOptions
 	{
 		spec := $4.(*ast.PlacementSpec)
```
morgo, line 1633: Should we create an issue to track removing the `OldPlacementOptions`? I assume it will be a good first issue.

xhebox: I've added it in the todo of pingap/tidb#26581 :)

There will be a conflict between the implementation. Once the new syntax is implemented, the old syntax must be removed.

parser.y
```diff
@@ -691,6 +690,16 @@ import (
 	learner               "LEARNER"
 	verboseType           "VERBOSE"
 	voter                 "VOTER"
+	constraints           "CONSTRAINTS"
```
AilinKid, line 693: we used to put this alphabetically

ast/ddl.go
```diff
@@ -71,12 +71,25 @@ const (
 	DatabaseOptionCharset
 	DatabaseOptionCollate
 	DatabaseOptionEncryption
+	DatabaseOptionPlacementPrimaryRegion
+	DatabaseOptionPlacementRegions
+	DatabaseOptionPlacementFollowerCount
+	DatabaseOptionPlacementVoterCount
+	DatabaseOptionPlacementLearnerCount
+	DatabaseOptionPlacementSchedule
+	DatabaseOptionPlacementConstraints
```
AilinKid, line 80: offset calculation base on the consecutive definition here. 
once another definition inserted among here, a reflecting map is necessary.

AilinKid (commented):
Rest LGTM

ast/ddl.go
xhebox: It is intended. All options are same for db/table/partition anyway. I could assign every table/db enum one specific number like `0x3000 + iota`.

parser.y
xhebox: Done

AilinKid:
Screenshot_20210816_172122

Kenny has some suggestions with my draft PR (in fact), PTAL

xhebox:
> 
> 
> Screenshot_20210816_172122
> 
> Kenny has some suggestions with my draft PR (in fact on your base), PTAL

It is addressed.

AilinKid (approved):

AilinKid:
sure, better `make parser` again to resolve the conflicts

morgo (approved):

PR #1305, opened 2021-08-11
parser: create and alter placement policy

AilinKid:
support create and alter placement policy

add syntax for 
1: `create placement policy if not exists x placement_opts`
2: `alter placement policy if exists x placement_opts`

this is based on https://github.com/pingcap/parser/pull/1299, wait for it merged first.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


ast/ddl.go
```diff
@@ -71,12 +73,25 @@ const (
...
+	DatabaseOptionPlacementFollowerCount
+	DatabaseOptionPlacementVoterCount
+	DatabaseOptionPlacementLearnerCount
+	DatabaseOptionPlacementSchedule
+	DatabaseOptionPlacementConstraints
+	DatabaseOptionPlacementLeaderConstraints
+	DatabaseOptionPlacementLearnerConstraints
+	DatabaseOptionPlacementFollowerConstraints
+	DatabaseOptionPlacementVoterConstraints
+	DatabaseOptionPlacementPolicy
```
kennytm, line 87: how about defining the db options to have the same values as the table options like

```go
    DatabaseOptionPlacementPrimaryRegion = DatabaseOptionType(TableOptionPlacementPrimaryRegion)
    DatabaseOptionPlacementRegions = DatabaseOptionType(TableOptionPlacementRegions)
    // etc
```

...

ast/ddl.go
```diff
@@ -94,6 +109,13 @@ func (n *DatabaseOption) Restore(ctx *format.RestoreCtx) error {
 		ctx.WriteKeyWord("ENCRYPTION")
 		ctx.WritePlain(" = ")
 		ctx.WriteString(n.Value)
+	case DatabaseOptionPlacementPrimaryRegion, DatabaseOptionPlacementRegions, DatabaseOptionPlacementFollowerCount, DatabaseOptionPlacementLeaderConstraints, DatabaseOptionPlacementLearnerCount, DatabaseOptionPlacementVoterCount, DatabaseOptionPlacementSchedule, DatabaseOptionPlacementConstraints, DatabaseOptionPlacementFollowerConstraints, DatabaseOptionPlacementVoterConstraints, DatabaseOptionPlacementLearnerConstraints, DatabaseOptionPlacementPolicy:
+		tblOpt := TableOption{
+			Tp:        TableOptionType(n.Tp-DatabaseOptionPlacementPrimaryRegion) + TableOptionPlacementPrimaryRegion,
```
kennytm, line 114: ... the thing is, here we rely on the db options are declared in the same order as the table options, which is not clear on either declaration site (lines 76 and 1908). 

if the order in either side is changed (as there is no explicit relationship mentioned) this will be restored to the wrong option.

AilinKid:
move to https://github.com/pingcap/parser/pull/1313

PR #1306, opened 2021-08-13
charset: clean up some code about charset

xiongjiwei:

- use a map to store the charset info directly
- remove `Desc` struct, it is definitely same with `Charset`

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/charset/charset.go
+++ b/charset/charset.go
@@ -15,6 +15,7 @@ package charset
 
 import (
 	"strings"
+	"sync/atomic"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/parser/mysql"
@@ -24,7 +25,28 @@ import (
 var (
 	ErrUnknownCollation         = terror.ClassDDL.NewStd(mysql.ErrUnknownCollation)
 	ErrCollationCharsetMismatch = terror.ClassDDL.NewStd(mysql.ErrCollationCharsetMismatch)
+
+	// enableCharsetFeat is 1 indicate the charset feature is enabled.
+	enableCharsetFeat = int32(0)
 )
+
+func EnableCharsetFeat() {
+	SetCharsetFratEnabledForTest(true)
+}
+
+// SetCharsetFratEnabledForTest set charset feature enabled. Only used in test.
+func SetCharsetFratEnabledForTest(flag bool) {
+	if flag {
+		atomic.StoreInt32(&enableCharsetFeat, 1)
+		return
+	}
+	atomic.StoreInt32(&enableCharsetFeat, 0)
+}
+
+// CharsetFeatEnabled return true if charset feature is enabled.
+func CharsetFeatEnabled() bool {
+	return atomic.LoadInt32(&enableCharsetFeat) == 1
+}
 
 // Charset is a charset.
 // Now we only support MySQL.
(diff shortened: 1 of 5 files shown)


charset/charset.go
```diff
@@ -69,17 +67,14 @@ var supportedCollationNames = map[string]struct{}{
...
-	Name             string
-	Desc             string
-	DefaultCollation string
-	Maxlen           int
-}
-
 // GetSupportedCharsets gets descriptions for all charsets supported so far.
-func GetSupportedCharsets() []*Desc {
-	return descs
+func GetSupportedCharsets() []*Charset {
```
tangenta, line 71: This change brings more memory allocations, can we avoid them?

xiongjiwei: actually, I think it's fine, I search the usage of this function in tidb, it is only used for SQL like `show charset`.
I believe there are more codes like this, it is not a big problem. If we want to avoid it anyway, we must maintain another slice to hold the same information, it hurts the brain.

tangenta: > I search the usage of this function in tidb, it is only used for SQL like show charset.

Ok.

> it hurts the brain.

I don't think this is a problem. We can encapsulate the slice into a struct to keep the readability.

charset/charset_test.go
```diff
@@ -78,10 +78,8 @@ func (s *testCharsetSuite) TestValidCustomCharset(c *C) {
 }
 
 func (s *testCharsetSuite) TestGetSupportedCharsets(c *C) {
-	charset := &Charset{"test", "test_bin", nil, "Test", 5}
-	charsetInfos = append(charsetInfos, charset)
 	descs := GetSupportedCharsets()
-	c.Assert(len(descs), Equals, len(charsetInfos)-1)
+	c.Assert(len(descs), Equals, len(charsetInfos))
```
tangenta, line 82: I don't think this test is meaningful. How about removing it?

tangenta (commented):
There are a few exported functions/struct changes:
- Desc/GetCharsetDesc
- GetSupportedCharsets
- GetCharsetInfo

This requires the users to change the code when they upgrade the dependency. I think this is acceptable since the changes should be easy. Please also update the TiDB part.

tangenta (approved):

charset/charset.go
```diff
@@ -94,31 +89,26 @@ func ValidCharsetAndCollation(cs string, co string) bool {
 	if cs == "" {
 		cs = "utf8"
 	}
-	cs = strings.ToLower(cs)
-	c, ok := charsets[cs]
-	if !ok {
+	chs, err := GetCharsetInfo(cs)
```
xhebox, line 92: I think `ToLower` is missed, sure it is intended?

charset/charset.go
```diff
@@ -94,31 +89,26 @@ func ValidCharsetAndCollation(cs string, co string) bool {
...
 // GetDefaultCollation returns the default collation for charset.
 func GetDefaultCollation(charset string) (string, error) {
-	charset = strings.ToLower(charset)
-	if charset == CharsetBin {
-		return CollationBin, nil
-	}
-	c, ok := charsets[charset]
-	if !ok {
-		return "", errors.Errorf("Unknown charset %s", charset)
+	cs, err := GetCharsetInfo(charset)
```
xhebox, line 107: ditto.

xhebox (commented):
Do we also cleanup collation same way?

xhebox (approved):

xiongjiwei:
> Do we also cleanup collation same way?

the collation is actually on the tidb side.😅

xiongjiwei:
> Do we also cleanup collation same way?

I think the whole file should be rewritten. 😅

PR #1312, opened 2021-08-18
support parsing SQL with encodings other than utf8

tangenta:

Related to https://github.com/pingcap/tidb/issues/26812

Add field `encoding.Decoder` to the parser config.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

NA

Related changes

NA


zimulala (commented):
This PR should be associated with https://github.com/pingcap/tidb/issues/26813

charset/charset.go
```diff
@@ -57,15 +57,18 @@ var charsetInfos = map[string]*Charset{
...
-	CollationUTF8MB4: {},
-	CollationASCII:   {},
-	CollationLatin1:  {},
-	CollationBin:     {},
+	CollationUTF8:          {},
+	CollationUTF8MB4:       {},
+	CollationASCII:         {},
+	CollationLatin1:        {},
+	CollationBin:           {},
+	CollationGBKChineseCI:  {},
```
xiongjiwei, line 70: `supportedCollationNames` is not expected to change, it is used for `show collations` when new collation is not enabled.

charset/charset.go
```diff
@@ -57,15 +57,18 @@ var charsetInfos = map[string]*Charset{
 	CharsetASCII:   {CharsetASCII, CollationASCII, make(map[string]*Collation), "US ASCII", 1},
 	CharsetLatin1:  {CharsetLatin1, CollationLatin1, make(map[string]*Collation), "Latin1", 1},
 	CharsetBin:     {CharsetBin, CollationBin, make(map[string]*Collation), "binary", 1},
+	CharsetGBK:     {CharsetGBK, CollationGBKChineseCI, make(map[string]*Collation), "GBK", 2},
```
xiongjiwei, line 60: I have a config to control the parser can recognize the `gbk` or not, no need to add it here.
When this feature is GA, we will remove the config and add the do the things like this.

parser_test.go
```diff
@@ -6244,3 +6244,55 @@ func (s *testParserSuite) TestPlanRecreator(c *C) {
...
+	p.SetParserConfig(parser.ParserConfig{Decoder: decoder})
+	stmt, err = p.ParseOneStmt(sql, "", "")
+	c.Assert(err, IsNil)
+	_, _ = stmt.Accept(checker)
+	c.Assert(checker.tblName, Equals, "测试表")
+	c.Assert(checker.colName, Equals, "测试列")
+	c.Assert(checker.expr, Equals, "GBK测试用例")
+}
+
+type gbkEncodingChecker struct {
```
xiongjiwei, line 6272: what is this used for, did not see any usage in this PR?

yy_parser.go
```diff
@@ -132,11 +136,18 @@ func (parser *Parser) SetParserConfig(config ParserConfig) {
...
 	parser.SetStrictDoubleTypeCheck(config.EnableStrictDoubleTypeCheck)
 	parser.lexer.skipPositionRecording = config.SkipPositionRecording
+	parser.decoder = config.Decoder
 }
 
 // Parse parses a query string to raw ast.StmtNode.
 // If charset or collation is "", default charset and collation will be used.
 func (parser *Parser) Parse(sql, charset, collation string) (stmt []ast.StmtNode, warns []error, err error) {
+	if parser.decoder != nil {
+		sql, _, err = transform.String(parser.decoder, sql)
```
xiongjiwei, line 146: I assume we can not transform the whole sql like this, e.g. `select _utf8mb4 '汉字'`, because we have `_utf8mb4` before the string so that this string should be interpreted as utf8mb4 encoding.

parser_test.go
tangenta: This is used in the test.

charset/charset.go
zimulala, line 60: This character set is supported by the switch in [PR](https://github.com/pingcap/tidb/pull/27288), so it should not be added directly here.
Related code: https://github.com/pingcap/tidb/pull/27288/files#diff-dd39cedcb0d828a897f39576501c11cceabb22408b0f0a08afe34574ea87fc4dR55

lexer.go
```diff
@@ -175,6 +189,7 @@ func (s *Scanner) Lex(v *yySymType) int {
 		tok == stringLit &&
 		s.r.s[v.offset] == '"' {
 		tok = identifier
+		v.ident = s.tryDecodeToUTF8(v.ident)
```
xiongjiwei, line 192: there are many codes like this, they all convert `v.ident` to its utf8 version if the token is the identifier, can we just write it at L214 and remove the others.

lexer.go
```diff
@@ -892,11 +911,7 @@ func (r *reader) peek() rune {
 		return unicode.ReplacementChar
 	}
 	v, w := rune(r.s[r.p.Offset]), 1
-	switch {
-	case v == 0:
-		r.w = w
-		return v // illegal UTF-8 encoding
-	case v >= 0x80:
```
xiongjiwei, line 899: I think here should also consider the charset, including the `UNDERSCORE_CHARSET`, maybe you can check how MySQL implements the `lexer`  https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/sql/sql_lex.cc#L1359

yy_parser.go
```diff
@@ -139,6 +140,9 @@ func (parser *Parser) SetParserConfig(config ParserConfig) {
 func (parser *Parser) Parse(sql, charset, collation string) (stmt []ast.StmtNode, warns []error, err error) {
 	if charset == "" {
 		charset = mysql.DefaultCharset
+	} else {
+		enc, _ := chs.Lookup(charset)
```
zimulala, line 144: Do we need to filter out character sets that are not currently supported?

lexer.go
```diff
@@ -72,6 +76,8 @@ type Scanner struct {
 	// be pertained.
 	lastKeyword3 int
 
+	nextTokenCharset string
```
zimulala, line 79: Please add a comment.

yy_parser.go
tangenta: We had better filter them in TiDB because it is also used by other users.

xiongjiwei:
https://github.com/pingcap/parser/blob/562fed23b4fb6fffe012f8c3d46d8adf5c3ac744/lexer.go#L900-L903

`peek` is widely used in `lexer`, such as `incAsLongAs`. If the SQL is `gbk` encoding, there may be an error to decode it as `utf8` encoding?

tangenta:
If there is an utf8 decoding error, the bytes are interpreted into a four-byte integer. We can still parse it into gbk later because the information is not lost.

lexer.go
```diff
@@ -134,11 +143,72 @@ func (s *Scanner) AppendError(err error) {
...
+	sliceLen := 4
+	if charLength != nil {
+		sliceLen = charLength(Slice(lit))
+	}
+	if len(lit) >= sliceLen {
+		lit = lit[:sliceLen]
+	}
+	for {
+		result, size, err := transform.Bytes(decoder, Slice(lit))
+		if err != nil && size == 0 {
```
AilinKid, line 166:
```suggestion
		if err != nil || size == 0 { ?
```

lexer.go
```diff
@@ -134,11 +143,72 @@ func (s *Scanner) AppendError(err error) {
...
+			// Too much characters, we only need one.
+			lit = lit[:len(lit)-1]
+			continue
+		}
+		ur, _ := utf8.DecodeRune(result)
+		return ur, size
+	}
+}
+
+func (s *Scanner) tryDecodeToUTF8String(lit string) string {
```
AilinKid, line 179: go string always be the valid string with 0xfffd substituted. is that ok here?

xiongjiwei (approved):

misc.go
```diff
@@ -14,9 +14,10 @@
 package parser
 
 import (
-	"strings"
-
-	"github.com/pingcap/parser/charset"
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/transform"
+	"reflect"
+	"unsafe"
```
zimulala, line 20: Please format it.

charset/encoding_table.go
```diff
@@ -258,3 +259,26 @@ var encodings = map[string]struct {
...
+	// https://en.wikipedia.org/wiki/GBK_(character_encoding)#Layout_diagram
+	"gbk": func(bs []byte) int {
+		if len(bs) == 0 || bs[0] < 0x80 {
+			// A byte in the range 00–7F is a single byte that means the same thing as it does in ASCII.
+			return 1
+		}
+		return 2
+	},
+	"utf-8": func(bs []byte) int {
+		_, size := utf8.DecodeRune(bs)
```
xiongjiwei, line 281: `utf8.DecodeRune` is expensive, utf8 can use the first byte to determine the length.
```
b < 0x80  ==> 1
b < 0xe0  ==> 2
b < 0xf0  ==> 3
other        ==> 4
```

lexer.go
```diff
@@ -34,11 +38,72 @@ type Pos struct {
...
+			srcEnd = len(src)
+		}
+		nDest, nSrc, err := e.decoder.Transform(dest[destOffset:], src[srcOffset:srcEnd], false)
+		destOffset += nDest
+		srcOffset += nSrc
+		if err == nil {
+			if srcOffset >= len(src) {
+				return string(dest[:destOffset]), ok
+			}
+		} else if err == transform.ErrShortDst {
```
zimulala, line 87: Why is there no error here?

misc.go
```diff
@@ -991,18 +990,13 @@ func (s *Scanner) isTokenIdentifier(lit string, offset int) int {
...
-		return identifier
-	}
-	lval.ident = cs.Name
-	return underscoreCS
+// Slice converts string to slice without copy.
+// Use at your own risk.
+func Slice(s string) (b []byte) {
+	pbytes := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+	pstring := (*reflect.StringHeader)(unsafe.Pointer(&s))
+	pbytes.Data = pstring.Data
```
zimulala, line 998:
```suggestion
	pBytes.Data = pString.Data
```

lexer.go
tangenta: Because this error indicates the destination bytes array is too short, we need to enlarge the capacity.

zimulala (approved):
LGTM

zimulala:
@xiongjiwei Do you need another look?

xiongjiwei (approved):

kennytm:
why is the Circle CI requirement still there?

zimulala (approved):

PR #1313, opened 2021-08-18
parser: support create and alter placement policy

AilinKid:
support create and alter placement policy

add syntax for 
1: `create placement policy if not exists x placement_opts`
2: `alter placement policy if exists x placement_opts`

this is based on https://github.com/pingcap/parser/pull/1299, wait for it merged first.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -851,6 +851,7 @@ import (
 	CreateIndexStmt            "CREATE INDEX statement"
 	CreateImportStmt           "CREATE IMPORT statement"
 	CreateBindingStmt          "CREATE BINDING  statement"
+	CreatePolicyStmt           "CREATE PLACEMENT POLICY statement"
 	CreateSequenceStmt         "CREATE SEQUENCE statement"
 	CreateStatisticsStmt       "CREATE STATISTICS statement"
 	DoStmt                     "Do statement"
@@ -1300,6 +1301,7 @@ import (
 	PlacementRole                          "Placement rules role option"
 	OldPlacementOptions                    "Placement rules options"
 	PlacementOption                        "Anonymous or direct placement option"
+	PlacementOptionList                    "Anomymous or direct placement option list"
 	PlacementSpec                          "Placement rules specification"
 	PlacementSpecList                      "Placement rules specifications"
 	AttributesOpt                          "Attributes options"
@@ -1521,6 +1523,20 @@ PlacementLabelConstraints:
 	"CONSTRAINTS" "=" stringLit
 	{
 		$$ = $3
+	}
+
+PlacementOptionList:
+	PlacementOption
+	{
+		$$ = []*ast.TableOption{$1.(*ast.TableOption)}
+	}
+|	PlacementOptionList PlacementOption
+	{
+		$$ = append($1.([]*ast.TableOption), $2.(*ast.TableOption))
+	}
+|	PlacementOptionList ',' PlacementOption
+	{
+		$$ = append($1.([]*ast.TableOption), $3.(*ast.TableOption))
 	}
 
 PlacementOption:
@@ -10803,6 +10819,7 @@ Statement:
 |	CreateUserStmt
 |	CreateRoleStmt
 |	CreateBindingStmt
+|	CreatePolicyStmt
 |	CreateSequenceStmt
 |	CreateStatisticsStmt
 |	DoStmt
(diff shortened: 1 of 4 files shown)


parser.y
```diff
@@ -1523,6 +1526,20 @@ PlacementLabelConstraints:
 		$$ = $3
 	}
 
+PlacementOptionList:
+	PlacementOption
+	{
+		$$ = []*ast.TableOption{$1.(*ast.TableOption)}
```
xhebox, line 1532: It is `ast.PlacementOption` not `TableOption`.

parser_test.go
```diff
@@ -3361,6 +3361,36 @@ func (s *testParserSuite) TestDDL(c *C) {
...
+		{"create placement policy x constraints='ww'", true, "CREATE PLACEMENT POLICY `x` CONSTRAINTS = 'ww'"},
+		{"create placement policy x leader_constraints='ww'", true, "CREATE PLACEMENT POLICY `x` LEADER_CONSTRAINTS = 'ww'"},
+		{"create placement policy x follower_constraints='ww'", true, "CREATE PLACEMENT POLICY `x` FOLLOWER_CONSTRAINTS = 'ww'"},
+		{"create placement policy x voter_constraints='ww'", true, "CREATE PLACEMENT POLICY `x` VOTER_CONSTRAINTS = 'ww'"},
+		{"create placement policy x learner_constraints='ww'", true, "CREATE PLACEMENT POLICY `x` LEARNER_CONSTRAINTS = 'ww'"},
+		{"create placement policy x primary_region='cn' regions='us' schedule='even'", true, "CREATE PLACEMENT POLICY `x` PRIMARY_REGION = 'cn' REGIONS = 'us' SCHEDULE = 'even'"},
+		{"create placement policy x primary_region='cn', leader_constraints='ww', leader_constraints='yy'", true, "CREATE PLACEMENT POLICY `x` PRIMARY_REGION = 'cn' LEADER_CONSTRAINTS = 'ww' LEADER_CONSTRAINTS = 'yy'"},
+		{"create placement policy if not exists x regions = 'us', follower_constraints='yy'", true, "CREATE PLACEMENT POLICY IF NOT EXISTS `x` REGIONS = 'us' FOLLOWER_CONSTRAINTS = 'yy'"},
```
lcwangchao, line 3378: What happens if we execute "create placement policy x placement policy y"? Set `placement policy <policy>` is also a placement option, maybe we should separate it  with other options?

AilinKid: nice catch, done

xhebox: `PlacementOption2` is not a good name, what about `DirectPlacementOption` or somethine similar?

parser.y
```diff
@@ -1568,6 +1586,9 @@ PlacementOption:
 	{
 		$$ = &ast.PlacementOption{Tp: ast.PlacementOptionLearnerConstraints, StrValue: $3}
 	}
+
+PlacementOption:
+	DirectPlacementOption
 |	"PLACEMENT" "POLICY" EqOpt stringLit
```
lcwangchao, line 1592: Is `PlacementOption` used now? May be below is better because a database or table can not assign a policy and set the direct placement option at the same time.

```
PlacementOptionListOrPolicy:
	PlacementOptionList
|       "PLACEMENT" "POLICY" EqOpt stringLit
```

AilinKid: `DirectPlacementOption` is a basic option set for create policy usage.
`PlacementOption` is super set of `DirectPlacementOption`, plus placement policy for create table usage.

`a database or table can't assign a policy and set the direct placement option at the same time`, @morgo Is this true

morgo: Yes that's true. For simplification we decided setting a placement policy unsets all placement options (and vice versa)

AilinKid: After code trying. we couldn't convert `PlacementPolicyOrOptionList` to a single database option since databaseOptList assume it's element is every single database option, no a list here.

for example
`placement policy = x character set='utf8' follower_constraints='yy'`
parser will treat every token individually, every element is a database option, we couldn't divide/tell the placement_options list and placement_policy apart at the first beginning. For this case character_set is not a placement option.

xhebox (approved):

parser.y
lcwangchao: I think you are right, we should at least guarantee `follower=1 character set='utf8' follower_constraints='yy'` can be accepted by parser.

lcwangchao (approved):

PR #1314, opened 2021-08-18
parser: add sample rate for analyze option

winoros:

Add new option `SAMPLERATE` to support collect samples by sample rate for statistics

Add new grammar

### Check List 

Tests 

 - Unit test


tangenta:
What is the unit? Percentage?

tangenta (approved):

PR #1321, opened 2021-08-23
parser: add enable field for mutliSchemaInfo in job

AilinKid:

add a enable field to mutliSchemaInfo in job

add a enable field to mutliSchemaInfo in job

### Check List 

Tests 

 - No code
 it will be tested in the latter TiDB pull request

Related changes

### DIFF
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -198,7 +198,8 @@ func NewDDLReorgMeta() *DDLReorgMeta {
 
 // MultiSchemaInfo keeps some information for multi schema change.
 type MultiSchemaInfo struct {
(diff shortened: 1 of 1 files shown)


tangenta (commented):
Since the `MultiSchemaInfo` is a pointer in the `Job`, can we check `job.MultiSchemaInfo != nil` to determine whether it is enabled?

AilinKid:
> Since the `MultiSchemaInfo` is a pointer in the `Job`, can we check `job.MultiSchemaInfo != nil` to determine whether it is enabled?

for this case, we can

PR #1322, opened 2021-08-25
move placement into not keywords

xhebox:
Signed-off-by: xhe <xw897002528@gmail.com>

`placement` is not a keyword in mysql. Precedence is used to avoid ambiguity.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -210,7 +210,6 @@ import (
 	over              "OVER"
 	partition         "PARTITION"
 	percentRank       "PERCENT_RANK"
-	placement         "PLACEMENT"
 	precisionType     "PRECISION"
 	primary           "PRIMARY"
 	procedure         "PROCEDURE"
@@ -661,6 +660,7 @@ import (
 	max                   "MAX"
 	now                   "NOW"
 	optRuleBlacklist      "OPT_RULE_BLACKLIST"
+	placement             "PLACEMENT"
 	plan                  "PLAN"
 	position              "POSITION"
 	primaryRegion         "PRIMARY_REGION"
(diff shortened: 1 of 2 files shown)


parser.y
```diff
@@ -2360,6 +2361,7 @@ KeyOrIndexOpt:
 |	KeyOrIndex
 
 ColumnKeywordOpt:
+	/* empty */ %prec empty
```
AilinKid, line 2364: why should we put a prec here?

dveeden:
This looks related to https://github.com/pingcap/docs/pull/6195

parser.y
xhebox: `ColumnKeywordOpt` conflicts with `placement`. No prec will skip the conflict resolution of goyacc, however.

xhebox:
> This looks related to [pingcap/docs#6195](https://github.com/pingcap/docs/pull/6195)

Yes, I was meant to update the document. But after an internal discussion, we now agreed to workaround the parser and make it a real unreserved/not-token keyword, instead of adding it to the reserved(which is not compatible with mysql).

I'll review other tokens proposed in the doc repository later.

tangenta (approved):

tisonkun:
@xhebox could you explicit `@` the participants of the "internal discussion"? "With an offline discussion with @a and @b" or so.

xhebox:
> 
> 
> @xhebox could you explicit `@` the participants of the "internal discussion"? like `After an offline discussion with @a and @b, ...`.

It is online but in the lark. @bb7133 @djshow832 

Anyway, I will merge it now.

/merge

PR #1326, opened 2021-08-31
mysql: add customized release version config option

AilinKid:

Add customized release version config option which will be initialized in the TiDB makefile.

Add customized release version config option which will be initialized in the TiDB makefile.
Related-TiDB-PR https://github.com/pingcap/tidb/pull/27692

### Check List 

Tests 

Related changes

### DIFF
--- a/mysql/const.go
+++ b/mysql/const.go
@@ -30,9 +30,18 @@ var (
 	// TiDBReleaseVersion is initialized by (git describe --tags) in Makefile.
 	TiDBReleaseVersion = "None"
 
+	// CustomizedReleaseVersion is initialized by (your definition) in Makefile.
+	CustomizedReleaseVersion = "None"
+
 	// ServerVersion is the version information of this tidb-server in MySQL's format.
-	ServerVersion = fmt.Sprintf("5.7.25-TiDB-%s", TiDBReleaseVersion)
+	ServerVersion = fmt.Sprintf("5.7.25-%s", CustomizedReleaseVersion)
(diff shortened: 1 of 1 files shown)


tangenta:
Can we move this constant to TiDB?

mysql/const.go
```diff
@@ -30,10 +30,19 @@ var (
 	// TiDBReleaseVersion is initialized by (git describe --tags) in Makefile.
 	TiDBReleaseVersion = "None"
 
+	// CustomizedReleaseVersion is initialized by (your definition) in Makefile.
+	CustomizedReleaseVersion = "None"
+
 	// ServerVersion is the version information of this tidb-server in MySQL's format.
-	ServerVersion = fmt.Sprintf("5.7.25-TiDB-%s", TiDBReleaseVersion)
+	ServerVersion = fmt.Sprintf("5.7.25-%s", CustomizedReleaseVersion)
```
zimulala, line 37: Why not use `TiDBReleaseVersion` as the default value?

AilinKid: refer the related TiDB pr to see what user want
```
select version()
UbiSQL-v1.0.0-rc1
```

tangenta (approved):

mysql/const.go
zimulala: I mean, the default is the TiDB release version. After the user sets it, we change the value.

zimulala (approved):
LGTM

PR #1327, opened 2021-09-02
model: move placement policy definition into 2 parser

AilinKid:

particial close https://github.com/pingcap/tidb/issues/26581

Since we want to store the direct placement option and policy ref into tableInfo, we better move the policyInfo and it's sub struct to parser, which will eliminate the dependency from parser ---> TiDB

move placement policy definition into 2 parser

### Check List 

Tests 

 - No code

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -15,6 +15,7 @@ package model
 
 import (
 	"encoding/json"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


morgo (approved):

xhebox (approved):

AilinKid:
> What does "2 parser" mean?

mistake, cleaned

PR #1330, opened 2021-09-06
parser: move policyinfo test from tidb to parser

AilinKid:

In https://github.com/pingcap/parser/pull/1327 we move policyInfo into parser but the related test case.
So this pr move the missed policyInfo test from TiDB repo to Parser repo

move some tests.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -16,6 +16,7 @@ package model
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


model/model_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	. "github.com/pingcap/check"
 	"github.com/pingcap/parser/mysql"
 	"github.com/pingcap/parser/types"
+	"github.com/stretchr/testify/assert"
```
kennytm, line 25: you need to update go.mod.

AilinKid: good

lcwangchao (approved):

xhebox (approved):

PR #1333, opened 2021-09-07
*: support select for update of tables

jackysp:

For a multi-table join, a for update stmt locks all tables, some tables may only be read-only tables that participate in it.

Supports the of tables syntax, which specifies that only some of the tables are locked.

### Check List 

Tests

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -1063,6 +1063,7 @@ import (
 	OnDuplicateKeyUpdate                   "ON DUPLICATE KEY UPDATE value list"
 	OnCommitOpt                            "ON COMMIT DELETE |PRESERVE ROWS"
 	DuplicateOpt                           "[IGNORE|REPLACE] in CREATE TABLE ... SELECT statement or LOAD DATA statement"
+	OfTablesOpt                            "OF table_name [, ...]"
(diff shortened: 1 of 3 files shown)


parser.y
```diff
@@ -9175,42 +9176,71 @@ SelectLockOpt:
...
+|	"FOR" "UPDATE" OfTablesOpt
 	{
-		$$ = &ast.SelectLockInfo{LockType: ast.SelectLockForUpdate}
+		$$ = &ast.SelectLockInfo{
+			LockType: ast.SelectLockForUpdate,
+			Tables:   $3.([]*ast.TableName),
+		}
 	}
-|	"FOR" "SHARE"
+|	"FOR" "SHARE" OfTablesOpt
```
coocood, line 9186: We don't need to support `FOR SHARE`.

jackysp: I'm referring to the postgres syntax https://www.postgresql.org/docs/current/sql-select.html. `for share` would have been invalid as well, but the syntax seems to be fine for full support.

coocood (approved):

cfzjywxk (approved):

PR #1335, opened 2021-09-08
parser: support 'binary' symbol in set statements

tangenta:

Using `BINARY` in SET statements gets a syntax error because it is a keyword.

```sql
mysql> set @@character_set_results = binary;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 36 near ""
```

MySQL handled this problem by adding a rule:

https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/sql/sql_yacc.yy#L15902

Add a rule for `SetExpr`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

NA

Side effects

NA

Related changes

NA

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -9645,6 +9645,10 @@ SetExpr:
 	{
 		$$ = ast.NewValueExpr("ON", parser.charset, parser.collation)
 	}
+|	"BINARY"
+	{
+		$$ = ast.NewValueExpr("BINARY", parser.charset, parser.collation)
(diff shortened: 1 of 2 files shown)


zimulala (approved):
LGTM

xiongjiwei (approved):

PR #1336, opened 2021-09-09
mysql/charset: add a mapping from charset ID to charset name

tangenta:

In PR https://github.com/pingcap/tidb/pull/27875, the semantic of `character_set_results == null` is to return the result string encoded with the column charset. 

However, the charset is represented as `CharsetID`(`uint8`), I need a function to map this ID to charset name to `Lookup` the corresponding `Encoding` struct.

Add a mapping from charset ID to charset name.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

None

Side effects

None

Related changes

None


zimulala (approved):
LGTM

xiongjiwei (commented):
there is a `collationsIDMap` in `charset/charset.go` why not use that

PR #1342, opened 2021-09-14
model: support database placement option and partition placement option

zhaoxugang:
support database placement option and partition placement option

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -84,7 +84,7 @@ const (
 	ActionCreatePlacementPolicy         ActionType = 51
 	ActionAlterPlacementPolicy          ActionType = 52
 	ActionDropPlacementPolicy           ActionType = 53
-	ActionAlterTablePartitionOptions    ActionType = 54
+	ActionAlterTablePartitionPolicy     ActionType = 54
 )
 
 var actionMap = map[ActionType]string{
--- a/model/ddl.go
+++ b/model/ddl.go
@@ -84,6 +84,7 @@ const (
 	ActionCreatePlacementPolicy         ActionType = 51
 	ActionAlterPlacementPolicy          ActionType = 52
 	ActionDropPlacementPolicy           ActionType = 53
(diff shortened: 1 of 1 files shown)


zhaoxugang:
PTAL @AilinKid @xhebox

model/ddl.go
```diff
@@ -84,6 +84,7 @@ const (
 	ActionCreatePlacementPolicy         ActionType = 51
 	ActionAlterPlacementPolicy          ActionType = 52
 	ActionDropPlacementPolicy           ActionType = 53
+	ActionAlterTablePartitionOptions    ActionType = 54
```
xhebox, line 87: Maybe a smaller name. We will transform `ast.TableOption` into different actions, refer https://github.com/pingcap/tidb/blob/257140a154a13061b55993f499aa25f0cd07cdf6/ddl/ddl_api.go#L2530-L2573. We don't really have `AlterTableOption` type.

xhebox (commented):
Overlap with #1340, looks like. Maybe you need to discuss with @mjonss .

mjonss:
I just merged https://github.com/pingcap/parser/pull/1340 so the model now has both `PlacementPolicyRef` and `DirectPlacementOpts` on both `DBInfo` and `PartitionDefinition`.

model/ddl.go
zhaoxugang: > Maybe a smaller name. We will transform `ast.TableOption` into different actions, refer https://github.com/pingcap/tidb/blob/257140a154a13061b55993f499aa25f0cd07cdf6/ddl/ddl_api.go#L2530-L2573. We don't really have `AlterTableOption` type.

the constant is used for ddl.runDDLJob() to mark job type not ddl.AlterTable()

xhebox (approved):

AilinKid (approved):

PR #1352, opened 2021-09-27
terror: provide a RegisterFinish API

tiancaiamao:

Prevent regression like https://github.com/pingcap/tidb/issues/28190

TiDB will register all the error kinds in its initialization, and then call this API
After that, misuse of the terror will panic, thus the bug is exposed early.

### Check List 

Tests 

In TiDB

Code changes


tiancaiamao:
https://github.com/pingcap/tidb/pull/28438

terror/terror.go
```diff
@@ -107,6 +108,18 @@ func (m *code2ErrClassMap) Put(key string, err ErrClass) {
 	m.data.Store(key, err)
 }
 
+var registerFinish uint32
+
+// RegisterFinish makes the register of new error panic.
+// The use pattern should be regist all the error during initialization, and then call RegisterFinish.
```
kennytm, line 114:
```suggestion
// The use pattern should be register all the errors during initialization, and then call RegisterFinish.
```

ichn-hu (approved):
LGTM

tiancaiamao:
My main point is actually here https://github.com/pingcap/tidb/pull/25405#issuecomment-861173193

> Is there some way to prevent this kind of error in the future?
> Maybe nobody read the document and maybe people still make this kind of mistakes in the future.

We have got stuck on the error before, and fixed it then.
But later a contributor file a PR and misuse the terror package again...
This time we're not so lucky, the reviewers don't catch it and our customer trigger the bug

kennytm (approved):

PR #1170, opened 2021-02-20
model: add a new TableInfoVersion for separating id allocators

tangenta:
### What problem does this PR solve? 

Issue: https://github.com/pingcap/tidb/issues/982

TiDB PR: https://github.com/pingcap/tidb/pull/20708

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has exported variable/fields change

Side effects

N/A

Related changes

 - Need to update the documentation
 - Need to be included in the release note

PR #1354, opened 2021-09-28
support separating auto_increment ID from row ID

tangenta:

The parser part of https://github.com/pingcap/tidb/pull/28448.

- Support syntax `alter table t modify [force] row_id = xxx`. This is useful in the last stage when importing with Lightning.
- Add a job action `ActionRebaseRowID`.
- Change `TableInfoVersion4` to `TableInfoVersion5`. `TableInfoVersion4` is a version that is equivalent to `TableInfoVersion3` for some reasons.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

NA

Side effects

NA

Related changes

NA

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -259,6 +259,12 @@ const (
 	CurrLatestTableInfoVersion = TableInfoVersion5
 )
 
+// AutoIncrementIDIsSeparated indicates whether auto_increment ID of is separated from _tidb_row_id.
+// For details, see https://github.com/pingcap/tidb/issues/982.
+func AutoIncrementIDIsSeparated(ver uint16) bool {
+	return ver >= TableInfoVersion5
+}
+
 // ExtraHandleName is the name of ExtraHandle Column.
 var ExtraHandleName = NewCIStr("_tidb_rowid")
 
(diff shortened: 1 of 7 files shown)


model/model.go
```diff
@@ -249,16 +249,22 @@ const (
...
 	// charsets / collations will be converted to lower-case while loading from the storage.
 	TableInfoVersion3 = uint16(3)
-	// TableInfoVersion4 indicates that the auto_increment allocator in TiDB has been separated from
+	// TableInfoVersion5 indicates that the auto_increment allocator in TiDB has been separated from
 	// _tidb_rowid allocator. This version is introduced to preserve the compatibility of old tables:
-	// the tables with version < TableInfoVersion4 still use a single allocator for auto_increment and _tidb_rowid.
+	// the tables with version < TableInfoVersion5 still use a single allocator for auto_increment and _tidb_rowid.
 	// Also see https://github.com/pingcap/tidb/issues/982.
-	TableInfoVersion4 = uint16(4)
+	TableInfoVersion5 = uint16(5)
```
kennytm, line 256: ?? what happened to version 4

tangenta: Version 4 is introduced in https://github.com/pingcap/parser/pull/1170...

It is released to v5.0, so we need another version to identify that the auto IDs are separated.

